PR #15, opened 2014-09-08
Optimized SetAssemblyVersionString.ps1

ThomasBarnekow:
The SetAssemblyVersionString.ps1 PowerShell script is a pre-build script that sets the version string of AssemblyInfo.cs. While already ignoring minutes in computing the version number, the current version on master still touches AssemblyInfo.cs every time you build the SDK in Visual Studio, even if you do that during the same hour. This leads to unnecessary builds, because Visual Studio always performs a time-consuming build because at least AssemblyInfo.cs is changed by the pre-build script. At least I build the solution more often than absolutely required, just to be on the safe side. So, with the current version, I have to wait for the build to finish even though I might not have changed anything really. 

The optimized script only touches AssemblyInfo.cs if it has to. Thus, building the solution now effectively does nothing if you haven't changed anything else, just like it is supposed to be. 

The SDK is also not released multiple times a day (i.e., no CI). Therefore, the version string just uses the number of days between `[DateTime]::UtcNow` and June 1, 2014, 00:00:00 plus 5632 as its third component. The fourth component is always 0. 

Note the `UtcNow` in `[DateTime]::UtcNow`. This used to be `[DateTime]::Now`. I've used UTC so everybody uses the same time zone, i.e., UTC or Coordinated Universal Time, when computing the version string.

If required, we can certainly be more fancy (now or later) and have the version string depend on the actual time of day (in whatever resolution, e.g., hours or minutes).


ThomasBarnekow:
Rebased on development branch. Will create new pull request

PR #28, opened 2015-01-18
Removed SetAssemblyVersionString.ps1 and associated pre-build event

ThomasBarnekow:
When using the Visual Studio project, the project is sometimes built without the intention to produce a shipping assembly. However, each build results in a new AssemblyInfo.cs file even if no other change was made. This leads to unnecessary commits in the history. Therefore, the idea is to use the command line build script to build shipping assemblies with new version numbers. 

One question would also be whether we could use the default algorithm used by Visual Studio to emit version strings, e.g., when the version string is defined as "2.5.*". The third component (Build Number) would then be the number of days since January 1, 2000. The fourth component (Release) would be the number of seconds since midnight divided by two. 

The interesting thing is that the third component is currently always greater than the number of days since January 1, 2000.


ThomasBarnekow:
Replaced by #36

PR #29, opened 2015-01-21
Missing Files in Package - More Detailed Error

a-h:
When a package isn't found in the document, a generic "Specified part does not exist in the package." error is produced.  This doesn't guide the developer in which out of the many possible files in the package is
missing.

In my case, I had this problem from a file I didn't generate, so I was lost as to how to resolve the issue.  Looking on stackoverflow, these post were useful (http://stackoverflow.com/questions/11212374/specified-part-does-not-exist-in-the-package  http://www.telkoth.net/blog/?p=30), but the error message was quite vague.

### DIFF
--- a/src/Framework/OpenXmlBasePart.cs
+++ b/src/Framework/OpenXmlBasePart.cs
@@ -81,6 +81,11 @@ namespace DocumentFormat.OpenXml.Packaging
             this._uri = uriTarget;
 
             // TODO: should we delay load?
+            if (!this.OpenXmlPackage.Package.PartExists(uriTarget))
+            {
+                throw new OpenXmlPackageException(string.Format("The part with URI target {0} does not exist in the package.", uriTarget));
+            }
+
             PackagePart metroPart = this.OpenXmlPackage.Package.GetPart(uriTarget);
 
             if (this.IsContentTypeFixed &&


src/Framework/OpenXmlBasePart.cs
```diff
@@ -81,6 +81,11 @@ internal void Load(OpenXmlPackage openXmlPackage, OpenXmlPart parent, Uri uriTar
             this._uri = uriTarget;
 
             // TODO: should we delay load?
+            if (!this.OpenXmlPackage.Package.PartExists(uriTarget))
+            {
+                throw new OpenXmlPackageException(string.Format("The part with URI target {0} does not exist in the package.", uriTarget));
+            }
```
ThomasBarnekow: @a-h, I've never run into a situation where a part did not exist. Therefore, the extra invocation of `PartExists()` is overhead that we might want to save.

I assume `GetPart()` throws the exception with the unhelpful message. Therefore, there would be two approaches in my view:

1. Wrap `GetPart()` in a try-catch statement and throw your exception.
2. Have `GetPart()` throw a more verbose exception.

Given that the Open XML SDK now uses `System.IO.Packaging` developed by @EricWhiteDev, it would be possible to do the right thing right there.

@EricWhiteDev, what do you think?

src/Framework/OpenXmlBasePart.cs
```diff
@@ -81,6 +81,11 @@ internal void Load(OpenXmlPackage openXmlPackage, OpenXmlPart parent, Uri uriTar
...
+
             PackagePart metroPart = this.OpenXmlPackage.Package.GetPart(uriTarget);
```
ThomasBarnekow: Because you can't otherwise see this in the conversation, here is essentially the line that should be wrapped in a try-catch statement if you did not fix the exception message in `GetPart()` itself.

ThomasBarnekow:
@a-h and @EricWhiteDev I created another pull request in which I added a missing .resx file that had lead to failed unit tests (unrelated to this PR). Among many other strings, that .resx file also includes the unhelpful exception message you mentioned above.

I now changed that message in the .resx file to reflect the exception message proposed by @a-h and amended `Package.GetPart(Uri):PackagePart` to throw an `InvalidOperationException` with a formatted exception message that includes the non-existent URI rather than that previous, unhelpful message. This way, we don't need an extra invocation of `Package.PartExists(Uri):bool`.

EricWhiteDev:
Closing this, as the changes necessary for this PR are included in PR #51

src/Framework/OpenXmlBasePart.cs
a-h: Thanks Thomas, the situation where a part didn't exist occurred when opening an Office document that a user uploaded. It turned out that they'd used a 3rd party OCR tool which outputs to Open XML. The file opened fine in Word, but failed to process in my tool, so it was reported to me as a defect. I wanted to collect the data about what was failing so that it could be manually fixed up if required. I was building a system which processed thousands of documents a day for translation, so I saw all sorts of dodgy files. :)

a-h:
Thanks everyone. :+1:

src/Framework/OpenXmlBasePart.cs
EricWhiteDev: @a-h Question: does the document with a missing part validate using the Open XML SDK Productivity Tool?  Or does it throw an exception from within the tool?

EricWhiteDev: @a-h Another question:  I love to put dodgy files into my test file repo.  Are these files available, i.e. the do not contain confidential information?

a-h: Hi Eric, I don't work for that company any more, but I've been in touch with the team and they're going to send me the details from the old bug and some example files I can share with you. They reckon I'll get them today.

a-h: OK, so the repro is open a PDF in Adobe Acrobat 10.0.0 and save it as an MS Word file. Acrobat messes up when it detects words it thinks it can OCR (like header images) etc. Opening the file in Microsoft Word will work fine, but the SDK will fail. The workaround for the users was to open the file in MS Word, make a minor change, and hit save, MS Word would fix up the fault. Example public files can be downloaded from: https://drive.google.com/folderview?id=0By6nm3ACZguvfnJlRWdsWWtUTTN1dkh4cnVXbkhXVnpaRXdwQS0tejNOREVlTDdrem8tYWM&usp=sharing

a-h: So, while this is a fault with Acrobat, the disparity of behaviour between MS Word (which didn't show any warning dialogs etc.) and the Open XML SDK (which threw an error) caused confusion.

EricWhiteDev: Ultimately, I'd like the SDK to have the same behavior as Word - if a file is missing a part, and it is possible to add the part with reasonable content (in some cases just an empty part), then we should do so.

Our priorities currently are:
- Get the build system / xunit tests squared away
- Office 2016 support
- Fix the validator
- following this, priority is TBD

EricWhiteDev: Thanks, I've downloaded the document.

PR #36, opened 2015-04-14
Extend Open-XML-SDK

ThomasBarnekow:
I needed to rewrite my commit history and, therefore, combined the previous pull requests into a single request. The very first pull requests -- some of which I already combined once -- come as individual commits in this pull request.

The "Save and Clone functionality" and "Flat OPC conversion" is what @EricWhiteDev and I discussed quite a while ago. CreateFromTemplate lets you create a document from a template.

The next commit adds an ```XName``` property to both ```OpenXmlElement``` and ```OpenXmlAttribute```. This supports integration with projects using Linq for XML.

Finally, when using the Visual Studio project, the project is sometimes built without the intention to produce a shipping assembly. However, each build results in a new AssemblyInfo.cs file even if no other change was made. This leads to unnecessary commits in the history. Therefore, the idea is to use the command line build script to build shipping assemblies with new version numbers.

One question would also be whether we could use the default algorithm used by Visual Studio to emit version strings, e.g., when the version string is defined as "2.5.*". The third component (Build Number) would then be the number of days since January 1, 2000. The fourth component (Release) would be the number of seconds since midnight divided by two.

The interesting thing is that the third component is currently always greater than the number of days since January 1, 2000.


ThomasBarnekow:
Based on our discussion today, @EricWhiteDev created a vNext branch in this repo. We decided that I should create another pull request with my changes so Eric could pull it into the newly created vNext branch. Thus, closing this PR, which has also changed its content because of the changes to the branches.

PR #43, opened 2015-06-19
Simpler cross-platform solution

stever:
These changes to the solution remove PowerShell and makefile scripts, but will build on with either Visual Studio or xbuild in Mono, as a typical assembly project without automatic version numbering. The makefile didn't work for me and this seems like the simplest solution to maintain.


ThomasBarnekow:
@stever It seems that most of your good points have been covered by changes we've made on the vNext branch. Can I ask you to please have a look at vNext and see whether there is anything missing from what you wanted to achieve? Here is a high-level assessment from my point of view:

* "Fixing case-insensitive path ..." **is covered**.
* "Use the default Debug and Release output paths" **is not covered**. I would also support using the default paths. And while this would normally be a breaking change, the other changes we've made will already break existing paths. So, now would be the time to make the change you are suggesting.
* "Removing makefile and powershell scripts" **is covered**, at least mostly. I've just seen one leftover `makefile.def` in the Open-XML-SDK project folder (you'll only see that folder on vNext).
* "Replaced the OpenXmlObsolete ..." **is partly covered**. Travis CI complained at one point, so I replaced that one string. But we could certainly be consistent and replace them all.
* "Updated README" **is not covered**. We still have to update the README file. However, there are other updates pending, so @EricWhiteDev will probably have to do this while reflecting your changes.

I will let @EricWhiteDev chime in on how we want to move forward with this PR. One way would be for you to create a feature branch off of vNext, manually reapply the outstanding changes, and create a new PR for merging your branch into vNext (not master). Otherwise, you'd have to rebase on vNext (which already contains some of the changes on top of structural changes).

stever:
Hi @ThomasBarnekow. Thanks for taking the time to respond. I appreciate that these things are being handled along the same lines and there isn't much that is important in this pull request. I'll pick up the changes from trunk. I might contribute something more useful in future. I've got more than a passing interest in the library, but not a lot of time I suppose. Still I'm using this and also cross-platform which is really good to see. It doesn't take much work to do that. Hopefully stays that way :)

ThomasBarnekow:
Hi @stever, you'd have to pick them up from vNext rather than master. Not sure when @EricWhiteDev will merge vNext into master, but that might take a little while. Anyhow, vNext will always have the latest and greatest functionality.

We are also looking to improve cross-platform availability. Currently, there are issues with System.IO.Compression on Mono, for example. And xUnit 2.0 doesn't work on Mono due to a bug, so we can't run our test cases at the moment. We'll have to wait a little while until this gets fixed.

stever:
Sure I'll take a look at vNext sometime. Thanks again.

PR #45, opened 2015-07-12
Extend Open XML SDK

ThomasBarnekow:
@EricWhiteDev This is the PR we talked about. It contains:

* Save and Clone functionality (thread-safe through locking),
* Flat OPC conversion (leveraging some of the code that you published),
* CreateFromTemplate, and
* XName properties for both OpenXmlElement and OpenXmlAttribute.

The "Save and Clone functionality" and "Flat OPC conversion" is what @EricWhiteDev and I discussed quite a while ago. CreateFromTemplate lets you create a document from a template.

Another commit adds an XName property to both OpenXmlElement and OpenXmlAttribute. This supports integration with projects using Linq for XML.

NUnit-based tests can be found here: https://github.com/ThomasBarnekow/Open-XML-SDK-Tests. We should talk about turning them into xUnit tests (because other tests @EricWhiteDev has implemented are also based on xUnit) and including them in a companion test project.

All my tests (not only the unit tests) pass. And those unit tests also cover multithreading scenarios (because I'm using the functionality in multithreaded server-side scenarios).


ThomasBarnekow:
@EricWhiteDev Thanks for merging this!

PR #47, opened 2015-07-19
Enable AppVeyor and Travis CI

ThomasBarnekow:
Here are the things I've done to enable CI for the Open-XML-SDK:

* Add appveyor.yml and .travis.yml to configure CI servers, skipping unit tests
on Travis CI for the time being. Due to a bug in Mono, we can't xUnit 2.0 tests.

* Fix ValidationResources.resx by correcting the binary file names. Those file
names did not have the correct case in the .resx file, leading to an error
on Travis CI.

* Fix OpenXmlPackage.cs by replacing a constant with a string. This was necessary
due to a weird error message on Travis CI.

* Add *.DotSettings files to .gitignore.

* Enable NuGet Package Restore to automatically download the required NuGet
packages during the build process.

* Add build.open-xml-sdk.cmd to provide a way to build and run unit tests on
the command line.

To actually use AppVeyor (see http://www.appveyor.com) and Travis CI (see https://travis-ci.org), you will have to sign up using the OfficeDev account. This adds web hooks which will trigger AppVeyor and Travis CI builds. Both AppVeyor and Travis CI should then use the GitHub Status API to tell you whether pull requests pass or fail the checks.

For the Windows platform, AppVeyor will build the Open-XML-SDK solution and run all tests contained in the Open-XMK.SDK.Tests and System.IO.Packaging.Tests projects. Passing that check means the Open XML SDK, including System.IO.Packaging, builds and passes all tests.

With Travis CI, we are currently testing whether the SDK, including System.IO.Packaging, can be built on Mono. Due to a bug in Mono, we can't currently run the xUnit 2.0 based tests. We'll have to activate tests at a later point in time.


.gitignore
```diff
@@ -5,6 +5,7 @@
 *.suo
 *.user
 *.sln.docstates
+*.DotSettings
```
ThomasBarnekow: This is related to ReSharper settings files created for solutions.

.nuget/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="xunit.runner.console" version="2.0.0" />
```
ThomasBarnekow: The console runner can generally be used to run tests on the console or on Travis CI, for example. However, due to that Mono bug, we can't use it right now. But the build script leverages that console runner to run all tests.

.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: csharp
+solution: Open-XML-SDK.sln
+install:
+  - nuget restore Open-XML-SDK.sln
+script:
+  - xbuild /p:Configuration=Release Open-XML-SDK.sln
+  # The following tests are currently not run due to a bug in Mono
+  # - mono ./packages/xunit.runner.console.2.0.0/tools/xunit.console.exe ./System.IO.Packaging.Tests/bin/Release/System.IO.Packaging.Tests.dll
+  # - mono ./packages/xunit.runner.console.2.0.0/tools/xunit.console.exe ./DocumentFormat.OpenXml.Tests/bin/Release/DocumentFormat.OpenXml.Tests.dll
```
ThomasBarnekow: Due to the Mono bug, the tests are commented out for the time being. Once that bug is fixed, we can uncomment those lines.

DocumentFormat.OpenXml/DocumentFormat.OpenXml.csproj
```diff
@@ -249,6 +249,11 @@
       <Name>System.IO.Packaging</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <None Include="src\GeneratedCode\O12SchemaConstraintDatas.bin" />
+    <None Include="src\GeneratedCode\O14SchemaConstraintDatas.bin" />
+    <None Include="src\GeneratedCode\O15SchemaConstraintDatas.bin" />
```
ThomasBarnekow: Those files were added when I corrected the ValidationResources.resx, which used all lowercase file names. That lead to a build error on Travis CI.

DocumentFormat.OpenXml/src/Framework/OpenXmlPackage.cs
```diff
@@ -2079,7 +2079,9 @@ protected static Package FromFlatOpcDocumentCore(XDocument document, Package pac
     /// <summary>
     /// Specifies event handlers that will handle OpenXmlPackage validation events and the OpenXmlPackageValidationEventArgs. 
     /// </summary>
-    [Obsolete(ObsoleteAttributeMessages.ObsoleteV1ValidationFunctionality, false)]
+    // Building on Travis CI failed, saying that ObsoleteAttributeMessages does not contain a definition
+    // for 'ObsoleteV1ValidationFunctionality'. Thus, we've replaced the member with its value.
+    [Obsolete("This functionality is obsolete and will be removed from future version release. Please see OpenXmlValidator class for supported validation functionality.", false)]
```
ThomasBarnekow: That change was due to a really weird build error on Travis CI. While using `ObsoleteAttributeMessages.ObsoleteV1ValidationFunctionality` was no problem elsewhere, a compile error occurred at this point. Therefore, I created that workaround.

DocumentFormat.OpenXml/src/ofapi/Validation/ValidationResources.resx
```diff
@@ -113,21 +112,11 @@
...
   </resheader>
-  <assembly alias="System.Windows.Forms" name="System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" />
-  <data name="O12SchemaConstraintDatas" type="System.Resources.ResXFileRef, System.Windows.Forms">
-    <value>..\..\generatedcode\o12schemaconstraintdatas.bin;System.Byte[], mscorlib, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
-  </data>
-  <data name="O14SchemaConstraintDatas" type="System.Resources.ResXFileRef, System.Windows.Forms">
-    <value>..\..\generatedcode\o14schemaconstraintdatas.bin;System.Byte[], mscorlib, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
-  </data>
-  <data name="O15SchemaConstraintDatas" type="System.Resources.ResXFileRef, System.Windows.Forms">
-    <value>..\..\generatedcode\o15schemaconstraintdatas.bin;System.Byte[], mscorlib, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
```
ThomasBarnekow: These file names lead to build errors on Travis CI because their case differs from the actual case of the corresponding files.

DocumentFormat.OpenXml/src/ofapi/Validation/ValidationResources.resx
```diff
@@ -347,4 +336,14 @@
...
   </data>
+  <assembly alias="System.Windows.Forms" name="System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" />
+  <data name="O12SchemaConstraintDatas" type="System.Resources.ResXFileRef, System.Windows.Forms">
+    <value>..\..\GeneratedCode\O12SchemaConstraintDatas.bin;System.Byte[], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </data>
+  <data name="O14SchemaConstraintDatas" type="System.Resources.ResXFileRef, System.Windows.Forms">
+    <value>..\..\GeneratedCode\O14SchemaConstraintDatas.bin;System.Byte[], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </data>
+  <data name="O15SchemaConstraintDatas" type="System.Resources.ResXFileRef, System.Windows.Forms">
+    <value>..\..\GeneratedCode\O15SchemaConstraintDatas.bin;System.Byte[], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
```
ThomasBarnekow: These are the correct file names with the correct case.

appveyor.yml
```diff
@@ -0,0 +1,15 @@
...
+- ps: nuget restore Open-XML-SDK.sln
+build:
+  project: Open-XML-SDK.sln
+  verbosity: minimal
+test:
+  assemblies: '**\*.Tests.dll'
+notifications:
+- provider: Email
+  to:
+  - thomas@barnekow.info
```
ThomasBarnekow: You should replace that with your own email address or at least add yours to the list.

build.open-xml-sdk.cmd
```diff
@@ -0,0 +1,18 @@
...
+
+REM Build release
+set FrameworkVersion=v4.0.30319
+set FrameworkDir=%SystemRoot%\Microsoft.NET\Framework
+
+if exist "%SystemRoot%\Microsoft.NET\Framework64" (
+  set FrameworkDir=%SystemRoot%\Microsoft.NET\Framework64
+)
+
+"%FrameworkDir%\%FrameworkVersion%\msbuild.exe" /p:Configuration=Release Open-XML-SDK.sln
```
ThomasBarnekow: By using the full path for msbuild.exe, you can run the script from normal command prompts.

build.open-xml-sdk.cmd
```diff
@@ -0,0 +1,18 @@
+@echo off
+
+REM Restore NuGet packages
+.nuget\NuGet.exe restore Open-XML-SDK.sln
```
ThomasBarnekow: This will restore all NuGet packages, including the console test runner used below to run the tests.

akoeplinger:
Great :+1: 

Couple notes:

- MSBuild based NuGet package restore (i.e. checking in NuGet.targets etc.) isn't the preferred solution by the NuGet team anymore. Automatic package restore is recommended now which happens automatically when using NuGet 2.7+ in VS, on the commandline it's recommended to just run `nuget.exe restore ...` like you've done in the build scripts. See [NuGet docs](https://docs.nuget.org/consume/package-restore).
- The Mono xunit bug is fixed in the Mono master branch already, I tried running the tests with that but a lot of them fail due to exceptions in System.IO.Compression, which look like Mono bugs as well.
- Since building the csproj's via xbuild works fine, can we get rid of [Makefile-Linux-Mono](https://github.com/OfficeDev/Open-XML-SDK/blob/master/Makefile-Linux-Mono)? I don't think it's been kept up to date and I see no value in keeping it. I'd much prefer a `build.open-xml-sdk.sh` file similar to the .cmd which just invokes xbuild.
- If you run into problems with Travis, I'm one of the maintainers of the C# language support there, so feel free to reach out to me :smile:

DocumentFormat.OpenXml/src/Framework/OpenXmlPackage.cs
akoeplinger: This is indeed really weird, as it works fine a few lines below. I'll file an mcs bug.

ThomasBarnekow:
@akoeplinger Thanks. 

* What would be the best way to change what we have so that the preferred solution by the NuGet team is implemented? Should I completely revert what was done by clicking on "Enable NuGet Package Restore" (which I thought is the right way of enabling NuGet package restore)?

* It seems the latest Mono release used by Travis CI still has that bug. I don't even run into those System.IO.Compression errors. @EricWhiteDev noted there might be issues with System.IO.Compression on Mono, so it would be interesting to see whether there is a way to avoid those as well. Otherwise, the Open-XML-SDK can currently only be used on Windows.

* I have never used Makefile-Linux-Mono. While @EricWhiteDev has made changes recently, using `xbuild` would be much easier. So I would be fine with deleting it.

* Any help with Travis would be much appreciated. I've used Travis for the first time today.

akoeplinger:
@ThomasBarnekow 

- Yes. In fact, the "Enable NuGet Package Restore" context menu item was removed in NuGet 3.0. You can leave nuget.exe checked in if you want so it's easier to build on a completely fresh clone via the cmd file (some other projects just fetch it from https://nuget.org/nuget.exe on demand, whichever you like better).
- Yes, the xunit fix is only in the master (= development) branch of Mono, it's not in a stable release yet so it still fails on Travis. You can try using the [Weekly Snapshot packages](http://www.mono-project.com/docs/getting-started/install/weekly-packages/).
- Any specific issue you ran into with Travis?

ThomasBarnekow:
@akoeplinger That was quick.

* I also thought that having NuGet.exe available would be helpful. As I'm not a batch / shell programmer, could you tell me how to best fetch NuGet.exe (e.g., in a script before using it in that same script)?

* How would you use those Weekly Snapshot packages on Travis?

* I did not have any other issues than the ones I already reported. And I'm not even sure whether those are related to Travis itself. I think it was easy enough for a first-time user like myself.

I also found some documentation on "Migrating to Automatic Package Restore". I'll use that to revert those changes because I've made other changes to the projects that I don't want to use.

akoeplinger:
- Downloading a file using cmd/batch is quite impossible without relying on third-party tools. It's easy using powershell, but I guess I'd just be pragmatic and check the nuget.exe file in.
- You can use `mono: nightly` as described in the [docs](http://docs.travis-ci.com/user/languages/csharp/#Choosing-Mono-version-to-test-against) (the weekly packages were once called nightly, I'll need to fix that at one point :smile: ).
- Great to hear!

ThomasBarnekow:
@akoeplinger OK, here you go. I've migrated (back) to "Automatic Package Restore". It works nicely when building with Visual Studio and `build.open-xml-sdk.cmd`.

@EricWhiteDev Let me squash those commits into one before merging this pull request.

EricWhiteDev:
@ThomasBarnekow @akoeplinger With regards to Mono, I have heard that developers have had good success when using the System.IO.Packaging that comes with Mono, as opposed to using System.IO.Compression, which is used by the new System.IO.Packaging.  I believe that there are a couple of minor issues, but that it is possible to work around them.  So this may be our best way forward

ThomasBarnekow:
@EricWhiteDev Rebased the enable-ci branch on vNext and squashed everything into one commit. Unless there are any further issues or questions, this could now be merged into vNext.

ThomasBarnekow:
@EricWhiteDev and @akoeplinger I've already added a Mono build script, replacing the legacy Makefile-Linux-Mono. I thought this was easy enough and is also related to this PR so it didn't make sense to me to create a second PR for this. In conjunction with the other commit, we now have build scripts for Windows and Mono (which I tested on Mac OS X with Mono 4.0.2).

ThomasBarnekow:
@akoeplinger OK, good point. I've only seen `nuget` being used like that on AppVeyor, Travis CI, and OS X. I've basically mimicked the build script that's used on Travis CI and tested it on OS X.

As @EricWhiteDev has already merged it, we could update the script in a later commit.

akoeplinger:
@ThomasBarnekow afaik, the Mono for OSX installer always bundles `nuget`, it's just Linux where it's a separate package that may be missing. I agree it can be fixed later on.

DocumentFormat.OpenXml/src/Framework/OpenXmlPackage.cs
akoeplinger: Looks like it has something to do with the event, I was able to reduce it to a minimal sample: https://bugzilla.xamarin.com/show_bug.cgi?id=32451

ThomasBarnekow: @akoeplinger Thanks.

Issue #49, opened 2015-07-22
Test cases failing under circumstances to be determined


ThomasBarnekow:
Depending on where tests are run (e.g., AppVeyor or certain other machines vs. my machine), either all tests pass or just a few tests fail.

This is the result I'm getting when running the tests in `System.IO.Packaging.Tests.dll`, using the `build.open-xml-sdk.cmd` build script:

```
xUnit.net console test runner (64-bit .NET 4.0.30319.34209)
Copyright (C) 2015 Outercurve Foundation.

Discovering: System.IO.Packaging.Tests
Discovered:  System.IO.Packaging.Tests
Starting:    System.IO.Packaging.Tests
   System.IO.Packaging.Tests.Tests.T170_InvalidRelationshipId [FAIL]
      Assert.Throws() Failure
      Expected: typeof(System.Xml.XmlException)
      Actual:   typeof(System.ArgumentNullException): Value cannot be null.
      Parameter name: format
      Stack Trace:
            at System.String.Format(String format, Object arg0)
         System.IO.Packaging\InternalRelationshipCollection.cs(188,0): at System.IO.Packaging.InternalRelationshipCollection.ThrowIfInvalidXsdId(String id)
         System.IO.Packaging\Package.cs(618,0): at System.IO.Packaging.Package.DeleteRelationship(String id)
Finished:    System.IO.Packaging.Tests

=== TEST EXECUTION SUMMARY ===
   System.IO.Packaging.Tests  Total: 122, Errors: 0, Failed: 1, Skipped: 0, Time: 2,307s
```

That same build script also runs the tests contained in the `DocumentFormat.OpenXml.Tests.dll`:

```
xUnit.net console test runner (64-bit .NET 4.0.30319.34209)
Copyright (C) 2015 Outercurve Foundation.

Discovering: DocumentFormat.OpenXml.Tests
Discovered:  DocumentFormat.OpenXml.Tests
Starting:    DocumentFormat.OpenXml.Tests
   DocumentFormat.OpenXml.Tests.XlsxTests01.X001_XlsxValidation [FAIL]
      Assert.Equal() Failure
      Expected: 2
      Actual:   20
      Stack Trace:
         DocumentFormat.OpenXml.Tests\TestXlsx01.cs(203,0): at DocumentFormat.OpenXml.Tests.XlsxTests01.XlsxValidationHelper(String docName, Int32 expectedErrorCount)
[...]

PR #50, opened 2015-07-23
Fix SemanticConstraint.GetAttrNumVal method

ThomasBarnekow:
Attribute strings containing decimal numbers were not correctly converted to double for all cultures. This commit fixes that issue.

See #49 for details.


ThomasBarnekow:
Hi Eric,

Thanks. 

Do you have an idea what the underlying issue of the error (see my other recent emails) in System.IO.Packaging is? Can this be fixed very easily?

It looks like you forgot to include a resource file (or files) or you are referencing an “existing" System assembly that doesn’t in fact exist.

I’ve played a bit with resources and if you at least have the main resource file, you won’t run into the exception that we are seeing. GetString would return the default, which is likely the English resource string.

So, if you can, please at least tell me what the issue is, because that bothers me (that’s how I am, I want to solve problems and don’t stop until they are solved ;-)

Cheers, Thomas

PR #51, opened 2015-07-26
Fix resources and CLS compliance

ThomasBarnekow:
This pull request includes two fixes. Firstly, it creates a .resx file that contains the missing message strings. Secondly it does away with the 27 warnings that System.IO.Packing is not CLS compliant.

I've performed multiple commits to finally migrate to a default Resources class and related .resx file. Depending on whether or not we want to go all the way, we could only do it partially. This way, however, the solution would use the standard means of defining resources.

Together with the already merged PR #50, this PR completely fixes issue #49.


System.IO.Packaging/ContentType.cs
```diff
@@ -87,7 +88,7 @@ internal ContentType(string contentType)
             else
             {
                 if (IsLinearWhiteSpaceChar(contentType[0]) || IsLinearWhiteSpaceChar(contentType[contentType.Length - 1]))
-                    throw new ArgumentException(SR.ContentTypeCannotHaveLeadingTrailingLWS);
+                    throw new ArgumentException(Resources.ContentTypeCannotHaveLeadingTrailingLWS);
```
ThomasBarnekow: This is an example how `SR` is replaced with `Resources`.

System.IO.Packaging/ContentType.cs
```diff
@@ -43,13 +43,14 @@
...
-using System.Collections.Generic;   // For Dictionary<string, string>
-using System.Text;                  // For StringBuilder
-using System.Diagnostics;           // For Debug.Assert
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
 
 namespace System.IO.Packaging
 {
+    using Properties;
```
ThomasBarnekow: The
```
using Properties;
```
statement (with `Properties` being shorthand for `System.IO.Packaging.Properties`) must be positioned within the 
```
namespace System.IO.Packaging
``` 
declaration to make the compiler find the `System.IO.Packaging.Properties.Resources` class before it finds the `System.Resources` namespace. If we put a
```
using System.IO.Packaging.Properties;
```
statement outside of the namespace declaration (together with the other `using` statements), `Resources` would be interpreted as the `System.Resources` *namespace* rather than the `System.IO.Packaging.Properties.Resources` *class*.

System.IO.Packaging/FileFormatException.cs
```diff
@@ -27,7 +25,7 @@ public class FileFormatException : FormatException
         /// This message takes into account the current system culture.
         /// </summary>
         public FileFormatException()
```
ThomasBarnekow: Why do we implement the `System.IO.FileFormatException` at all? Isn't this already provided by other parts of the .NET Framework, also as part of `System.IO` namespace?

System.IO.Packaging/FileFormatException.cs
```diff
@@ -8,12 +8,10 @@
 //
 //---------------------------------------------------------------------------
 
-using System;
-using System.Security;
-using System.IO.Packaging;
-
 namespace System.IO
```
ThomasBarnekow: I don't think it is a good idea to declare a namespace outside of `System.IO.Packaging` or any other namespace within that (e.g., `System.IO.Packaging.Properties`).

System.IO.Packaging/FileFormatException.cs
```diff
@@ -61,8 +59,8 @@ public FileFormatException(string message, Exception innerException)
         public FileFormatException(Uri sourceUri)
             : base(
                 sourceUri == null
-                ? SR.FileFormatException
-                : SR.Format(SR.FileFormatExceptionWithFileName, sourceUri))
+                ? Resources.FileFormatException
+                : Formatter.Format(Resources.FileFormatExceptionWithFileName, sourceUri))
```
ThomasBarnekow: Regardless of my other question regarding `FileFormatException`, this is an example showing how this PR splits `SR` into `Resources` and `Formatter`. This way, the generic formatting methods are separated from the generated properties provided by the `Resources` class. This makes sense because, other than with `Settings`, the `Resources` class is not generated as a partial class.

ThomasBarnekow:
Rebased and ready to get pulled.

System.IO.Packaging/FileFormatException.cs
EricWhiteDev: System.IO.FileFormatException was part of the original WindowsBase dll, and it was in the System.IO namespace.  It would be a breaking change to have it in a different namespace, and further, we do not want to require WindowsBase.dll just for the purposes of getting that exception - it causes name conflicts with System.IO.Packaging, and there is nothing else in that DLL that is commonly used.  The option I chose was to put it into System.IO.Packaging, and in its original namespace.

System.IO.Packaging/FileFormatException.cs
EricWhiteDev: It was in WindowsBase - see my note below.

PR #54, opened 2015-08-25
Add X64 Support

EricWhiteDev:


akoeplinger:
Why is this required? The normal DocumentFormat.OpenXml.64.csproj already uses `AnyCPU` ?

EricWhiteDev:
Sorry, you are right.  This is not required.  I'll remove it.

PR #97, opened 2016-06-21
Fix unit tests on environments with non-US regional settings

MpDzik:


tomjebo:
@MpDzik merged. Thank you for the contribution!

PR #109, opened 2016-10-21
Add link to download page

smehrbrodt:


README.md
```diff
@@ -41,6 +41,11 @@ See License.txt in the project root for license information.
 Portions of this project are licensed under the MIT license.
 See MIT-License.txt in the project root for license information.
 
+Download
+========
+
+Downloads are available on [this site](https://www.microsoft.com/en-us/download/details.aspx?id=30425).
```
tomjebo: If we are going to point to the download, we should be noting that this is for the 2.5 version. We don't have a download site (separate) for the current github builds.

tomjebo:
@smehrbrodt added a comment to the commit. We should make note that the download is for 2.5. Not sure it's even needed but we can still add if you want.

smehrbrodt:
So where are the downloads of the current version?
I did this PR when I was looking for downloads of this project, but could hardly find them. Finding this Github project was easy in comparison.
So I thought it'd be good to point users to where they get binaries.

tomjebo:
The download for the latest version will be provided via a NuGet package. We are currently working on that process and will add a section to the readme which describes how to get the package.

PR #115, opened 2016-11-06
Convert the project to project.json for multitargeting

twsouthwick:
This converts to the `project.json` format. This will be replaced in Visual Studio 15 with a modified csproj format, but for now, this is the easiest way to provide multitargeting to .NET 4.5 and .NET Standard 1.3.

This switch necessitates the change from using ResX for data storage to using embedded streams.

@tomjebo


DocumentFormat.OpenXml.Tests/project.json
```diff
@@ -0,0 +1,18 @@
...
+    "dotnet-test-xunit": "2.2.0-preview2-build1029",
+    "DocumentFormat.OpenXml": "3.0.0-*"
+  },
+  "frameworks": {
+    "net451": {
+      "frameworkAssemblies": {
+        "System.Drawing": "",
+        "System.Windows.Forms": "",
+        "Microsoft.VisualBasic": ""
+      }
```
rayao: No need to add netstandard here?

PR #130, opened 2016-12-12
Update signing steps to use .NET Foundation

twsouthwick:


yingfangdu:
Hi, Does this version support mono build? I see before it can build for mono while it is is removed?

twsouthwick:
This project is now built using the DotNet CLI (see the [README](https://github.com/OfficeDev/Open-XML-SDK/blob/vNext/README.md) for more details).

The library now targets both .NET 4.5 and .NET Standard 1.3. If you are not aware of .NET Standard, please read this: https://docs.microsoft.com/en-us/dotnet/articles/standard/library. It is a much better replacement for portable libraries.

As far as the packaging library, .NET 4.5 still targets WindowsBase.dll while .NET Standard 1.3 targets System.IO.Packaging (which only supports .NET 4.6+ for the .NET Framework). They are listed as dependencies and installed via NuGet if needed. If this set up does not work on a specific version of mono, please file an issue.

yingfangdu:
Thanks Taylor, could you please give more instruction about how to make this work for Xamarin.Mac/Mono?

I am kind of confusing here. XamairnMac does not allow to use WindowsBase.dll. While I need to use this framework in our product. Before around a couple of months ago, I remembered the source code & instruction I downloaded here can build a OpenXMLLib.dll and System.IO.Packaging.dll. while at that point, it does not work as I expected. 

Today when I come back, the instruction is gone. And I have no idea how to do that for mono.

Appreciate all your help.

Thanks,
Yvonne

twsouthwick:
@yingfangdu Since this PR is not related to your question, I've created an issue to discuss this and moved the discussion there.

twsouthwick:
@tomjebo Anything blocking merging this in?

PR #153, opened 2017-01-23
fix version for automated build and change 3.0.0 to 0.0.1 placeholder

tomjebo:


CZEMacLeod:
Can you have a look at issue #200 as the packages may have the expected version but the contained DLL doesn't.

PR #156, opened 2017-01-25
Readme edits

ClareMSYanGit:


tomjebo:
@ClareMSYanGit Thanks for fixing these.

PR #161, opened 2017-01-31
fix appveyor script to build release

tomjebo:


appveyor.yml
```diff
@@ -12,10 +12,10 @@ build_script:
...
   - ps: dotnet restore
-  - ps: dotnet --verbose build .\DocumentFormat.OpenXml.Tests
+  - ps: dotnet --verbose build -c Release .\DocumentFormat.OpenXml.Tests
 
 after_build:
-  - ps: dotnet --verbose pack .\DocumentFormat.OpenXml -o .
+  - ps: dotnet --verbose pack -c Release .\DocumentFormat.OpenXml -o .
   - ps: .\build\SignPackage.ps1
 
 test_script:
```
twsouthwick: You should add the `-c Release` to the `dotnet test` command as well - otherwise it will build everything there in debug and test that

tomjebo: got it, I'll amend with the new key as well.

PR #162, opened 2017-01-31
Exclude symbols from signing and artifacts

twsouthwick:
The symbols are too large to push to myget so we are disabling them. There
is currently no option to exclude symbol package generation, so we will
manually do it in the sign script. This is fixed in preview4 of the tooling
and should simplify the sign script.

@tomjebo


twsouthwick:
Didn't find artifacts...

twsouthwick:
Looks like the symbols were found on the second push. Should be good to go

PR #164, opened 2017-02-01
updated PartUriHelper.GetNextSequenceNumber

### LINKED ISSUE #123: Different part name between Excel and Open-XML-SDK
When I call the function `WorksheetPart.AddNewPart<DrawingsPart>`, the result part filename is /xl/drawings/drawing.xml, which is different from what Excel 2013 names it (/xl/drawings/drawing1.xml). Note the "1". This wouldn't matter, **except** that the Open XML SDK 2.5 Productivity Tool for Microsoft Office won't compare DrawingsPart s created the two different ways.

What's causing this is that line 7719 of DocumentFormat.OpenXml\src\GeneratedCode\package.cs returns the TargetNameConstant which lacks the "1":
`    internal const System.String TargetNameConstant =@"drawing";`

I need to control the name of the DrawingsPart so the Productivity Tool can generate Open-XML-SDK code for me to use in my program. Can you help me? If this requires a change to the Productivity Tool, I have other ideas for enhancements too.

Howard Rubin

Josh-Curry:
Added a new list to the PartUriHelper class that contains all of the contentTypes that should have a "1" appended to the name of the first item in the package. Updated the GetNextSequenceNumber method to check the type and return a value of "1" instead of a blank string. Closes #123


.gitignore
```diff
@@ -180,3 +180,6 @@ project.lock.json
 .vs/
 Tests/
 TestFiles/
+DocumentFormat.OpenXml.Tests/Spreadsheet.xltx
```
twsouthwick: It would be better to put these in the TestFiles directory instead and then they automatically get ignored

DocumentFormat.OpenXml/src/Framework/OpenXmlPackage.cs
```diff
@@ -1384,6 +1384,56 @@ internal class PartUriHelper
             private Dictionary<string, int> _sequenceNumbers = new Dictionary<string, int>(20);
             private Dictionary<string, int> _reservedUri = new Dictionary<string, int>();
 
+            //Added for issue #123, some contentTypes need to have a 1 appended to the name
+            //for the first item in the package
+            private List<string> _numberedContentTypes = new List<string>
```
twsouthwick: For performance reasons, this should be HashSet<string> with a supplied comparer (either `StringComparer.Ordinal` or `StringComparer.OrdinalIgnoreCase` depending on if case is important). Also, make it static readonly so it's only created once for all packages

DocumentFormat.OpenXml/src/Framework/OpenXmlPackage.cs
```diff
@@ -1466,7 +1516,17 @@ private string GetNextSequenceNumber(string contentType)
                 else
                 {
                     this._sequenceNumbers.Add(contentType, 1);
-                    return "";
+                    //return "";
+
+                    //Certain types need to contain an initial 1 value. Issue #123
+                    if (_numberedContentTypes.Contains(contentType))
```
twsouthwick: For readability, you can use a ternary operator.

DocumentFormat.OpenXml/src/Framework/OpenXmlPackage.cs
```diff
@@ -1466,7 +1516,17 @@ private string GetNextSequenceNumber(string contentType)
                 else
                 {
                     this._sequenceNumbers.Add(contentType, 1);
-                    return "";
+                    //return "";
+
+                    //Certain types need to contain an initial 1 value. Issue #123
```
twsouthwick: Is there a reason you're including the issue number? I haven't seen that anywhere else in the code base

.gitignore
twsouthwick: I don't see any new tests using these. Is that expected? Or did you not add these files and the tests?

twsouthwick:
The commits should all be squashed

.gitignore
Josh-Curry: They were generated by one or more of the tests I believe.

twsouthwick: Ok - I've create #166 to track cleaning up the test infrastructure. For now, we probably don't need this change (it'll be a good impetus to fix that issue)

DocumentFormat.OpenXml/src/Framework/OpenXmlPackage.cs
twsouthwick: Also, better to just remove the return instead of commenting it out

Josh-Curry:
I've submitted the suggested changes.

twsouthwick:
LGTM. Can you squash your commits?

DocumentFormat.OpenXml/src/Framework/OpenXmlPackage.cs
```diff
@@ -1384,6 +1384,56 @@ internal class PartUriHelper
             private Dictionary<string, int> _sequenceNumbers = new Dictionary<string, int>(20);
             private Dictionary<string, int> _reservedUri = new Dictionary<string, int>();
 
+            //List of contentTypes that need to have a 1 appended to the name
+            //for the first item in the package
+            private static readonly HashSet<string> _numberedContentTypes = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
twsouthwick: Are these types from somewhere in the spec or is it an adhoc list?

Josh-Curry: They are from the ISO/IEC 29500 spec. I listed the section numbers for reference.

twsouthwick: Got it. As someone not very familiar with the spec, I would find it helpful for the comment above the hashset to indicate that the numbers in the comments in the list refer to the section from the spec

Josh-Curry:
I updated the comment regarding the ISO/IEC 29500 standard.

PR #165, opened 2017-02-02
Avoid hard coded path separators in the SaveAndClone/FlatOpc tests

Numpsy:
needed to get the tests to work on OSX

### DIFF
--- a/DocumentFormat.OpenXml.Tests/FlatOpcFixture.cs
+++ b/DocumentFormat.OpenXml.Tests/FlatOpcFixture.cs
@@ -44,13 +44,13 @@ namespace DocumentFormat.OpenXml.Tests
 
         private const string DirectoryPath = "FlatOpc";
 
-        private const string DocumentPath = DirectoryPath + "\\Document.docx";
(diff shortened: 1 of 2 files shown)


DocumentFormat.OpenXml.Tests/FlatOpcFixture.cs
```diff
@@ -44,13 +44,13 @@ private static string TestResultsDirectory
 
         private const string DirectoryPath = "FlatOpc";
 
-        private const string DocumentPath = DirectoryPath + "\\Document.docx";
-        private const string PresentationPath = DirectoryPath + "\\Presentation.pptx";
-        private const string SpreadsheetPath = DirectoryPath + "\\Spreadsheet.xlsx";
+        private readonly string DocumentPath = DirectoryPath + Path.DirectorySeparatorChar + "Document.docx";
```
twsouthwick: Can you switch these to Path.Combine instead? They can also be made static

Numpsy: ok, i'll change it tomorrow.

PR #170, opened 2017-02-09
Revert to previous behavior for part numbering.

### LINKED ISSUE #169: Why are package items numbered using Hex format?
I came across these lines of code while fixing issue #123.

```
// Let's use the number string in hex
return Convert.ToString(this._sequenceNumbers[contentType], 16);
```

I believe that once 10 or more items of the same contentType, such as an Excel worksheet, are added to the package we will see the same problem as in issue #123 where comparisons using the Productivity Tool will not work correctly after the file is resaved by Excel. I will investigate and make changes as necessary.

Josh-Curry:
Reverted to previous behavior where we use base10 numbering format for parts in the package. closes #169

### DIFF
--- a/DocumentFormat.OpenXml/src/Framework/OpenXmlPackage.cs
+++ b/DocumentFormat.OpenXml/src/Framework/OpenXmlPackage.cs
@@ -1507,10 +1507,7 @@ namespace DocumentFormat.OpenXml.Packaging
                 {
                     this._sequenceNumbers[contentType] += 1;
                     // use the default read-only NumberFormatInfo that is culture-independent (invariant). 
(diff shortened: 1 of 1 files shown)


twsouthwick:
LGTM - although it doesn't make sense to have two commits for this. Try squashing your commits.

twsouthwick:
Does the spec say anything about hex vs decimal for the numbering format? What does Word/Excel/etc do?

Josh-Curry:
The spec doesn't say anything about how the items should be numbered. The Office products use standard numbering. I can't find any reason why it was switched to hex.

PR #176, opened 2017-03-15
Merge test fix from master

twsouthwick:
The PR for this had been against master so the changes aren't in the current branch.


twsouthwick:
@tomjebo

PR #182, opened 2017-03-17
Fix some project and schema issues

twsouthwick:
This fixes a couple of project structure issues introduced with VS17, as well as fixes the schema source generator to not have a warning and to run correctly as a signed assembly.


twsouthwick:
I'm going to roll this into a more general change.

Issue #174, opened 2017-03-10
WindowsBase crashes when feeding data


MarcoMartins86:
This happens every time, with all versions that are dependent of WindowsBase.dll (2.5, 2.7 not 2.6 since uses independent implementation of System.IO.Packaging).
When feeding a huge amount of data (in my case 137MB of valid XML content into a worksheet part),  throws a System.IO.IsolatedStorage.IsolatedStorageException ("Unable to determine the identity of domain.") 

at System.IO.IsolatedStorage.IsolatedStorage._GetAccountingInfo(Evidence evidence, Type evidenceType, IsolatedStorageScope fAssmDomApp, Object& oNormalized)
   at System.IO.IsolatedStorage.IsolatedStorage.GetAccountingInfo(Evidence evidence, Type evidenceType, IsolatedStorageScope fAssmDomApp, String& typeName, String& instanceName)
   at System.IO.IsolatedStorage.IsolatedStorage._InitStore(IsolatedStorageScope scope, Evidence domainEv, Type domainEvidenceType, Evidence assemEv, Type assemblyEvidenceType, Evidence appEv, Type appEvidenceType)
   at System.IO.IsolatedStorage.IsolatedStorage.InitStore(IsolatedStorageScope scope, Type domainEvidenceType, Type assemblyEvidenceType)
   at System.IO.IsolatedStorage.IsolatedStorageFile.GetStore(IsolatedStorageScope scope, Type domainEvidenceType, Type assemblyEvidenceType)
   at MS.Internal.IO.Packaging.PackagingUtilities.ReliableIsolatedStorageFileFolder.GetCurrentStore()
   at MS.Internal.IO.Packaging.PackagingUtilities.ReliableIsolatedStorageFileFolder..ctor()
   at MS.Internal.IO.Packaging.PackagingUtilities.GetDefaultIsolatedStorageFile()
   at MS.Internal.IO.Packaging.PackagingUtilities.CreateUserScopedIsolatedStorageFileStreamWithRandomName(Int32 retryCount, String& fileName)
   at MS.Internal.IO.Packaging.SparseMemoryStream.EnsureIsolatedStoreStream()
   at MS.Internal.IO.Packaging.SparseMemoryStream.SwitchModeIfNecessary()
   at MS.Internal.IO.Packaging.SparseMemoryStream.Write(Byte[] buffer, Int32 offset, Int32 count)
[...]

PR #185, opened 2017-03-18
Add support for .NET 3.5, .NET 4.0, and .NET 4.6

### LINKED ISSUE #180: Add support for .NET 3.5
It appears that ClosedXML is holding off on taking a dependency on the latest versions because support for .NET 3.5 is dropped (see https://github.com/ClosedXML/ClosedXML/issues/225). Is there interest for a .NET 3.5 build of the library?

twsouthwick:
This adds target frameworks for more .NET Framework versions, including
moving to .NET 4.6 instead of .NET 4.5. That way, .NET 4.6 and later can
use System.IO.Packaging, while frameworks below that will have access to
the library, but will use WindowsBase.

Fixes #180 and #174


twsouthwick:
Looks like the `dotnet` tool doesn't understand .NET 3.5 reference assemblies. I'll look into switch things to msbuild as that does understand it.  See https://github.com/Microsoft/msbuild/issues/1333

ThomasBarnekow:
I assume that does not impact people targeting 4.5.x in any way, right?

On a somewhat related note, does anybody know whether there are .NET Framework "deployment statistics". For example, the question is what percentage of corporate desktop PCs have .NET Framework 4.6.x installed.

Google provides information regarding the percentage of Android devices using which Android release, so you can decide which one you would want to target in order to capture whatever percentage of the market.

twsouthwick:
@ThomasBarnekow, this won't affect 4.5.x in any way. However, if you target 4.6, you'll get the updated `System.IO.Packaging` that has a few fixes that are not in the `WindowsBase` version of the packaging tools.

HollyAM:
@ThomasBarnekow, 
.NET 4.6 or a greater version is present on all win10 machines, so you can count on them being on over 300 Million active devices: https://blogs.windows.com/windowsexperience/2016/05/05/windows-10-now-on-300-million-active-devices-free-upgrade-offer-to-end-soon/

They are also present on other OSes windows 7 or greater. We don't typically make those numbers public, but if you can explain your need more specifically, I might be able to provide more details.

ThomasBarnekow:
@HollyAM, thanks for responding. I will contact you by email to explain what I would need to know.

twsouthwick:
@tomjebo This is ready to review. Thanks!

tomjebo:
@twsouthwick sorry I didn't get to this until now. I decided to leave the two commits separate as they seemed to be logically independent enough. Hope that's ok with you. Thanks for straightening all of this out.

Issue #189, opened 2017-03-21
Update the Readme.md to correctly reference VS 2015 


tomjebo:
Currently, Build Instructions says:

...
To build the Open XML SDK:

Clone the repo at https://github.com/OfficeDev/Open-XML-SDK
Open the solution using Visual Studio 2015. Community Edition of VS2015 works just fine.

That last line should say 2017, as we now have a requirement for 2017.

tarunchopra:
This table too needs an update:

Table 1: The latest builds are available via a MyGet feed
Client
Feed URL
NuGet V3 (Visual Studio 2015+)
https://dotnet.myget.org/F/open-xml-sdk/api/v3/index.json

NuGet V2 (Visual Studio 2012+)
https://dotnet.myget.org/F/open-xml-sdk/api/v2

tomjebo:
@tarunchopra I think for using the library it's ok

PR #190, opened 2017-03-21
changed 2015 to 2017 for VS build instructions

tomjebo:
We should at least add this change in the instructions. should we also add a note for the VS requirement now?

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ This project uses the `project.json` for the project build system. Please see th
 
 To build the Open XML SDK:
 - Clone the repo at https://github.com/OfficeDev/Open-XML-SDK
-- Open the solution using Visual Studio 2015.  Community Edition of VS2015 works just fine.
+- Open the solution using Visual Studio 2017.  Community Edition of VS2017 works just fine.
 - Build the solution (using either Debug or Release configuration)
 - Run the Xunit tests to verify the installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,7 +109,7 @@ This project uses the `project.json` for the project build system. Please see th
 
 To build the Open XML SDK:
 - Clone the repo at https://github.com/OfficeDev/Open-XML-SDK
-- Open the solution using Visual Studio 2015.  Community Edition of VS2015 works just fine.
+- Open the solution using Visual Studio 2017.  Community Edition of VS2017 works just fine.
```
twsouthwick: I would add that they will install side by side with previous versions of VS and provide a link

tomjebo:
This addresses issue #189

README.md
tomjebo: would this link be right: [https://msdn.microsoft.com/en-us/library/ms246609.aspx](url)?

twsouthwick: VS 2017 is actually not the only supported tool. I'd rephrase along the lines of "Open the solution with an editor that supports the latest .NET project files, such as VS 2017, VS Code, VS for Mac, or .NET CLI. See https://www.microsoft.com/net/download/core for details."

twsouthwick:
fyi: if you change your initial post on this PR to have the line "Fixes #189", that issue will automatically close when this is merged.

tomjebo:
thanks, that's what I was trying to do but thought it would close the issue when I commented. will do it next time. And I'll squash these two commits with the merge.

twsouthwick:
You can edit the initial post to include it

Issue #186, opened 2017-03-19
Unit tests take too long to run


ThomasBarnekow:
I've seen the following statement in the release notes for Version 2.6.1:

> Added hundreds of XUnit tests. There are now a total of 1333 tests. They take about 20 minutes to run, so be patient.

20 minutes for running "unit tests" is clearly too long and defeats the purpose. Who will make a code change and then run those unit tests, possibly even multiple times?

Mind you I have also written a few of those unit tests. Those tests, like many others, access the file system. Like any other persistence operation, we should avoid this in "real" unit tests.

For example, in recent unit tests that I've written for a file service client library, which provides an abstraction of the local file system, OneDrive for Business, and SharePoint Online at the moment, I've used the [Moq](https://github.com/moq/moq4) and [System.IO.Abstractions.TestingHelpers](https://www.nuget.org/packages/System.IO.Abstractions.TestingHelpers/) libraries for mocking to avoid file and network I/O. Many of those unit tests run in the order of milliseconds (and a good number in just 1 millisecond) or tens of milliseconds.

I've also produced "integration tests" which access OneDrive for Business and SharePoint Online. These take a couple of hundred milliseconds (e.g., up to 700-900 milliseconds) to run. So, just because those "integration tests" perform actual I/O, they are two orders of magnitude -- or up to a factor of 900 in my example -- slower than the "real" unit tests.

20 minutes for 1.333 tests means 900 milliseconds on average per test. Based on my example above, we might be able to reduce this by a factor of 10 to 100. One approach might be to:

- classify tests, identifying those that really require file I/O (the "**Integration Tests**") vs. those that don't, even if they currently access the file system unnecessarily (the "**Unit Tests**");
- have two test projects, one for Integration Tests and one for Unit Tests;
- move the tests to their respective projects; and
- fix the Unit Tests to improve their performance.

Check out the [System.IO.Abstractions](https://github.com/tathamoddie/System.IO.Abstractions) for a library that provides an injectable and testable API that otherwise provides the same functionality as the System.IO namespace. I've used this library to implement my local file service client, because that turns it into a library which, through injection by way of an IoC container (e.g., [Autofac)](https://github.com/autofac/Autofac), can be unit tested without actually accessing the local file system.

twsouthwick:
This is a better description of what #166 was trying to define :)

twsouthwick:
[...]

PR #228, opened 2017-08-17
Consolidate OpenXmlDomTestExtensions

twsouthwick:
This is a start at addressing #186. This consolidates some of the main code used to open files - it was previously spread across two files which were virtually identical, and had a massive amount of duplicated code.


Shanghai2017:
lgtm

PR #233, opened 2017-08-18
Merge vNext into Office2016 branch

twsouthwick:


twsouthwick:
@tomjebo We don't want to squash and merge this branch - we want to keep the history since this is a feature branch

PR #263, opened 2017-10-16
Clean up description and add summary

### LINKED ISSUE #257: "High fidelity conversion of DOCX => HTML/CSS and HTML/CSS => DOCX."
The Readme file mentions that is possible to convert docx to HTML, but I didn't find any example or documentation explaining how to do it. Can anyone provide an example? I have only found using https://github.com/OfficeDev/Open-Xml-PowerTools, but due to some restrictions, I can't use this library.

Thanks.

twsouthwick:
The lack of a summary field in the NuGet package means the description gets used, but it ends up being truncated. 

This change also cleans up the description to be a bit clearer:

- The description also includes information that is not about this project, but the OpenXml PowerTools, so I've removed that with this change as well.
- DOCX, XLSX, and PPTX usage is replaced with Word, Excel, and PowerPoint

Fixes #257


twsouthwick:
@tomjebo can you take a look at this?

twsouthwick:
I'll plan on merging this in by the EOD if no one objects

PR #276, opened 2017-10-30
Refactor initial markup compatibility tests to decrease unit test duration

twsouthwick:
This change fixes the test cases that take the most amount of time, shaving about 30% of the unit test duration. Subsequent PRs will address the rest.


twsouthwick:
This was part of #186

PR #290, opened 2017-11-12
Allow saving when autosave is false

### LINKED ISSUE #262: Explicit call to Save() does nothing when AutoSave is set to false
In current implementation explicitly calling Save() from OpenXmlPackage does nothing when AutoSave option is set to false.
It's because Save() calls SavePartContents() which checks if AutoSave is enabled.

This makes sense when SavePartContents() is called from Dispose() method, because we shouldn't save automatically during disposal when this flag is set to false, but shouldn't explicit call to Save() actually save package regardless of AutoSave state?

twsouthwick:
Currently, if a document is opened with `AutoSave = false`, then calls to `.Save()` will fail. This change allows packages that are opened with `AutoSave = false` to be manually saved if desired, while maintaining previous behavior when `AutoSave = true`.

Fixes #262

### DIFF
--- a/DocumentFormat.OpenXml.Tests/SaveAndCloneTests.cs
+++ b/DocumentFormat.OpenXml.Tests/SaveAndCloneTests.cs
@@ -269,6 +269,33 @@ namespace DocumentFormat.OpenXml.Tests
         }
 
         [Fact]
+        public void CanSaveAutosaveFalse()
+        {
+            using (var stream = new MemoryStream())
+            {
+                using (var package = WordprocessingDocument.Create(stream, WordprocessingDocumentType.Document))
+                {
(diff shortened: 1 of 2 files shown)


twsouthwick:
@tarunchopra Can you take a look at this?

twsouthwick:
@tomjebo @tarunchopra Unless I hear an objection, I'll merge this in tomorrow.

tarunchopra:
Sure Taylor. Thx for the help.

twsouthwick:
@tarunchopra I'm assuming that that was a sign off. If so, can you click the button that you approve the change?

twsouthwick:
I'm going to go ahead and merge this. In the future, please approve or block :)

PR #295, opened 2017-11-15
[WIP] Flush package when save is called

### LINKED ISSUE #294: Calling Save() doesn't flush to the stream
**Description**

I am writing a wrapper class for SpreadsheetDocument. **I would like to be able to Save the document more than once**. Currently the only way to force SpreadsheetDocument to flush to the stream is by calling Close() on it. Is there anyway to avoid that?

**Information**

- .NET Target: .Net Standard 2.0
- DocumentFormat.OpenXml Version: 2.7.2

**Repro**

```csharp
using DocumentFormat.OpenXml;
using DocumentFormat.OpenXml.Spreadsheet;
using System;
using System.IO;

namespace OpenXmlDemo
{
    class Program
    {
        {
            using (var stream = new MemoryStream())
            {
                var document = SpreadsheetDocument.Create(stream: stream,
                                                          type: SpreadsheetDocumentType.Workbook,
                                                          autoSave: true);

                var workbookPart = document.AddWorkbookPart();
                var workbook = workbookPart.Workbook = new Workbook();
                var sheetsElement = new Sheets();
                workbook.AppendChild(sheetsElement);

                InsertSheet(document, workbookPart, sheetsElement, "Sheet A", 1);

                // Save the workbook and document
                workbook.Save();
                document.Save();

                // document.Close();

                // Returns an empty array if the document isn't closed
                // stream.Position = 0; // Some places suggested this, unfortunately this doesn't seem to help.
                var bytes = stream.ToArray(); 
                File.WriteAllBytes(@"D:\sheet.xlsx", bytes);
            }
        }

        private static Sheet InsertSheet(SpreadsheetDocument doc,
                                      WorkbookPart workbookpart,
                                      Sheets sheetsElement,
                                      string name,
                                      uint id)
        {
            WorksheetPart newWorksheetPart = workbookpart.AddNewPart<WorksheetPart>();
            newWorksheetPart.Worksheet = new Worksheet(new SheetData());

            string relationshipId = workbookpart.GetIdOfPart(newWorksheetPart);

            // Give the new worksheet a name.
            if (string.IsNullOrEmpty(name))
                name = "Sheet" + id;

            // Append the new worksheet and associate it with the workbook.
            Sheet sheet = new Sheet() { Id = relationshipId, SheetId = id, Name = name };
            sheetsElement.AppendChild(sheet);

            return sheet;
        }
    }
}
```

**Observed**

The stream is empty.

**Expected**

The stream should contain the document.

twsouthwick:
When Save is called, the underlying package is currently not flushed. This makes Save act in a weird way where you expect that the data is persisted, but it has not been. This change fixes that so saving flushes the package to whatever data source it is from. This does not, however, remove any unused data parts, which will still only be removed on close.

Fixes #294

### DIFF
--- a/DocumentFormat.OpenXml.Tests/SaveAndCloneTests.cs
+++ b/DocumentFormat.OpenXml.Tests/SaveAndCloneTests.cs
@@ -302,6 +302,36 @@ namespace DocumentFormat.OpenXml.Tests
         }
 
         [Fact]
+        public void CanSaveWithOutClosing()
+        {
+            byte[] GetNewSpreadsheet()
+            {
+                using (var stream = new MemoryStream())
+                using (var source = SpreadsheetDocument.Create(stream, SpreadsheetDocumentType.Workbook))
+                {
+                    source.AddWorkbookPart();
+                    source.WorkbookPart.Workbook = new Spreadsheet.Workbook();
+                    source.WorkbookPart.Workbook.AppendChild(new Spreadsheet.Sheets());
+                    source.Save();
+                    source.WorkbookPart.Workbook.AppendChild(new Spreadsheet.Sheets());
+
+                    Assert.Equal(2, source.WorkbookPart.Workbook.ChildElements.Count);
+
+                    return stream.ToArray();
+                }
+            }
+
+            var bytes = GetNewSpreadsheet();
+            Assert.NotEmpty(bytes);
+
+            using (var stream = new MemoryStream(bytes))
+            using (var source = SpreadsheetDocument.Open(stream, false))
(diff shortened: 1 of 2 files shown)


twsouthwick:
@tarunchopra Can you take a look at this?

twsouthwick:
I'm not sure if we should delete unused data parts on close. Any thoughts?

twsouthwick:
Hmm appears that it is not working on .NET Core

twsouthwick:
I'm going to close this now as we are blocked on .NET Core support

PR #301, opened 2017-11-22
Use a consolidated serializer for schema constraint data on all platforms

twsouthwick:
Currently, there are two ways that schema constraint data is stored:

- On .NET Framework, there is a BinaryFormatter serialized set of data
- On .NET Standard, there are a few large files that contain C# code derived from the BinaryFormatter data

The problem with this is that BinaryFormatter is not a good serialization technique for cross-runtime/platform usage and having two systems is really complicated. The second system was put in place due to not having access to the original BinaryFormatter generator. This is available now, and so this change consolidates the two systems into one using the DataContractSerializer.

Further changes may remove the binary aspect of the files so that diffs can be better generated and easily viewed.


twsouthwick:
@tomjebo Can you take a look at this?

twsouthwick:
@tomjebo I'd like to merge this in and update the Office2016 branch accordingly. If I don't hear otherwise, I'll plan on merging this tomorrow.

PR #304, opened 2017-11-28
Add CellValue overload to accept DateTime or DateTimeOffset

### LINKED ISSUE #142: Adding dates to cells is not easily discoverable
Using the following example to determine cell value:

        private static CellValues GetDataType( object DataColumn )
        {
            CellValues cvs = new CellValues();

            if( DataColumn is int || DataColumn is short || DataColumn is long ||
                DataColumn is float || DataColumn is double || DataColumn is uint ||
                DataColumn is ulong || DataColumn is ushort || DataColumn is decimal)
            {
                cvs = CellValues.Number;
            }
            else if (DataColumn is DateTime)
            {
                cvs = CellValues.Date;
            }
            else
            {
                cvs = CellValues.String;
            }

            return cvs;
        }

Any time I use CellValues.Date spreadsheet needs repair. Everything displays correctly but ideally I would like it not to need repairing on opening.

Allowing DateTime to fall into the CellValues.String also works.

I don't know if it's a bug or not, so I am just asking if it's a valid cell value.

twsouthwick:
A common need is to add a date object to a cell, but it requires a specific format that is not easy to track down. The text needs to be in ISO 8601 format, so this adds a couple of overloads to do that so it's more discoverable.

Fixes #142


twsouthwick:
@tomjebo Is this correct? I'm not certain of the spec (or where to even look), but I found a SO post about it: https://stackoverflow.com/questions/114983/given-a-datetime-object-how-do-i-get-an-iso-8601-date-in-string-format

PR #307, opened 2017-11-30
Use SignService v2

onovotny:
This PR is to move to the new sign service. Please contact me on how to share the secrets with you.

These changes should be cherry-picked & applied to all branches you plan to build from as we want to deprecate the old service.

Thanks

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,9 +3,9 @@ os: Visual Studio 2017
 environment:
   image: Visual Studio 2017
   SignClientSecret:
-    secure: <add value here>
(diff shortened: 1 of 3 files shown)


twsouthwick:
Thanks @onovotny! I'll pm you to get the secrets so we can get this merged in.

twsouthwick:
I'm going to merge this without the appveyor build (can't get it to respond), but these changes will only be seen on a branch we have secrets enabled anyway.

onovotny:
We'll be decommissioning the old service at some point in the next month or two, so if there's other branches you plan on building from, these changes should be put there too. LMK if you run into any issues. Thx!

twsouthwick:
Sounds good. There are a couple others, so we'll get those merged into those as well

PR #308, opened 2017-11-30
Add code coverage to test runs

twsouthwick:


twsouthwick:
Visual Studio 2017 15.5 seems to have fixed some weirdness in vstest.console.exe, so if that is installed, this new target and build script works. AppVeyor is not on 15.5 yet, so we are blocked by https://github.com/appveyor/ci/issues/1963

Issue #315, opened 2017-12-08
Organize code to follow namespace hierarchy


twsouthwick:
It is very difficult to find code and namespaces are mixed around even in the same file. We should start organizing the code so the folder structure follows the namespaces and only have a single class in a file.

twsouthwick:
This will be tracked under #197

PR #322, opened 2017-12-12
Move packaging classes to own files

twsouthwick:
Many of the files have 5+ classes in them (and some have many more). This makes it very difficult to find classes and makes for very long, unmaintainable files. This is part of #315 and starts with bringing all the classes within the `DocumentFormat.OpenXml.Packaging` namespace into the folder, with each class being in their own file.


twsouthwick:
@tomjebo @tarunchopra  I've started classes into their own files based off of types and folders based off of namespace.

PR #323, opened 2017-12-12
Separate sources1.cs to own files

twsouthwick:
This is a part of #315 and moves classes from sources1.cs to their own files. This file had around multiple namespaces and classes which are now located in folders based on their namespace and files based on their type name.


twsouthwick:
@tomjebo @tarunchopra  I've started classes into their own files based off of types and folders based off of namespace.

PR #342, opened 2017-12-20
East Asia is better term

davidchesnut:


twsouthwick:
@davidchesnut You created the PR against master which was quite far behind vNext (the current dev branch). We are, however, in the process of moving to committing directly to master, so master is now up to date. However, your change must be updated before it can be merged.

davidchesnut:
I'll resubmit on the vNext branch.

twsouthwick:
@davidchesnut You actually submitted as I was switching our branching strategy. The vNext branch is no more. You can reopen this one and submit against master.

twsouthwick:
Thanks! LGTM. We'll have to update the generator for this as well, but this will fix it for now.

PR #362, opened 2017-12-30
Add initial OfficeSchemaProcessor

twsouthwick:
This handles packages.xsd transformations into C#. This is from the ooxml schema project and is part of an initial effort to bring that processing into the open SDK. This is a rewrite of it, but generates the same logical code, but in a much cleaner way. A subsequent PR will push updated generated code.

@tomjebo This change will also make it much easier to support Office 2016+ with minimal code churn (I've tested it with the Office 2016 packages.xsd and the changes are simple and easy to understand).


twsouthwick:
To see the updated files, see the code in this folder: https://github.com/twsouthwick/Open-XML-SDK/tree/updated-packing-code/DocumentFormat.OpenXml/GeneratedCode/Packaging. Instead of a single `packages.g.cs`, this generates a file for each class for easy finding.

PR #375, opened 2018-01-05
Keep a single centralized changelog

twsouthwick:
@tomjebo Here's some of that "MSBuild magic" to keep a single changelog instead of multiple.


twsouthwick:
Yeah that would be nice

PR #382, opened 2018-01-17
tests for #380

daniel-white:
a few skipped tests for #380 to illustrate the problem and have some coverage when the document is dynamically generated.

Issue #380, opened 2018-01-16
ToFlatOpcString fails with generated document


daniel-white:
**Description**

`ToFlatOpcString` throws "Root element is missing." I'm trying to put this format into a VSTO add-in. When I enumerate `wordprocessingDocument.Package.GetParts()`, I get the relationship part back as an empty string (`ContentType = "application/vnd.openxmlformats-package.relationships+xml"`). Shouldn't `AddMainDocumentPart` take care of this under the covers?

**Information**

- .NET Target: .NET Core 2.0, .NET Framework 4.7
- DocumentFormat.OpenXml Version: 2.8.1

**Repro**

```csharp
using System;
using System.IO;
using DocumentFormat.OpenXml;
using DocumentFormat.OpenXml.Packaging;
using DocumentFormat.OpenXml.Wordprocessing;

namespace ConsoleApp1
{
    class Program
    {
        static void Main(string[] args)
        {
            using (MemoryStream stream = new MemoryStream())
            using (WordprocessingDocument wordprocessingDocument = WordprocessingDocument.Create(stream, WordprocessingDocumentType.Document))
            {
                MainDocumentPart mainDocumentPart = wordprocessingDocument.AddMainDocumentPart();
                Body body = new Body(new Paragraph(new Run(new RunProperties(new Bold()),
                    new Text("Hello world this should be bold"))));
                mainDocumentPart.Document = new Document(body);

                wordprocessingDocument.Save();

                Console.WriteLine(wordprocessingDocument.ToFlatOpcDocument());
            }

        }
    }
}
``` 

**Observed**

Exception:

```
System.Xml.XmlException: Root element is missing.
   at System.Xml.XmlTextReaderImpl.Throw(Exception e)
   at System.Xml.XmlTextReaderImpl.ParseDocumentContent()
   at System.Xml.XmlTextReaderImpl.Read()
   at System.Xml.XmlReader.MoveToContent()
   at System.Xml.Linq.XElement.Load(XmlReader reader, LoadOptions options)
   at DocumentFormat.OpenXml.Packaging.OpenXmlPackage.GetContentsAsXml(PackagePart part)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at System.Xml.Linq.XContainer.AddContentSkipNotify(Object content)
   at System.Xml.Linq.XContainer.AddContentSkipNotify(Object content)
[...]

PR #383, opened 2018-01-17
Add tests for flatopc without flush

tomjebo:
@daniel-white I added these from your previous PR. Seems they should go into this FlatOpcTests.cs.


tomjebo:
@daniel-white I opened this PR and added the code in a rush but don't want to take your credit. Would you like to fix up your PR #382 so it can be merged?

daniel-white:
@tomjebo go ahead with this one. i appreciate you taking the time to fix up my silly mistake

DocumentFormat.OpenXml.Tests/FlatOpcTests.cs
```diff
@@ -277,6 +294,21 @@ public void CanCreateFlatOpcSpreadsheetDocuments8()
...
+        {
+            using (MemoryStream stream = new MemoryStream())
+            using (PresentationDocument presentation = PresentationDocument.Create(stream, PresentationDocumentType.Presentation))
+            {
+                PresentationPart presentationPart = presentation.AddPresentationPart();
+                presentationPart.Presentation = new Presentation.Presentation();
+
+                presentation.Save();
+
+                string opc = presentation.ToFlatOpcString();
```
twsouthwick: Can you add an assert as to what is expected here?

twsouthwick: Same thing for the other tests

DocumentFormat.OpenXml.Tests/FlatOpcTests.cs
```diff
@@ -145,6 +147,28 @@ public void CanCreateFlatOpcPresentationDocuments8()
...
+                    WorkbookPart workbookPart = spreadsheet.AddWorkbookPart();
+                    workbookPart.Workbook = new Workbook();
+
+                    spreadsheet.Save();
+
+                    string opc = spreadsheet.ToFlatOpcString();
+                }
+                catch (System.Exception e)
+                {
+                    Assert.True(false, e.Message);
```
twsouthwick: What's this supposed to test? Is it expected to throw an exception? If so, you want to use the pattern:

```csharp
Assert.Throws<Exception>(()=> [expression]);
```

However, I don't think that's the intended behavior - that's just what happens right now. Can you find the expected value here and assert that the contents are that? I'm assuming you can use the work around described in the issue to save the document, reopen the document, and call `ToFlatOpcString()`

DocumentFormat.OpenXml.Tests/FlatOpcTests.cs
```diff
@@ -145,6 +147,32 @@ public void CanCreateFlatOpcPresentationDocuments8()
             }
         }
 
+        [Fact(Skip = "#380")]
+        public void CanCreateFlatOpcPresentationDocumentsDynamically()
+        {
+            using (MemoryStream stream = new MemoryStream())
+            using (PresentationDocument presentation = PresentationDocument.Create(stream, PresentationDocumentType.Presentation))
+            {
+                try
```
twsouthwick: Can you remove the try/catch. It's okay if these tests fail (in fact they should) since we have the skip attribute on them. It's not expected that they throw exceptions

DocumentFormat.OpenXml.Tests/FlatOpcTests.cs
```diff
@@ -145,6 +147,25 @@ public void CanCreateFlatOpcPresentationDocuments8()
             }
         }
 
+        [Fact(Skip = "#380")]
+        public void CanCreateFlatOpcPresentationDocumentsDynamically()
+        {
+            using (MemoryStream stream = new MemoryStream())
+            using (PresentationDocument presentation = PresentationDocument.Create(stream, PresentationDocumentType.Presentation))
+            {
+                    PresentationPart presentationPart = presentation.AddPresentationPart();
```
twsouthwick: nit: looks like the whitespace is off. Did you format the doc before committing? ctrl+k+d will do that in VS.

tomjebo: I didn't. I'll do it.

twsouthwick:
This adds tests for #380

PR #390, opened 2018-01-25
Create single ValidationContext and remove mutability from validators

twsouthwick:
This simplifies some of the validation logic and removes more mutability. There's probably some more that can be addressed later, but this creates a single `ValidationContext` to track errors and whether or not validation is cancelled. By doing this, a number of eventing patterns can be removed, which should help a bit with performance on large validation scenarios.


twsouthwick:
This is a continuation of #388

Issue #197, opened 2017-04-14
Standardize on style for code base


twsouthwick:
Style feels inconsistent in the code base (or at least not what I'm used to). In either case, it would be good to define a set of conventions for what styles are, and then use StyleCop (I find the analyzers to work well) and editor config to enforce some of these (which is available in VS2017 and other editors as well).

I generally recommend projects follow the style used by CoreFX (https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md), but we can tweak it for this project if people have strong opinions one way or another.

Thoughts?

/cc @tomjebo @ThomasBarnekow 

Here's a list of analyzers I'm planning on turning on (subject to change). The text following the ID is what will trigger the analyzer. For details, look up the documentation at https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/Master/documentation

 ## Maintainability

- [x] **SA1119** A C# statement contains parenthesis which are unnecessary and should be removed.
- [x] **SA1400** The access modifier for a C# element has not been explicitly defined.
- [x] **SA1402** A C# code file contains more than one unique type.
- [x] **SA1403** A C# code file contains more than one namespace.
- [x] **SA1404** A Code Analysis SuppressMessage attribute does not include a justification.
- [ ] ~~**SA1405** A call to Debug.Assert in C# code does not include a descriptive message.~~
- [x] **SA1406** A call to Debug.Fail in C# code does not include a descriptive message.
- [x] **SA1410** A call to a C# anonymous method does not contain any method parameters, yet the statement still includes parenthesis.
- [x] **SA1411** An attribute declaration does not contain any parameters, yet it still includes parenthesis.
- [x] **SA1412** The encoding of the file is not UTF-8 with byte order mark.
- [x] **SA1413** A multi-line initializer should use a comma on the last item.

## Readability

- [x] **SA1100** A call to a member from an inherited class begins with base., and the local class does not contain an override or implementation of the member.
- [x] **SA1106** The C# code contains an extra semicolon.
- [x] **SA1107** The C# code contains more than one statement on a single line.
- [x] **SA1108** A C# statement contains a comment between the declaration of the statement and the opening brace of the statement. 
- [x] **SA1113** A comma between two parameters in a call to a C# method or indexer, or in the declaration of a method or indexer, is not placed on the same line as the previous parameter.
[...]

PR #464, opened 2018-04-05
Use string.Empty instead of ""

twsouthwick:
Part of #197


src/DocumentFormat.OpenXml/Attributes/Translator/DocumentTagAttributeTranslator.cs
```diff
@@ -8,7 +8,7 @@ internal class DocumentTagAttributeTranslator : TagAttributeTranslator
         private static string[] arrayOfOrigTagAttrNames = { "conformance" };
         private static string[] arrayOfNewTagAttrNames = { "conformance" };
         private static string[] arrayOfOrigTagAttrValues = { "strict" };
-        private static string[] arrayOfNewTagAttrValues = { "" }; // Must be "".
+        private static string[] arrayOfNewTagAttrValues = { string.Empty }; // Must be "".
```
daniel-white: can the comment be updated for clarity?

twsouthwick: ... I have no idea why it 'must be ""'... I saw it while doing this change and thought about removing it because it doesn't seem to add any value

twsouthwick: ... I have no idea why it 'must be ""'... I saw it while doing this change and thought about removing it because it doesn't seem to add any value

twsouthwick: I've just removed it. None of the other items have a comment about why it is necessary, so I'm under the assumption that they are all necessary.

Issue #485, opened 2018-04-14
OpenXmlSimpleType children do not implement equality


ThomasBarnekow:
**Description**

Classes like `StringValue` and `HexBinaryValue` do not implement equality and you must use their `Value` properties to compare them. This is not handy and can lead to unexpected results.

**Information**

- .NET Target: Any
- DocumentFormat.OpenXml Version: 2.8.1

**Repro**

```csharp
using System.Collections.Generic;
using System.Linq;
using DocumentFormat.OpenXml;
using Xunit;

namespace DocumantFormat.OpenXml.Tests
{
    public class OpenXmlSimpleValueTests
    {
        private const string HexBinary = "12345678";
        private const string Value = "Hello World";

        [Fact]
        public void HexBinaryValue_ValuesAreEqual_HexBinaryValuesAreNotEqual()
        {
            var hexBinaryValue1 = new HexBinaryValue(HexBinary);
            var hexBinaryValue2 = new HexBinaryValue(HexBinary);

            Assert.Equal(hexBinaryValue1.Value, hexBinaryValue2.Value);

            // Those two HexBinaryValue instances should be deemed equal.
            Assert.NotEqual(hexBinaryValue1, hexBinaryValue2);
        }

        [Fact]
        public void HexBinaryValue_ValueIsContainedInList_ListDoesNotContainHexBinaryValue()
        {
            var hexBinaryValues = new List<HexBinaryValue>
            {
                new HexBinaryValue(HexBinary),
                new HexBinaryValue("ABCDEF00")
            };

            var hexBinaryValue = new HexBinaryValue(HexBinary);

            Assert.Contains(hexBinaryValue.Value, hexBinaryValues.Select(hbv => hbv.Value));

            // The HexBinaryValue should be deemed contained in the list.
            // The fact that this wasn't the case was a gotcha.
            Assert.DoesNotContain(hexBinaryValue, hexBinaryValues);
        }

        [Fact]
        public void StringValue_ValuesAreEqual_StringValuesAreNotEqual()
        {
            var stringValue1 = new StringValue(Value);
            var stringValue2 = new StringValue(Value);

            Assert.Equal(stringValue1.Value, stringValue2.Value);

            // The two StringValue instances should be deemed equal.
            Assert.NotEqual(stringValue1, stringValue2);
        }
    }
}
``` 

**Observed**

Two distinct instances of `HexBinaryValue` or `StringValue`, for example, will not be deemed equal even if their values are equal.

[...]

PR #487, opened 2018-04-20
Make simple types comparable and equatable

ThomasBarnekow:
In conjunction with #485, this commit introduces two base classes,
`OpenXmlComparableReferenceType` and `OpenXmlComparableValueType`, which
implement the `IComparable`, `IComparable<T>`, and `IEquatable<T>`
interfaces. With the exception of `EnumValue` and `ListValue`, all children
or descendants of `OpenXmlSimpleType` are changed to become subclasses of
one of those two base classes.

This commit further adds unit tests for the new behavior. The core
tests are contained in the `OpenXmlComparableReferenceTypeTests` and
`OpenXmlComparableValueTypeTests` base classes. For each concrete simple
type, subclasses of those base classes create the concrete values to be
used in the generic unit tests.

Lastly, this commit fixes existing unit tests within the `SimpleTypeTest`
and `OpenXmlSimpleValueTest` classes, because those unit tests asserted
that two instances with same value but constructed separately are NOT
equal.

Note that the implementation is not yet complete as it lacks comments
for some public methods, for example. We might also have to add further
unit tests.


ThomasBarnekow:
@twsouthwick I've seen that the build failed because what are mere warnings in a local build is taken as an error on the build server. For example, the warning (or error on the build server) is that certain classes should implement Equals because they are implementing `IComparable`. However, they don't need to implement Equals, because all they would do is delegate that to the base class, which renders that implementation unnecessary.

I did not want to disable warnings etc., because I wanted to discuss this whole approach with you anyhow. But note that it builds just fine on my machine and all unit tests pass.

src/DocumentFormat.OpenXml/SimpleTypes/EnumValue.cs
```diff
@@ -10,15 +10,14 @@ namespace DocumentFormat.OpenXml
     /// Represents the enum value for attributes.
     /// </summary>
     /// <typeparam name="T">Every enum value must be an enum with the EnumStringValueAttribute object.</typeparam>
-    [DebuggerDisplay("{InnerText}")]
+    [DebuggerDisplay("{" + nameof(InnerText) + "}")]
     public class EnumValue<T> : OpenXmlSimpleValue<T>
         where T : struct
```
twsouthwick: For v3.0, we can add a constraint for Enum (as of C# 7.3) :)

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
```diff
@@ -0,0 +1,176 @@
...
+    /// </summary>
+    /// <typeparam name="T">The type of the value.</typeparam>
+    public abstract class OpenXmlComparableReferenceType<T> : OpenXmlSimpleType,
+        IComparable, IComparable<OpenXmlComparableReferenceType<T>>, IEquatable<OpenXmlComparableReferenceType<T>>
+        where T : class, IComparable, IComparable<T>, IEquatable<T>
+    {
+        /// <summary>
+        /// Creates a new instance of <see cref="OpenXmlComparableReferenceType{T}"/>.
+        /// </summary>
+        protected OpenXmlComparableReferenceType()
```
twsouthwick: Can you make these constructors `private protected`? I'd like to limit visibility of new types/methods as much as possible unless we have a reason to make them public

twsouthwick:
I've created an issue on their repo that you can reference when you disable the code with a pragma: https://github.com/dotnet/roslyn-analyzers/issues/1671

src/DocumentFormat.OpenXml/SimpleTypes/EnumValue.cs
ThomasBarnekow: @twsouthwick Do you want me to do anything here? ReSharper suggested this change and I've used `nameof` in my code as much as possible to avoid issues as names change (which might not be the case here).

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
ThomasBarnekow: Sure, can do.

src/DocumentFormat.OpenXml/SimpleTypes/EnumValue.cs
twsouthwick: Yeah, I saw that change as I quickly glanced through. I like the use of `nameof` as much as possible and see the value here. However, that name won't be changing, so I'd probably not make the change myself as it takes a bit more time to grok. I'll leave that up to you, though

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
ThomasBarnekow: What about the existing constructors of `OpenXmlSimpleType` and `OpenXmlSimpleValue<T>`, which are all `protected`? Should we consider making those `private protected` in v3.0?

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// </summary>
+        /// <param name="source">The source <see cref="OpenXmlComparableReferenceType{T}"/> instance.</param>
+        private protected OpenXmlComparableReferenceType(OpenXmlComparableReferenceType<T> source) : base(source)
+        {
+        }
+
+        /// <summary>
+        /// Gets or sets the value of this instance.
+        /// </summary>
+        public abstract T Value { get; set; }
```
twsouthwick: Does this need to be abstract? Or can it just be virtual (or even non-virtual)? It would nice to follow the pattern of `OpenXmlSimpleValue<T>`

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        /// <inheritdoc />
+        public int CompareTo(object obj)
+        {
+            if (obj is OpenXmlComparableReferenceType<T> other)
+            {
+                return CompareTo(other);
+            }
+
+            throw new ArgumentException();
```
twsouthwick: Should there be a message here?

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                return false;
+            }
+
+            if (Value == null)
+            {
+                return other.Value == null;
+            }
+
+            return Value.Equals(other.Value);
```
twsouthwick: Can this be collapsed into `Equals(Value, other?.Value)`?

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableValueType.cs
```diff
@@ -0,0 +1,162 @@
...
+
+using System;
+
+namespace DocumentFormat.OpenXml
+{
+    /// <summary>
+    /// Represents a comparable and equatable value for simple value types (Int32, UInt32, Byte, struct, etc).
+    /// </summary>
+    /// <typeparam name="T">The type of the value.</typeparam>
+    public abstract class OpenXmlComparableValueType<T> : OpenXmlSimpleValue<T>,
```
twsouthwick: What if `OpenXmlComparableValueType<T> : OpenXmlComparableReferenceType<T?>`? This could probably consolidate a bit of code.

twsouthwick: Actually, that's probably a breaking change (it would change the inheritance chain of some of the current values)

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
ThomasBarnekow: I thought about that as well. Currently, only three subclasses inherit from `OpenXmlComparableReferenceType<T>`:

- `Base64BinaryValue`,
- `HexBinaryValue`, and
- `StringValue`.

All of those basically delegate to `OpenXmlSimpleType.TextValue`. Therefore, I thought about two options:

1. turn `OpenXmlComparableReferenceType<T>` into something like `OpenXmlStringType` and provide a concrete implementation of the `Value` property, or
2. leave `OpenXmlComparableReferenceType<T>` generic to enable reuse for non-string types.

For the second option, which I chose for this implementation, I thought it was more elegant to just keep the that property abstract. However, we could also go for Option 1.

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
ThomasBarnekow: Easy answer: yes.

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
ThomasBarnekow: That depends on whether a non-null instance with a null Value property should be equal to a null instance.

For example, I could have two non-null `StringValue` instances, with `Value` being null for both of them. Those two should be considered equal. However, when I take any of those, I'd say that `stringValue.Equals(null)` should be false. In this case, the shortened `Equals` method could look like this:
```
        public bool Equals(OpenXmlComparableReferenceType<T> other)
        {
            return !(other is null) && Equals(Value, other.Value);
        }
```

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
twsouthwick: Ah - I hadn't realized it was a generic type. I like the current implementation better so it's more reusable.

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
twsouthwick: haha now the hard question is figuring out what to put :)

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableValueType.cs
twsouthwick: Yeah that's good for now. We can get the functionality in, and then mark it for clean up with an issue opened against the v3.0 milestone

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
ThomasBarnekow: Here's my current thinking (and I am very much open for suggestions):
```
        public int CompareTo(object obj)
        {
            switch (obj)
            {
                case null:
                    return 1;

                case OpenXmlComparableReferenceType<T> other:
                    return CompareTo(other);
            }

            throw new ArgumentException("Incompatible type: " + obj.GetType().FullName);
        }
```
I also used a switch statement with pattern matching.

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
```diff
@@ -0,0 +1,179 @@
...
+
+using System;
+
+namespace DocumentFormat.OpenXml
+{
+    /// <summary>
+    /// Represents a comparable and equatable reference.
+    /// </summary>
+    /// <typeparam name="T">The type of the value.</typeparam>
+    public abstract class OpenXmlComparableReferenceType<T> : OpenXmlSimpleType,
```
twsouthwick: The name feels a little weird. Can we follow the previous pattern and go with `OpenXmlComparableSimpleType`?

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableValueType.cs
twsouthwick: As with the previous comment for the reference type, I think `OpenXmlComparableSimpleValue` would fit better with the current naming

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
twsouthwick: I'm fine with that - just stick it in the resource file

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
twsouthwick: Hmm I still think that `Equals(Value, other?.Value)` would work, but I guess I'm not sure off the top of my head what `Equals(null,null)` is. Do you have unit tests covering the cases you bring up? If so you can just try it out. I'll leave this up to you - no need to rathole into it as the current implementation works

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
ThomasBarnekow: Yes, it is a little weird, so I am open for better names. But let me explain the thinking behind the current name.

On the `master` branch, we have `OpenXmlSimpleType` and `OpenXmlSimpleValue<T>` where the latter derives from the former. The subclasses representing value types (e.g., `DecimalValue`) and `EnumValue<T>` derive from `OpenXmlSimpleValue<T>`. The string-valued (reference) types (e.g., `StringValue`, `HexBinaryValue`) and `ListValue<T>` derive from `OpenXmlSimpleType`.

Now, to not introduce a breaking change, I needed to introduce:

1. one subclass of `OpenXmlSimpleValue<T>` (and therefore a grandchild of `OpenXmlSimpleType`) for all those comparable and equatable value types, hence the name `OpenXmlComparableValueType<T>`; and
2. one subclass of `OpenXmlSimpleType` that would become the superclass of the "reference types" `StringValue`, `HexBinaryValue`, and `Base64BinaryValue`.

For the second, the best name I could come up with was `OpenXmlComparableReferenceType`, because that was in contrast to `OpenXmlComparableValueType`, which was kind of given by its parent class `OpenXmlSimpleValue`.

Now, since both of those new, abstract base classes are "comparable simple types", we should not call just one of them `OpenXmlComparableSimpleType`.

Having said that, I am totally not wedded to the current names. If we find something that is logical, I am fine with that.

twsouthwick: Yeah, naming is hard 😆 I see your thought process and I think the issue I'm having with the name is "ValueType"/"ReferenceType" doesn't imply a kind of `OpenXmlSimpleType`. I want to make sure it's easy to see the relationship in the names (if possible). How about `OpenXmlComparableSimpleValue<T>` and `OpenXmlComparableSimpleReference<T>`. Not sure if they should be suffixed with "Type", but I think what I'm trying to get is a name that is obviously a subclass of `OpenXmlSimpleType`.

ThomasBarnekow: `OpenXmlComparableSimpleValue<T>` and `OpenXmlComparableSimpleReference<T>` would be fine for me. I think I thought about the SimpleReference as well, but wasn't sure whether that introduces an unwanted connotation. But if you are fine with those, we can certainly also use those two.

And if you like `OpenXmlComparableSimpleType<T>` better, we could also go with that. As noted in my other comment, my only thought would be whether `OpenXmlComparableSimpleValue<T>` should derive from `OpenXmlComparableSimpleType<T>`, if possible.

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
ThomasBarnekow: `Equals(null, null)` should be true.

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableReferenceType.cs
twsouthwick: I like the first two. I agree they're not ideal, but it's a good compromise to keep naming consistent.

If we can get `OpenXmlComparableSimpleValue<T>` to derive from `OpenXmlComparableSimpleType<T>` that would be awesome, but I think it would break the hierarchy. Note the PR has a `package-compat-analyzer` check I wrote to validate that we don't break binary compatibility, so if you're not sure, you can submit an update and see what that analyzer says.

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableValueType.cs
twsouthwick: If we were to derive such as `OpenXmlComparableSimpleValue<T> : OpenXmlComparableSimpleType<T?>` would that work?

ThomasBarnekow:
@twsouthwick I've pushed a commit with the changes we've discussed.

BTW, I am getting messages that Travis CI builds are failing.

twsouthwick:
What Travis CI failures are you referring to? I only know of AppVeyor

test/DocumentFormat.OpenXml.Tests/OpenXmlDomTest/OpenXmlSimpleValueTest.cs
```diff
@@ -84,7 +85,13 @@ public void BooleanValueTest()
             Log.Comment("Verifying reference type behavior...");
             objA = new BooleanValue(validValue);
             objB = new BooleanValue(validValue);
-            Log.VerifyFalse(object.Equals(objA, objB), "Two instances with same value but constructed separated are Equal.");
+
+            // This is the previous assertion, which fails due to the proposed change:
```
twsouthwick: Can you remove these comments and update it to use `Assert.Equal` instead - the `Log` file system is something I'd like to move away from as XUnit provides most of the testing stuff (and helpful diagnostics) we need

ThomasBarnekow:
The Travis CI failures came from an outdated build that I had once configured for my repo when we originally set up build servers for the Open XML SDK.

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableSimpleReference.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="left"></param>
+        /// <param name="right"></param>
+        /// <returns></returns>
+        public static bool operator <(OpenXmlComparableSimpleReference<T> left, OpenXmlComparableSimpleReference<T> right)
+        {
```
ThomasBarnekow: What about the comments on those operators (==, !=, <, <=, >, >=)? I have not completed them but don't know whether we really need them.

test/DocumentFormat.OpenXml.Tests/OpenXmlDomTest/OpenXmlSimpleValueTest.cs
twsouthwick: Sounds good. I'm not saying remove all the Log.Verify[...] calls, just switch to XUnit for any new code :) I've been chipping away at the removal of the old Log infrastructure for a while now... I definitely don't expect that in this PR :)

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableSimpleReference.cs
twsouthwick: hmmm for completeness, either fill them in (probably some boiler plate comments), or just remove them (unless if stylecop complains - I can't remember if I've turned those on yet)

twsouthwick:
Ah, yes, if you set up TravisCI for an old build system, it will most definitely break. The new project uses new csproj and requires VS 2017 Update 3 (or 5... can't remember). Right now we're using AppVeyor, but it would be nice to expand that out to test on Mono and multiple locales (and potentially Xamarin and .NET Native), but that would be for a different (or set of) PRs

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableSimpleReference.cs
ThomasBarnekow: I added those empty comments because Style Cop did complain. I will steal something elsewhere.

twsouthwick:
The test that failed is one I just added... I'll merge this and then figure out what's causing those failures

PR #494, opened 2018-05-02
Add OpenXmlElement.StrictTranslateAttribute to handle strict translation

### LINKED ISSUE #480: static method AttributeTranslator.DicOfTranslators is not thread safe
default
When loading excel files from different threads this method can be executed at same time. On screenshot we got race condition when inserting tblLook in non thread safe class Dictionary

twsouthwick:
This change removes the race conditions that could be encountered when parsing attributes when strict translation is on. Instead of a dictionary of types that it applies to, this adds an overload to OpenXmlElement to handle any strict translation needed by the attributes. This way, the type itself can opt into any strict translation that it requires.

Fixes #480


twsouthwick:
@tomjebo The original implementation had this line:

https://github.com/OfficeDev/Open-XML-SDK/blob/master/src/DocumentFormat.OpenXml/Attributes/Translator/AttributeTranslator.cs#L30

Which implies that it should be applied for the local name `jc`. However, due to the fact that no entry for `jc` was added to the dictionary, it would actually never be translated.

I'm not familiar enough with the spec to know where to find the answer for this, but enabled it for elements with `jc` as the local name. Can you verify that this is the correct behavior?

PR #534, opened 2019-01-10
Run tests on .NET Core 2.1

twsouthwick:


twsouthwick:
This is broken due to https://github.com/NuGet/Home/issues/7414

twsouthwick:
Done via #573

PR #586, opened 2019-04-01
Add Office2016 support

twsouthwick:


twsouthwick:
@tomjebo 

FYI the number changes in O*UnionHelper.g.cs, *.constraints, and *.restraints.xml are due to order differences by inserting properties in. They are calculated during the schema generation and can be considered "magic" numbers. Long term, we can hopefully remove them (I've some progress in removing dependency on them), but for now, it's still needed.

PR #623, opened 2019-08-30
Update Azure DevOps builds to use yaml config

twsouthwick:


twsouthwick:
@tomjebo Can you take a look at this?

PR #624, opened 2019-09-03
Speed up automated builds

twsouthwick:
This change does a couple of things:

1. Move GitVersion calculation to a build step so it's not done every build
2. Because of (1), we can now use multicore build with msbuild (the `/m` switch)
3. Set up stages for Azure DevOps so each test framework is now run on their own machine
4. Add benchmark test run (using InProcess version)


twsouthwick:
@tomjebo Can you review this?

PR #653, opened 2019-12-19
Explain ProjectLoadStyle

### LINKED ISSUE #652: Open-XML-SDK.sln fails to build due to incorrect ProjectLoadStyle environment variable
**Description**

The Open-XML-SDK.sln fails to build in both Visual Studio 2019 Version 16.4.2 and on the command line.

In Visual Studio, using the DocumentFormat.OpenXml.csproj as an example, the error message for all projects is:
```
Error    NETSDK1013    The TargetFramework value '' was not recognized. It may be misspelled. If not, then the TargetFrameworkIdentifier and/or TargetFrameworkVersion properties must be specified explicitly.    DocumentFormat.OpenXml    C:\Program Files\dotnet\sdk\3.1.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets    93
```

On the command line, the error message shown by `dotnet restore` is:

```
C:\Program Files\dotnet\sdk\3.1.100\NuGet.targets(123,5): error : Value cannot be null. (Parameter 'folderName') [[local path removed]\Open-XML-SDK\Open-XML-SDK.sln]
```

**Information**

- Visual Studio: Community 2019 Version 16.4.2 (with previous versions, the projects can't even be opened)
- .NET Target: Any (did not explicitly specify)
- DocumentFormat.OpenXml Version: commit 871600ce22399c69bbbbb4c7f2eed93a59aa580b on master
- `ProjectLoadStyle` environment variable set to `DevCore10` or `DevCore31`

**Repro**

1. Create a fresh clone of the Open XML SDK, using `git clone https://github.com/OfficeDev/Open-XML-SDK.git`
2. Build using Visual Studio or on the command line:
    - In Visual Studio: open and build the solution.
    - On the command line: go to the top-level folder with `.sln` file and execute `dotnet restore`

**Observed**

As described above, Visual Studio shows build errors and `dotnet restore` prints an error message.

**Expected**

It should be possible to build the latest commit on master. The build instructions should include important details (e.g., environment variables). Where environment variables, for example, have become obsolete, a default configuration should be built.

twsouthwick:
Fixes #652

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -15,7 +15,7 @@
          - DevCore31:      .NET Core 3.1
          - All: Will build for all platforms
          -->
-    <ProjectLoadStyle Condition=" '$(ProjectLoadStyle)' == '' ">DevCore21</ProjectLoadStyle>
+    <ProjectLoadStyle Condition=" '$(ProjectLoadStyle)' == '' ">DevCore31</ProjectLoadStyle>
   </PropertyGroup>
 
   <PropertyGroup>
@@ -64,7 +64,7 @@
         <SamplesFrameworks>netcoreapp2.1</SamplesFrameworks>
       </PropertyGroup>
     </When>
-    <When Condition=" '$(ProjectLoadStyle)' == 'DevCore30' ">
+    <When Condition=" '$(ProjectLoadStyle)' == 'DevCore31' ">
       <PropertyGroup>
         <ProductTargetFrameworks>netstandard2.0</ProductTargetFrameworks>
         <TestTargetFrameworks>netcoreapp3.1</TestTargetFrameworks>
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ This project uses the csproj format and the release versions of the tooling in V
 
 The project often requires the latest release of the C# compiler as many new features come on-line that greatly aid in ease of development. As of now, the C# 7.2 compiler is required (which was released in December 2017) and comes standard in Visual Studio 2017 Update 5, with other IDEs providing updates to the compiler, as well.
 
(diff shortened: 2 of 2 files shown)


twsouthwick:
@ThomasBarnekow I think the issue you're hitting is not having the .NET Core 2.1 SDK installed. If you set the `ProjectLoadStyle` to `DevFramework46` you should be able to build. This PR adds directions to the README about this - let me know if this doesn't help and we can dig in further.

twsouthwick:
This also changes the default dev environment to .NET Core 3.1 (I want to keep it on the current LTS)

ThomasBarnekow:
@twsouthwick, I just saw this after having opened a PR (see #658) that fixes the issue (see #652). Yes, an explanation in the README would be helpful. And it seems you have already fixed one of the issues that I also fixed in my PR (and you changed the default). Thus, I need to amend my PR, which adds one more thing on top of yours, i.e., it adds an `Otherwise` block that defines all variables based on the default. I used the `DevCore21` settings and should also change that to use those defined for `DevCore31` now that you defined that as the default.

ThomasBarnekow:
> I think the issue you're hitting is not having the .NET Core 2.1 SDK installed. If you set the ProjectLoadStyle to DevFramework46 you should be able to build. This PR adds directions to the README about this - let me know if this doesn't help and we can dig in further.

That was not the issue. I had everything installed, including the latest .NET Core 3.1 SDK. The problem was a now-deprecated `ProjectLoadStyle` environment variable, which was set to `DevCore10`. Through the `Otherwise` block mentioned above, problems of that kind will be avoided in the future.

twsouthwick:
Ah yes! I removed the DevCore10 since it's no longer supported. Good catch. I'll merge this in and then yours so we have the README changes and your changes as well. Thanks!

PR #659, opened 2019-12-19
Consider altChunk parts in Flat OPC transformation

### LINKED ISSUE #525: ToFlatOpcDocument and ToFlatOpcString deliver a document that is not useable in MS Word when there is an altChunk-Part
**Description**
ToFlatOpcDocument and ToFlatOpcString deliver a document that is not useable in MS Word when there is an altChunk-Part.

When a document contains an altChunk-Part with content-Type "application/xhtml+xm" the conversion routine returns a part with a pkg:xmlData-Element.
Such documents cannot be processed in MS Word (tested in 2010 and 2016).

MS Word expects altChunk-Parts always to be in a binary format in a Flat-OPC-document. Otherwise there is an error (Word says the document is corrupted and asks if it should repair the document).
The part information is lost when trying to repair the document in MS Word.

This is controlled in OpenXmlPackage.GetContentsAsXml by checking: part.ContentType.EndsWith("xml")
The standard lists the following content-types for altChunk:
- text/plain
- text/html
- application/vnd.openxmlformats-officedocument.wordprocessingml.document
- application/xhtml+xml

The 3rd entry seems like an error because +xml is missing in the standard. I did not try altChunk with Word-Processing-ML but I suppose Word expects a binary part too.
Considering that, a check for ending of "xml" or "+xml" is not really feasible for the problem.

To be able to tell, if this is an altChunk-Part that should be binary (for MS Word) the Relationship-Type ("http://schemas.openxmlformats.org/officeDocument/2006/relationships/aFChunk" or the purl-Version "http://purl.oclc.org/ooxml/officeDocument/relationships/aFChunk") of the part should be checked and not the content-Type of the part.
The check for "xml" could be done after this check for altChunk but should anyway rather be for "+xml" instead.

**Information**

- .NET Target: 4.6.1
- DocumentFormat.OpenXml Version: 2.7.2 but code in 2.9 seams to have the same issue.

**Repro**
See Unit-Test at:
https://gist.github.com/rominator1983/248dcd826802eeb0a768f18fe709f2b1

An additional file (docx with altchunk) is needed:
https://amforst.ddns.net/s/qd5RELXM6TeMX8t

**Observed**
altChunk-Part is returned as XML-Part instead of binary.

**Expected**
altChunk-Part is returned as binary-Part instead of XML-Part.

---
Copied from https://gist.github.com/rominator1983/248dcd826802eeb0a768f18fe709f2b1 for easier discoverability:

```csharp
using System.IO;
using System.IO.Packaging;
using DocumentFormat.OpenXml;
using NUnit.Framework;

namespace ConversionTestNamespace
{
  [TestFixture]
  public class ConvertTest
  {
    [Test]
    public void Convert ()
    {
      // download file from: https://amforst.ddns.net/s/qd5RELXM6TeMX8t
      using (var package = Package.Open (@"c:\\temp\\Doc with altChunk.docx", FileMode.Open))
      {
        using (var wordprocessingDocument = WordprocessingDocument.Create (package, WordprocessingDocumentType.Document))
        {
          var flatOpcDocumentString = wordprocessingDocument.ToFlatOpcString ();

          Assert.That (flatOpcDocumentString, Does.Contain("pkg:binaryData"));
        }
      }
    }
  }
}
```

ThomasBarnekow:
This commit enhances the

  ```XDocument OpenXmlPackage.ToFlatOpcDocument(XProcessingInstruction)```

method to consider "altChunk parts" (AlternativeFormatInputParts),
which must have their content rendered in binary form even if it is XML
content (e.g., XML, XHTML).

Fixes #525.

### DIFF
--- a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.IO;
 using System.IO.Packaging;
 using System.Linq;
@@ -14,10 +16,13 @@ namespace DocumentFormat.OpenXml.Packaging
     /// <summary>
     /// Represents a base class for strong typed Open XML document classes.
     /// </summary>
-    public abstract partial class OpenXmlPackage : OpenXmlPartContainer, IDisposable
+    public abstract partial class OpenXmlPackage
     {
-        private static readonly XNamespace pkg = "http://schemas.microsoft.com/office/2006/xmlPackage";
-        private static readonly XNamespace rel = "http://schemas.openxmlformats.org/package/2006/relationships";
+        private static readonly XNamespace Pkg = "http://schemas.microsoft.com/office/2006/xmlPackage";
+        private static readonly XNamespace Rel = "http://schemas.openxmlformats.org/package/2006/relationships";
+
+        private const string AltChunkRelationshipType =
+            "http://schemas.openxmlformats.org/officeDocument/2006/relationships/aFChunk";
 
         /// <summary>
         /// Converts an OpenXml package in OPC format to string in Flat OPC format.
@@ -49,6 +54,15 @@ namespace DocumentFormat.OpenXml.Packaging
             // to call it here.
             Save();
 
+            // Identify all AlternativeFormatInputParts (AltChunk parts).
+            // This is necessary because AltChunk parts must be treated as binary
+            // parts regardless of the actual content type, which might even be
+            // XML-related such as application/xhtml+xml.
+            var altChunkPartUris = new HashSet<Uri>(
+                new InternalPackageRelationships(Package)
+                    .Where(pr => pr.RelationshipType == AltChunkRelationshipType)
+                    .Select(pr => PackUriHelper.ResolvePartUri(pr.SourceUri, pr.TargetUri)));
+
             // Create an XML document with a standalone declaration, processing
             // instruction (if not null), and a package root element with a
             // namespace declaration and one child element for each part.
@@ -56,61 +70,69 @@ namespace DocumentFormat.OpenXml.Packaging
                 new XDeclaration("1.0", "UTF-8", "yes"),
                 instruction,
                 new XElement(
-                    pkg + "package",
-                    new XAttribute(XNamespace.Xmlns + "pkg", pkg.ToString()),
-                    Package.GetParts().Select(part => GetContentsAsXml(part))));
+                    Pkg + "package",
+                    new XAttribute(XNamespace.Xmlns + "pkg", Pkg.ToString()),
+                    Package.GetParts().Select(part => GetContentsAsXml(part, altChunkPartUris))));
         }
 
         /// <summary>
-        /// Gets the <see cref="PackagePart"/>'s contents as an <see cref="XElement"/>.
+        /// Gets the <see cref="PackagePart"/>'s XML or binary contents as an <see cref="XElement"/>.
         /// </summary>
         /// <param name="part">The package part.</param>
+        /// <param name="altChunkPartUris">The collection of AlternativeFormatInputPart URIs.</param>
         /// <returns>The corresponding <see cref="XElement"/>.</returns>
-        private static XElement GetContentsAsXml(PackagePart part)
+        private static XElement GetContentsAsXml(PackagePart part, IEnumerable<Uri> altChunkPartUris)
         {
-            if (part.ContentType.EndsWith("xml", StringComparison.Ordinal))
+            if (part.ContentType.EndsWith("xml") && !altChunkPartUris.Contains(part.Uri))
             {
                 using (Stream stream = part.GetStream())
-                using (StreamReader streamReader = new StreamReader(stream))
+                using (var streamReader = new StreamReader(stream))
                 using (XmlReader xmlReader = XmlReader.Create(streamReader))
                 {
                     return new XElement(
-                        pkg + "part",
-                        new XAttribute(pkg + "name", part.Uri),
-                        new XAttribute(pkg + "contentType", part.ContentType),
-                        new XElement(pkg + "xmlData", XElement.Load(xmlReader)));
+                        Pkg + "part",
+                        new XAttribute(Pkg + "name", part.Uri),
+                        new XAttribute(Pkg + "contentType", part.ContentType),
+                        new XElement(Pkg + "xmlData", XElement.Load(xmlReader)));
                 }
             }
-            else
+
+            return GetBinaryPartContentsAsXml(part);
+        }
+
+        /// <summary>
+        /// Gets the <see cref="PackagePart"/>'s binary contents as an <see cref="XElement"/>.
+        /// </summary>
+        /// <param name="part">The package part.</param>
+        /// <returns>The corresponding <see cref="XElement"/>.</returns>
+        private static XElement GetBinaryPartContentsAsXml(PackagePart part)
+        {
+            using (Stream stream = part.GetStream())
             {
-                using (Stream stream = part.GetStream())
-                using (BinaryReader binaryReader = new BinaryReader(stream))
-                {
-                    int len = (int)binaryReader.BaseStream.Length;
-                    byte[] byteArray = binaryReader.ReadBytes(len);
+                var byteArray = new byte[stream.Length];
+                stream.Read(byteArray, 0, byteArray.Length);
 
-                    // The following expression creates the base64String, then chunks
-                    // it to lines of 76 characters long.
-                    string base64String = System.Convert.ToBase64String(byteArray)
-                        .Select((c, i) => new { Character = c, Chunk = i / 76 })
-                        .GroupBy(c => c.Chunk)
-                        .Aggregate(
-                            new StringBuilder(),
-                            (s, i) =>
-                                s.Append(
-                                    i.Aggregate(
-                                        new StringBuilder(),
-                                        (seed, it) => seed.Append(it.Character),
-                                        sb => sb.ToString())).Append(Environment.NewLine),
-                            s => s.ToString());
+                // The following expression creates the base64String, then chunks
+                // it to lines of 76 characters long.
+                string base64String = Convert.ToBase64String(byteArray)
+                    .Select((@char, index) => new { Character = @char, Chunk = index / 76 })
+                    .GroupBy(charAndChunk => charAndChunk.Chunk)
+                    .Aggregate(
+                        new StringBuilder(),
+                        (sb, grouping) => sb
+                            .Append(grouping.Aggregate(
+                                new StringBuilder(),
+                                (chunkSb, charAndChunk) => chunkSb.Append(charAndChunk.Character),
+                                chunkSb => chunkSb.ToString()))
+                            .Append(Environment.NewLine),
+                        sb => sb.ToString());
 
-                    return new XElement(
-                        pkg + "part",
-                        new XAttribute(pkg + "name", part.Uri),
-                        new XAttribute(pkg + "contentType", part.ContentType),
-                        new XAttribute(pkg + "compression", "store"),
-                        new XElement(pkg + "binaryData", base64String));
-                }
+                return new XElement(
+                    Pkg + "part",
+                    new XAttribute(Pkg + "name", part.Uri),
+                    new XAttribute(Pkg + "contentType", part.ContentType),
+                    new XAttribute(Pkg + "compression", "store"),
+                    new XElement(Pkg + "binaryData", base64String));
             }
         }
 
@@ -167,23 +189,28 @@ namespace DocumentFormat.OpenXml.Packaging
                 throw new ArgumentNullException(nameof(package));
             }
 
+            if (document.Root is null)
+            {
+                throw new ArgumentException(ExceptionMessages.RootElementIsNull, nameof(document));
+            }
+
             // Add all parts (but not relationships).
-            foreach (var xmlPart in document.Root
+            foreach (XElement xmlPart in document.Root
                 .Elements()
                 .Where(p =>
-                    (string)p.Attribute(pkg + "contentType") !=
-                        "application/vnd.openxmlformats-package.relationships+xml"))
+                    (string) p.Attribute(Pkg + "contentType") !=
+                    "application/vnd.openxmlformats-package.relationships+xml"))
             {
-                string name = (string)xmlPart.Attribute(pkg + "name");
-                string contentType = (string)xmlPart.Attribute(pkg + "contentType");
-                if (contentType.EndsWith("xml", StringComparison.Ordinal))
+                var name = (string) xmlPart.Attribute(Pkg + "name");
+                var contentType = (string) xmlPart.Attribute(Pkg + "contentType");
+                if (contentType.EndsWith("xml"))
                 {
-                    Uri uri = new Uri(name, UriKind.Relative);
+                    var uri = new Uri(name, UriKind.Relative);
                     PackagePart part = package.CreatePart(uri, contentType, CompressionOption.SuperFast);
                     using (Stream stream = part.GetStream(FileMode.Create))
                     using (XmlWriter xmlWriter = XmlWriter.Create(stream))
                     {
-                        xmlPart.Element(pkg + "xmlData")
+                        xmlPart.Elements(Pkg + "xmlData")
                             .Elements()
                             .First()
                             .WriteTo(xmlWriter);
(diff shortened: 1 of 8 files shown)


ThomasBarnekow:
@twsouthwick, how can I check why this failed? I don't have access to the Azure DevOps project. 

All unit tests passed on my local machine, so I don't know what might have happened.

ThomasBarnekow:
I might have found another "bug" in `Directory.Build.props`, which still lists `net35`.

```
<ProductTargetFrameworks>netstandard1.3;netstandard2.0;net35;net40;net46</ProductTargetFrameworks>
```

After setting `ProjectLoadStyle` to `All`, the build failed also on my machine. This was due to the fact that `net35` does not provide the `ISet<T>` interface.

ThomasBarnekow:
So I added a fix for net35 and the solution was built successfully and all unit tests passed. What goes wrong? I don't have access to the details.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -49,68 +54,85 @@ protected XDocument ToFlatOpcDocument(XProcessingInstruction instruction)
...
             return new XDocument(
                 new XDeclaration("1.0", "UTF-8", "yes"),
                 instruction,
                 new XElement(
-                    pkg + "package",
-                    new XAttribute(XNamespace.Xmlns + "pkg", pkg.ToString()),
-                    Package.GetParts().Select(part => GetContentsAsXml(part))));
+                    Pkg + "package",
+                    new XAttribute(XNamespace.Xmlns + "pkg", Pkg.ToString()),
+                    Package.GetParts().Select(p => GetContentsAsXml(p, altChunkPartUris))));
```
rominator1983: I personally do not like abbreviations in code but I don't know about the style guide of this project. So that's up to you...

ThomasBarnekow: We might even be able to fit in `art` twice ;-)

ThomasBarnekow:
@rominator1983, do you have access to those details which might tell us why the check failed?

twsouthwick:
Good point about access. We're building it on the office dev ops instance so we can sign stuff... but we can't give access. I'll look into how we can expose that.

twsouthwick:
As for .NET 3.5, I do want to continue to support it, but it does limit some things. What I end up doing is adding stubs for the types to get things working. You can see under the `System` folder, a number of these, such as `IReadOnlyLIst` and others.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -49,68 +54,85 @@ protected XDocument ToFlatOpcDocument(XProcessingInstruction instruction)
...
-                    return new XElement(
-                        pkg + "part",
-                        new XAttribute(pkg + "name", part.Uri),
-                        new XAttribute(pkg + "contentType", part.ContentType),
-                        new XAttribute(pkg + "compression", "store"),
-                        new XElement(pkg + "binaryData", base64String));
-                }
+                // The following expression creates the base64String, then chunks
+                // it to lines of 76 characters long.
+                string base64String = Convert.ToBase64String(byteArray)
```
twsouthwick: Can you move this to a separate method?

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -270,5 +294,79 @@ protected static Package FromFlatOpcDocumentCore(XDocument document, Package pac
             package.Flush();
             return package;
         }
+
+        /// <summary>
+        /// Represents the collection of internal package relationships.
+        /// </summary>
+        private class InternalPackageRelationships : IEnumerable<PackageRelationship>
```
twsouthwick: Could this be done as an iterator function instead?

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -270,5 +294,79 @@ protected static Package FromFlatOpcDocumentCore(XDocument document, Package pac
...
+                if (relationships.Contains(r))
+                {
+                    return;
+                }
+
+                relationships.Add(r);
+#else
+            private void Enumerate(PackageRelationship r, ISet<PackageRelationship> relationships)
+            {
+                if (!relationships.Add(r))
```
twsouthwick: The logic here is different. How come you need `ISet<>`? Why not just ask for the HashSet<T> that is generated above? These are all internal, so no worry about leaking details

twsouthwick:
Thanks for taking a look at this. A few comments about some of the structure. I tend to avoid too much LINQ as recent profiling I've done has identified it as hot spots. I'm fine with starting off with it for getting correctness, but keeping them in separate methods will help during profiling to identify any specific queries that are problematic.

Also, I don't see any tests being added? Can you add some of those? Preferably more unit test style without needing a full file; just build up the components needed and the expected outcome.

ThomasBarnekow:
@twsouthwick, thanks for your feedback. 

Please have a look at the unit test in the `OpenXmlPackageTests` class in the `DocumentFormat.OpenXml.Packaging.Tests` project.

twsouthwick:
Must have missed the test - sorry about that. You can disregard that comment.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -270,5 +294,79 @@ protected static Package FromFlatOpcDocumentCore(XDocument document, Package pac
...
+            {
+                public bool Equals(PackageRelationship x, PackageRelationship y)
+                {
+                    return x?.SourceUri == y?.SourceUri && x?.TargetUri == y?.TargetUri;
+                }
+
+                public int GetHashCode(PackageRelationship obj)
+                {
+                    string uriString = obj.SourceUri + ":" + obj.TargetUri;
+                    return uriString.GetHashCode();
```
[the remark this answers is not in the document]
twsouthwick: I'm updating an internal method to support this. Once #660 is merged, you can build a hashcode with comparers.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
ThomasBarnekow: Done.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
ThomasBarnekow: I typically try to use the most abstract types to foster reuse. In this case, that would be `ISet<T>`. But that simply does not exist in `net35` (released 12 years ago).

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -14,10 +16,13 @@ namespace DocumentFormat.OpenXml.Packaging
     /// <summary>
     /// Represents a base class for strong typed Open XML document classes.
     /// </summary>
-    public abstract partial class OpenXmlPackage : OpenXmlPartContainer, IDisposable
+    public abstract partial class OpenXmlPackage
```
twsouthwick: I just realized you're removing class hierarchy - that's a breaking change. Why are you doing this?

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
twsouthwick: Yes, I agree with trying to do that, but do you *need* it to be `ISet<T>` here? What's the limitation of `ICollection<T>`?

That's the challenge with trying to support `net35` :( We'll probably drop it at some point in the near future, once we can figure out usage and the impact that will have.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
ThomasBarnekow: Yes, I will remove this general-purpose class and replace it with a much simpler, altChunk-specific expression.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
ThomasBarnekow: No, it does not have to be `ISet<T>`. However, if it is not some kind of set, the `Add(T)` method will not return a flag telling me whether the element was added. If it is an `ICollection<T>`, I will have to check for containment first. However, if the underlying collection is a `HashSet<T>`, that will also be O(1). But I need two method calls rather than just one.

As I'm now getting rid of this class, there is no issue anymore. I borrowed it from another use case where net35 is no factor.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
ThomasBarnekow: I don't remove class hierarchy. In a **partial class**, we don't have to mention the base class and interfaces again. I could add those back in if you like.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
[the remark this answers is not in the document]
ThomasBarnekow: I'll get rid of the class, so that's no longer needed.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -49,68 +54,85 @@ protected XDocument ToFlatOpcDocument(XProcessingInstruction instruction)
...
 
         /// <summary>
-        /// Gets the <see cref="PackagePart"/>'s contents as an <see cref="XElement"/>.
+        /// Gets the <see cref="PackagePart"/>'s XML or binary contents as an <see cref="XElement"/>.
         /// </summary>
         /// <param name="part">The package part.</param>
+        /// <param name="altChunkPartUris">The collection of AlternativeFormatInputPart URIs.</param>
         /// <returns>The corresponding <see cref="XElement"/>.</returns>
-        private static XElement GetContentsAsXml(PackagePart part)
+        private static XElement GetContentsAsXml(PackagePart part, IEnumerable<Uri> altChunkPartUris)
```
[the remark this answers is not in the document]
ThomasBarnekow: The formal parameter does not have to be a `HashSet<T>` to utilize the O(1) for direct lookup. While it turns out that an `IEnumerable<T>` is not the optimal choice, because `Contains(T)` is an extension method rather than a method defined by the interface, the `Contains(T)` implementation of `HashSet<T>` will be used when it is passed as an actual parameter.

Having said that, I also think it would be better to use `HashSet<T>` or at a minimum `ICollection<T>`, which defines the `Contains(T)` method for `HashSet<T>`.

ThomasBarnekow:
Made the changes requested by @twsouthwick. Rebased on master and fixed up some commits.

ThomasBarnekow:
@twsouthwick, can you tell me what went wrong here? Setting `ProjectLoadStyle` to `All`, the solution is successfully built and all unit tests pass on my machine.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
twsouthwick: Ah - did know it was defined on the other partial class. Yeah, I'm fine with not having it in this file (no need for duplication)

twsouthwick:
I have warnings as errors turned on. I recently added some FxCop rules to enforce StringComparer, so that's what's being raised:

```
src\DocumentFormat.OpenXml\Packaging\OpenXmlPackage.FlatOpc.cs(89,17): Error CA1307: The behavior of 'string.EndsWith(string)' could vary based on the current user's locale settings. Replace this call in 'DocumentFormat.OpenXml.Packaging.OpenXmlPackage.GetContentsAsXml(System.IO.Packaging.PackagePart, System.Collections.Generic.HashSet<System.Uri>)' with a call to 'string.EndsWith(string, System.StringComparison)'.
```

Not having access to the build is not good. I may create a separate devops instance for CI builds that I can make public. I'll look into that.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -49,71 +56,89 @@ protected XDocument ToFlatOpcDocument(XProcessingInstruction instruction)
...
+                var byteArray = new byte[stream.Length];
+                stream.Read(byteArray, 0, byteArray.Length);
+                return ToChunkedBase64String(byteArray);
             }
         }
 
+        private static string ToChunkedBase64String(byte[] byteArray)
+        {
+            return Convert.ToBase64String(byteArray)
+                .Select((@char, index) => new { Character = @char, Chunk = index / 76 })
```
twsouthwick: Can you make the `76` a const with a descriptive name as to what it is (ie `CharactersPerRow` or something?). I think it's something we'll forget what it is at some point :)

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -49,71 +56,89 @@ protected XDocument ToFlatOpcDocument(XProcessingInstruction instruction)
...
+        {
+            using (Stream stream = part.GetStream())
+            {
+                var byteArray = new byte[stream.Length];
+                stream.Read(byteArray, 0, byteArray.Length);
+                return ToChunkedBase64String(byteArray);
             }
         }
 
+        private static string ToChunkedBase64String(byte[] byteArray)
```
twsouthwick: I swear I've seen something like this in the code base before. Do you know if it's done else where? It may be useful if so to make it a general purpose utility

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -167,61 +192,63 @@ protected static Package FromFlatOpcDocumentCore(XDocument document, Package pac
...
+                    using (var binaryWriter = new BinaryWriter(stream))
                     {
-                        string base64StringInChunks = (string)xmlPart.Element(pkg + "binaryData");
-                        char[] base64CharArray = base64StringInChunks
-                            .Where(c => c != '\r' && c != '\n').ToArray();
-                        byte[] byteArray =
-                            System.Convert.FromBase64CharArray(
-                                base64CharArray, 0, base64CharArray.Length);
+                        var base64StringInChunks = (string) xmlPart.Element(Pkg + "binaryData");
+                        char[] base64CharArray = base64StringInChunks.Where(c => c != '\r' && c != '\n').ToArray();
```
twsouthwick: Can you move this out to a function as well?

ThomasBarnekow:
@twsouthwick, is it possible that you turned those FxCop rules on after my previous build? That would explain it because I did not change that line of code with the warning.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
ThomasBarnekow: Done. The 76 characters are based on the MIME specification, which defines a maximum line length of 76 characters for Base64-encoded lines.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
ThomasBarnekow: I looked around but could not find anything else. You might have seen it in the Flat OPC code before. I turned it into a function, which I then beautified (e.g., formatted, renamed parameters).

twsouthwick:
It was sometime yesterday (Pacific time) in the afternoon. Highly possible

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
twsouthwick: sounds good.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
twsouthwick: Good to know - can you ensure there's a comment to that effect?

twsouthwick:
I opened #666 as a tracking item for seeing build failures

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
ThomasBarnekow: Done. And you got me started on a more complete refactoring of the other methods as well.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
ThomasBarnekow: That's already included, of course ;-)

ThomasBarnekow:
@twsouthwick, unless you have further input, this would be it for now. I've updated CHANGELOG.md and also beautified the Flat OPC-related partial classes (where I also removed the redundant base type references.

PR #661, opened 2019-12-20
Add a RootPart property to OpenXmlPackage

twsouthwick:
Some code will manually do a switch against the document type and requires a knowledge of what types are which; this exposes a simple way to access this information from an `OpenXmlPackage`.


twsouthwick:
@tomjebo @ThomasBarnekow  Can you take a look at this? I'm not certain about the name, or if it would be useful to have `public`. Thoughts?

ThomasBarnekow:
@twsouthwick and @tomjebo, I like this idea.

`RootPart` works well for me because it corresponds to the `RootElement` property of the `OpenXmlPart` in terms of naming, function, and visibility. Therefore, I would also make it public.

PR #679, opened 2020-02-07
Implement OpenXmlElementFunctionalExtensions

### LINKED ISSUE #634: Increase flexibility of constructors, e.g., for pure functional transformations
**Description**

In the context of pure functional transformations, it would be beneficial to construct `OpenXmlElement` instances very flexibly. For example, one might want to construct a new element in the following way:
```csharp
return new Paragraph(
    new ParagraphProperties(new ParagraphStyleId { Val = "MyStyleId" }),
    FunctionReturningIEnumerableOfRun(someParams));
```

**Information**

- .NET Target: All
- DocumentFormat.OpenXml Version: Latest

See discussion in #627.

**Observed**

Using the `Paragraph` class as an example, the subclasses of `OpenXmlElement` define the following constructors:
```csharp
public Paragraph()
{
}

public Paragraph(IEnumerable<OpenXmlElement> childElements)
    : base(childElements)
{
}

public Paragraph(params OpenXmlElement[] childElements)
    : base(childElements)
{
}

public Paragraph(string outerXml)
    : base(outerXml)
{
}
```
With those constructors, the example instance can't be created, because we can't combine an `OpenXmlElement`  and an `IEnumerable<OpenXmlElement>`.

**Expected**

We should be able to create instances in the way shown in the description.

**Design Idea**

We could borrow the design from the `System.Xml.Linq.XElement` class, which, owing to how you can construct instances, is instrumental in pure functional transformations.

The two key `XElement` constructors for pure functional transformations are:
```csharp
public XElement(XName name, object content)
    : this(name)
{
    this.AddContentSkipNotify(content);
}

public XElement(XName name, params object[] content)
    : this(name, (object) content)
{
}
```
Note how the second constructor delegates to the first one. Thus, the magic happens in the `AddContentSkipNotify(object)` method, which works with pretty much any meaningful actual parameter you throw at it.

Thus, using the `Paragraph` example, we could replace:
- `Paragraph(IEnumerable<OpenXmlElement>)` with `Paragraph(object)` and
- `Paragraph(params OpenXmlElement[])` with `Paragraph(params object[])`.

ThomasBarnekow:
This commit implements the OpenXmlElementFunctionalExtensions class and
related unit tests.

Fixes #634.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Version 2.11.0
+### Added
+- Added `OpenXmlElementFunctionalExtensions.With` extension methods, which offer flexible means for constructing `OpenXmlElement` instances in the context of pure functional transformations.
+
 ## Version 2.10.0
 ### Added
 - Added initial Office 2016 support, including `FileFormatVersion.Office2016`, `ExtendedChartPart` and other new schema elements (#586)
--- a/src/DocumentFormat.OpenXml/OpenXmlElementFunctionalExtensions.cs
+++ b/src/DocumentFormat.OpenXml/OpenXmlElementFunctionalExtensions.cs
@@ -1,0 +1,94 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections;
+
+namespace DocumentFormat.OpenXml
+{
+    /// <summary>
+    /// Provides extension methods for pure functional transformations.
+    /// </summary>
+    public static class OpenXmlElementFunctionalExtensions
+    {
+        /// <summary>
+        /// Adds child elements or attributes to the given element.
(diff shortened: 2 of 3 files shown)


twsouthwick:
This looks great! Can you update the Changelog with this? I'll merge once that's in

twsouthwick:
I'm glad this method works without adding so many new constructors. We can explore that option, but if this works, that simplifies the addition of 1000s of new public APIs :)

ThomasBarnekow:
Just out of curiosity, what exactly would be the problem? If we added new constructors, those would all be implemented in the exact same way, making automatic generation easy. Not sure whether we could replace existing constructors without breaking things.

ThomasBarnekow:
And, yes, I will update the Changelog.

ThomasBarnekow:
@twsouthwick, regarding the [CHANGELOG](https://github.com/OfficeDev/Open-XML-SDK/blob/master/CHANGELOG.md), to which version will this be added? 2.11.0 (new feature) or 2.10.1?

twsouthwick:
It'll be 2.11.0 - we follow SemVer and this adds new surface area. Any revision (ie 2.10.x) is done as a hotfix-style (see the tag for 2.9.1 where we cherry-picked a fix for .NET Native)

twsouthwick:
The issue is just the number of APIs - not necessarily horrible, but it locks it in as we try not to remove APIs for backwards compatibility.

PR #685, opened 2020-02-20
East Asia is better term

twsouthwick:
Fix for documentation term with update in schema generator.


twsouthwick:
Fixes #342 again as it had never been fixed in the code generator. That is done now.

PR #695, opened 2020-02-28
Add FileFormatVersions.Office2019 for completeness

twsouthwick:


twsouthwick:
@tomjebo Thoughts? There's nothing in the SDK unique for 2019, but at least then someone won't think "I'm targeting 2019 but there's no enum value, so maybe this doesn't work".

PR #698, opened 2020-03-10
Add Obsolete to OpenXmlAttribute setters

twsouthwick:
This change takes advantage of a C# 8.0 features that allows obsolete messages for setters independent of the getter. Since `OpenXmlAttribute` is an attribute, we should remove the setters from it so the attribute becomes immutable. That's a breaking change, so we'll mark it as obsolete, and the next major version will remove it.

### DIFF
--- a/src/DocumentFormat.OpenXml/OpenXmlAttribute.cs
+++ b/src/DocumentFormat.OpenXml/OpenXmlAttribute.cs
@@ -13,6 +13,8 @@ namespace DocumentFormat.OpenXml
     /// </summary>
     public struct OpenXmlAttribute : IEquatable<OpenXmlAttribute>
     {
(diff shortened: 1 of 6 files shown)


src/DocumentFormat.OpenXml/OpenXmlAttribute.cs
```diff
@@ -13,6 +13,8 @@ namespace DocumentFormat.OpenXml
     /// </summary>
     public struct OpenXmlAttribute : IEquatable<OpenXmlAttribute>
```
lindexi: Is it possible marks as ReadOnly in future?

twsouthwick: Not that I can think of. I'm also not certain there's much of a value - adding `readonly` as a struct isn't breaking, so it's something we can do once the setters are removed (see https://blog.paranoidcoding.com/2019/03/27/readonly-struct-breaking-change.html)

PR #700, opened 2020-03-13
Generate code with new roslyn based generator

twsouthwick:


twsouthwick:
FYI - this looks to be a huge change, but the actual compiled code is the same. An API comparison shows that now APIs (internal nor public) have changed. It's mostly a ton of whitespace changes that Roslyn does (which is much cleaner and well-formatted).

PR #705, opened 2020-03-20
Use Cache.ReadOnlyCollection<OpenXmlAttribute> for OpenXmlDomReader.Attributes

lindexi:
The original code used the List type. This type will request a piece of memory by default.

Because we also need to support the .NET Framework versions below 4.6, and these versions do not exist Array.Empty method. See https://docs.microsoft.com/en-us/dotnet/api/system.array.empty

The `_emptyReadOnlyList` field  is only used here as the return value


src/DocumentFormat.OpenXml/OpenXmlDomReader.cs
```diff
@@ -89,7 +89,9 @@ public override ReadOnlyCollection<OpenXmlAttribute> Attributes
                 {
                     if (_emptyReadOnlyList == null)
                     {
-                        _emptyReadOnlyList = new ReadOnlyCollection<OpenXmlAttribute>(new List<OpenXmlAttribute>());
+#pragma warning disable CA1825 // Because we also need to support the .NET Framework versions below 4.6, and these versions do not exist Array.Empty method
```
twsouthwick: There is a utility for that in the SDK: `Cached.Array<OpenXmlAttribute>()`.

lindexi: @twsouthwick Thank you and I fixed it

src/DocumentFormat.OpenXml/OpenXmlDomReader.cs
```diff
@@ -89,7 +90,7 @@ public override ReadOnlyCollection<OpenXmlAttribute> Attributes
                 {
                     if (_emptyReadOnlyList == null)
                     {
-                        _emptyReadOnlyList = new ReadOnlyCollection<OpenXmlAttribute>(new List<OpenXmlAttribute>());
+                        _emptyReadOnlyList = Cached.ReadOnlyCollection<OpenXmlAttribute>();
```
twsouthwick: No need to store it in a field - you can just return the `Cached.ReadOnlyCollection`

twsouthwick:
Build's failing because usings aren't sorted. Sorry for the back and forth. We're working on getting the builds visible; we have warnings as errors, so if you see any errors locally, it'll fail.

lindexi:
@twsouthwick Thank you. You are very fast, I haven't compiled it locally, because my Internet speed is very poor, I'm still waiting for NuGet to restore

PR #706, opened 2020-03-23
Add missing values to `CompatSettingNameValues`

twsouthwick:
Fixes #451

PR #707, opened 2020-03-23
Ensure minimum office versions are included on enums

twsouthwick:


twsouthwick:
The schema generator hadn't been correctly identifying when an enum is not starting with 2007. This includes the updates for that, as well as a few changes to ensure that version is used correctly.

Related to #706

PR #716, opened 2020-05-11
Rename StrictTranslation to StrictRelationshipFound and make public

### LINKED ISSUE #711: Make StrictTranslation property public

Is this a:
- [X] Issue with the OpenXml library

**Description**

Please provide a simple description of the issue encountered.

**Information**

- .NET Target: All
- Latest commit on `master`

In https://github.com/OfficeDev/Open-XML-SDK/blob/0f1764a866e36d615e22d247e780fb7675046e9a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.cs#L240 `StrictTranslation` is internal. 

Would it be possible to make this property public (at least the `get`)?

Knowing whether a package was saved in strict or non-strict mode eases the subsequent parsing. For example, in strict mode, dates are stored in `yyyy-MM-dd` format and in non-strict the serial number is stored.

igitur:
Fixes #711 

This PR firstly renames `StrictTranslation` to `StrictRelationshipFound` and secondly makes the `OpenXmlPackage.StrictRelationshipFound` property public (from internal).

This will provide a hint to subsequent parsers that it may need to support strict mode parsing.

### DIFF
--- a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.cs
@@ -240,7 +240,7 @@ namespace DocumentFormat.OpenXml.Packaging
         /// <summary>
         /// Gets a value indicating whether this package contains Transitional relationships converted from Strict.
         /// </summary>
-        public bool StrictRelationshipFound { get; internal set; } = false;
+        public bool StrictRelationshipFound { get; private set; } = false;
 
         /// <summary>
         /// Gets the package of the document.
--- a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.cs
@@ -181,13 +181,13 @@ namespace DocumentFormat.OpenXml.Packaging
                 bool hasMainPart = false;
                 RelationshipCollection relationshipCollection = new PackageRelationshipPropertyCollection(_package);
 
-                // relationCollection.StrictTranslation is true when this collection contains Transitional relationships converted from Strict.
-                StrictTranslation = relationshipCollection.StrictTranslation;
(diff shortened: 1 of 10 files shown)


src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.cs
```diff
@@ -237,7 +237,7 @@ private void Load()
 
         internal virtual ApplicationType ApplicationType => ApplicationType.None;
 
-        internal bool StrictTranslation { get; set; } = false;
+        public bool StrictRelationshipFound { get; set; } = false;
```
twsouthwick, line 240: This needs XML docs, and the `set` needs to be `internal`

twsouthwick:
Also, can you update the CHANGELOG?

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.cs
igitur: Done.

igitur:
> Also, can you update the CHANGELOG?

Would this go under `Added` or `Changed`?

twsouthwick:
It's an addition, so under "Added"

twsouthwick:
Just pushed a change to make the property private instead of internal. Once build finishes, I'll merge it in

igitur:
I'd like to squash that commit into the previous to keep the history cleaner.

twsouthwick:
I swash and merge everything on each PR :)

Asbjoedt:
Hi! @igitur, @twsouthwick, @tomjebo 

Can you specify what this property does in practice?

The OOXML standard has several areas that are specific to Strict. See this for a quick'n'dirty overview: https://github.com/mikeebowen/ooxml-validator-vscode/issues/14#issuecomment-1057831570

Will the StrictRelationshipFound property guarantee that the document is conformant with OOXML Strict? How does it check for this, and does the property check for all those things listed in the linked post?

PR #721, opened 2020-05-19
Move schematron data to element class

### LINKED ISSUE #678: Convert semantic validation registration to declarative structure within OpenXmlElement
There is currently a semantic constraint registry that contains all the semantic constraints that get applied during validation. This should be separated out so that each element defines their own semantic constraints. The following must be applied:

- [ ] AttributeCannotOmitConstraint.cs
- [ ] AttributeMinMaxConstraint.cs
- [ ] AttributePairConstraint.cs
- [ ] AttributeValueLengthConstraint.cs
- [ ] AttributeValuePatternConstraint.cs
- [ ] AttributeValueRangeConstraint.cs
- [ ] AttributeValueSetConstraint.cs
- [ ] IndexReferenceConstraint.cs
- [ ] ParentTypeConstraint.cs
- [ ] ReferenceExistConstraint.cs
- [ ] RelationshipExistConstraint.cs
- [ ] RelationshipTypeConstraint.cs
- [ ] SemanticConstraint.cs
- [ ] UniqueAttributeValueConstraint.cs

We'll follow the pattern used by particles, ie something like:

```csharp
public class SomeOpenXmlElement : OpenXmlElement
{
  private static readonly SemanticConstraints[] _constraints = new[]
  {
    new ReferenceExistConstraint(...)
  }

  internal override SemanticConstraint[] SemanticConstraints => _constraints;
}
```

twsouthwick:
Previously, this data was in a separate registry and not easily
correlated with the OpenXml element itself. This change moves it to be
part of the class. This is the last major change needed to separate
assemblies by application type (Word/Excel/PowerPoint).

Fixes #678


src/DocumentFormat.OpenXml/GeneratedCode/schemas-microsoft-com_office_excel.g.cs
```diff
@@ -199,6 +200,13 @@ public ClientData(string outerXml) : base(outerXml)
         [Index(0)]
         public EnumValue<DocumentFormat.OpenXml.Vml.Spreadsheet.ObjectValues> ObjectType { get; set; }
 
+        private static readonly ISemanticConstraint[] _semanticConstraint = new ISemanticConstraint[] {
```
lindexi, line 203: In [corefx/coding-style.md at master · dotnet/corefx](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md ) NO.1, we use Allman style braces, where each brace begins on a new line. A single line statement block can go without braces but the block must be properly indented on its own line and must not be nested in other statement blocks that use braces (See issue 381 for examples). One exception is that a using statement is permitted to be nested within another using statement by starting on the following line at the same indentation level, even if the nested using contains a controlled block.

src/DocumentFormat.OpenXml/GeneratedCode/schemas-microsoft-com_office_office.g.cs
```diff
@@ -583,6 +584,12 @@ public SignatureLine() : base()
         [Index(12)]
         public StringValue SignatureProviderUrl { get; set; }
 
+        private static readonly ISemanticConstraint[] _semanticConstraint = new ISemanticConstraint[] {
+            new UniqueAttributeValueConstraint(2 /*:id*/, true, null)
```
lindexi, line 588:
```suggestion
            new UniqueAttributeValueConstraint(id: 2, true, null)
```

src/DocumentFormat.OpenXml/GeneratedCode/schemas-microsoft-com_office_excel.g.cs
twsouthwick: Yup, and I spent about 20 minutes trying to get the generator to produce that to no avail. Decided in this case it's fine; I have future plans to streamline the declaration of this metadata even more, so it'll change again in the next few months.

PR #722, opened 2020-05-19
Inherit parts to child parts

### LINKED ISSUE #720: Adding parts to WorksheetPart derived class throws exception
**Description**
When I try to add parts to a worksheet part within a class derived from WorksheetPart, such as SpreadsheetPrinterSettingsPart or VmlDrawingPart, an OpenXmlPackageException is thrown with a message "The part cannot be added here."  This just started when I updated the reference SDK to 10.0 or 10.1.  The code works as expected when built and run against SDK version 9.1 or earlier.

I'm still trying different things to understand this issue better but this is what I am observing in my code.  However, if I open and existing workbook and simply add the part to a worksheet, it works fine.  The main difference that I can tell is that the error happens in my code that inherits from WorksheetPart.  Please let me know if there are any additional questions.

**Information**
- netstandard2.0
- DocumentFormat.OpenXml Version: 10.0+

**Repro**

```csharp
// Please add a self-contained, minimum viable repro of the issue.
// If you require external resources, please provide a gist or GitHub repro
// An Xunit style test is preferred, but a console application would work too.
public class SampleWorksheetPart : WorksheetPart
{
  public SampleWorksheetPart() {}

  public void AddPrinterSettingsPart()
  {
    // The error would occur when this method is called in the line
    // below in 10.0+ but works as expected
    // in 9.1 or earlier.
    var psPart = this.AddNewPart<SpreadsheetPrinterSettingsPart>();
  }
}

class Program
{
  {
    var fInfo = new FileInfo(@"C:\Temp\NewWorkbook.xlsx");
    // This assumes that file that fInfo references already exists
    using (var xlsx = fInfo.Open(FileMode.Open, FileAcess.ReadWrite))
    {
      using (var doc = SpreadsheetDocument.Open(xlsx, true))
      {
        var workbook = doc.WorkbookPart;
        var worksheet = workbook.Parts.FirstOrDefault(w => w.OpenXmlPart is WorksheetPart).OpenXmlPart;

        // This works fine in all versions for some reason.
        var ps = worksheet.AddNewPart<SpreadsheetPrinterSettingsPart>();
      }
    }
  }
}
``` 
**Observed**
OpenXmlPackageException: The part cannot be added here
  --> Inner exception: ArgumentOutOfRangeException: Specified argument was out of the range of valid values.

**Expected**

The part is added to the current object.

twsouthwick:
Fixes #720

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added minimum Office versions for enum types and values (#707)
 - Added additional `CompatSettingNameValues` values: `UseWord2013TrackBottomHyphenation`, `AllowHyphenationAtTrackBottom`, and `AllowTextAfterFloatingTableBreak` (#706)
 - Added gfxdata attribue to Arc, Curve, Line, PolyLine, Group, Image, Oval, Rect, and RoundRect shape complex types per MS-OI29500 2.1.1783-1799 (#709)
-- Added `OpenXmlPartContainer.TryGetPartById` to enable chidl part retrieval without exception if it does not exist (#714)
+- Added `OpenXmlPartContainer.TryGetPartById` to enable child part retrieval without exception if it does not exist (#714)
 - Added `OpenXmlPackage.StrictRelationshipFound` property that indicates whether this package contains Transitional relationships converted from Strict (#716)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added `FileFormatVersions.2019` enum (#695)
 - Added `ChartSpace` and chart elements for the new 2016 namespaces. This allows the connecting pieces for building a chart part with chart styles like "Sunburst" (#687).
 - Added `OpenXmlElementFunctionalExtensions.With(...)` extension methods, which offer flexible means for constructing `OpenXmlElement` instances in the context of pure functional transformations (#679)
-- Added minimum Office versions for enum types and values (#707).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,11 +9,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
-- Added minimum Office versions for enum types and values (#707).
-- Added additional `CompatSettingNameValues` values: `UseWord2013TrackBottomHyphenation`, `AllowHyphenationAtTrackBottom`, and `AllowTextAfterFloatingTableBreak` (#706).
+- Added minimum Office versions for enum types and values (#707)
+- Added additional `CompatSettingNameValues` values: `UseWord2013TrackBottomHyphenation`, `AllowHyphenationAtTrackBottom`, and `AllowTextAfterFloatingTableBreak` (#706)
 - Added gfxdata attribue to Arc, Curve, Line, PolyLine, Group, Image, Oval, Rect, and RoundRect shape complex types per MS-OI29500 2.1.1783-1799 (#709)
-- Added `TryGetPartById` in `OpenXmlPartContainer`. This allows try to get the child part by the relationship ID (#714)
-- Added `StrictRelationshipFound` property on `OpenXmlPackage` indicating whether this package contains Transitional relationships converted from Strict. (#716)
+- Added `OpenXmlPartContainer.TryGetPartById` to enable chidl part retrieval without exception if it does not exist (#714)
```
lindexi, line 15:
```suggestion
- Added `OpenXmlPartContainer.TryGetPartById` to enable child part retrieval without exception if it does not exist (#714)
```

twsouthwick: good catch!

twsouthwick:
Thanks @lindexi for getting involved in these! It's great to have more eyes on these changes :)

PR #740, opened 2020-06-22
Alter DeletePartCore to use TryGetPartById

### LINKED ISSUE #713: The OpenXmlPartContainer.DeletePart will throw exception when the id not found

Is this a:
- [x] Issue with the OpenXml library

**Description**

As we can see the OpenXmlPartContainer.DeletePart can delete the part, but this API will throw ArgumentOutOfRangeException when the part with the specified id does not exist.

Because the `DeletePart` call the `DeletePartCore` which will call the `GetPartById` to get the part.

https://github.com/OfficeDev/Open-XML-SDK/blob/0f1764a866e36d615e22d247e780fb7675046e9a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPartContainer.cs#L1762

But the `GetPartById` will throw ArgumentOutOfRangeException when the part with the specified id does not exist.

https://github.com/OfficeDev/Open-XML-SDK/blob/0f1764a866e36d615e22d247e780fb7675046e9a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPartContainer.cs#L529-L547

Yes, it means that the branch will never enter when the id not found

https://github.com/OfficeDev/Open-XML-SDK/blob/0f1764a866e36d615e22d247e780fb7675046e9a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPartContainer.cs#L1764-L1768

**Information**

- .NET Target: All 
- DocumentFormat.OpenXml Version: 2.10.1

**Repro**

```csharp
public static void Delete(this string id, OpenXmlPart part)
{
     part.DeletePart(id);
}
``` 

**Observed**

The `part.DeletePart` will throw the exception when the part with the specified id does not exist.

**Expected**

As the other API design, the delete method should not throw exception the element does not exist.

stoiveyp:
Fixes #713, test included

### DIFF
--- a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPartContainer.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/OpenXmlPartContainer.cs
@@ -1777,11 +1777,8 @@ namespace DocumentFormat.OpenXml.Packaging
             var liveParts = new Dictionary<OpenXmlPart, bool>();
             var processedParts = new Dictionary<OpenXmlPart, bool>();
 
-            var child = GetPartById(id);
-
-            if (child == null)
+            if(!TryGetPartById(id, out var child) || child == null)
             {
-                // TODO: should we throw? or just return?
                 return false;
             }
 
--- a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPartContainer.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/OpenXmlPartContainer.cs
@@ -1777,7 +1777,7 @@ namespace DocumentFormat.OpenXml.Packaging
             var liveParts = new Dictionary<OpenXmlPart, bool>();
             var processedParts = new Dictionary<OpenXmlPart, bool>();
 
-            if(!TryGetPartById(id, out var child) || child == null)
+            if(!TryGetPartById(id, out var child))
             {
                 return false;
             }
--- a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPartContainer.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/OpenXmlPartContainer.cs
@@ -1781,7 +1781,7 @@ namespace DocumentFormat.OpenXml.Packaging
             {
                 return false;
             }
-            
+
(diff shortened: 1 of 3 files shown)


src/DocumentFormat.OpenXml/Packaging/OpenXmlPartContainer.cs
```diff
@@ -1777,11 +1777,8 @@ internal bool DeletePartCore(string id)
             var liveParts = new Dictionary<OpenXmlPart, bool>();
             var processedParts = new Dictionary<OpenXmlPart, bool>();
 
-            var child = GetPartById(id);
-
-            if (child == null)
+            if(!TryGetPartById(id, out var child) || child == null)
```
twsouthwick, line 1780: child be null if it failed; any reason why you need to check it again?

stoiveyp: My bad, fixed now 👍

test/DocumentFormat.OpenXml.Tests/ofapiTest/OpenXmlPartTest.cs
```diff
@@ -254,5 +254,20 @@ public void AddPartTest()
...
+        public void DeleteInvalidPartIdSafely()
+        {
+            using (Stream stream = new MemoryStream())
+            {
+                using (var testDocument = WordprocessingDocument.Create(stream, WordprocessingDocumentType.Document))
+                {
+                    var testId = "invalidId";
+                    var mainDocPart = testDocument.AddMainDocumentPart();
+                    Assert.False(mainDocPart.TryGetPartById(testId, out _));
+                    mainDocPart.DeletePart(testId);
```
twsouthwick, line 268: Can you Assert.False here as well?

twsouthwick:
Thanks @stoiveyp this is great! Once #744 is added, can you update the CHANGELOG with your change?

twsouthwick:
Ok - #744 is merged. Can you update the CHANGELOG and we can get this in :)

stoiveyp:
@twsouthwick commit got confused there for a moment, but changelog is now updated 👍

PR #741, opened 2020-06-22
Limit milliseconds to three digits

### LINKED ISSUE #724: CellValue ctor for DateTime and DateTimeOffset should use proper xml standard format
Issue with the OpenXml library

**Description**

CellValue created via DateTime constructor does not serialize xml string properly.
With the latest version of Excel the document xml has no problems loading, however with Excel 2016 it fails to load the xml value.

Further inspection showed the issue was with "midnight" date cell value "2019-09-25T00:00:00.0000000". 
If you trim down the miliseconds, it works, 
If you add actual miliseconds, it works.

To workaround the issue I have used the string constructor with date formatted via `XmlConvert.ToString`

For example:

```
new DateTime(2019, 9, 25, 0, 0, 0)
    should be
"2019-09-25T00:00:00"
    but was
"2019-09-25T00:00:00.0000000"
```

```
new DateTime(2019, 9, 25, 0, 0, 0).AddMilliseconds(123)
    should be
"2019-09-25T00:00:00.123"
    but was
"2019-09-25T00:00:00.1230000"
```

stoiveyp:
Fixes #724 and adds test to ensure that's all that changes
No standard format limits to three digits for milliseconds, but the format is the same as `"o"` used before in all other regards

### DIFF
--- a/src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
+++ b/src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
@@ -17,7 +17,7 @@ namespace DocumentFormat.OpenXml.Spreadsheet
         /// </summary>
         /// <param name="dateTime">DateTime for cell</param>
         public CellValue(DateTime dateTime)
-            : this(dateTime.ToString("o", CultureInfo.InvariantCulture))
+            : this(dateTime.ToString("yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fff", CultureInfo.InvariantCulture))
         {
         }
 
@@ -27,7 +27,7 @@ namespace DocumentFormat.OpenXml.Spreadsheet
         /// </summary>
         /// <param name="dateTimeOffset">DateTime for cell</param>
         public CellValue(DateTimeOffset dateTimeOffset)
-            : this(dateTimeOffset.ToString("o", CultureInfo.InvariantCulture))
+            : this(dateTimeOffset.ToString("yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffzzz", CultureInfo.InvariantCulture))
         {
         }
     }
--- a/src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
+++ b/src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
@@ -17,7 +17,7 @@ namespace DocumentFormat.OpenXml.Spreadsheet
         /// </summary>
         /// <param name="dateTime">DateTime for cell</param>
         public CellValue(DateTime dateTime)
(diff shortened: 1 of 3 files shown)


twsouthwick:
@tomjebo can you review this as it's more up your alley. Is the form consistent with the spec? Let's add the section from the spec for reference

twsouthwick:
@stoiveyp can you update the changelog to include this change? You may need to merge the latest changes from master if you don't have a 2.11.1 spot for the changes.

src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
```diff
@@ -17,7 +17,7 @@ public partial class CellValue
         /// </summary>
         /// <param name="dateTime">DateTime for cell</param>
         public CellValue(DateTime dateTime)
-            : this(dateTime.ToString("o", CultureInfo.InvariantCulture))
+            : this(dateTime.ToString("yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fff", CultureInfo.InvariantCulture))
```
twsouthwick, line 20: I'd prefer if this was a const, ie

```cs
private const string DateTimeFormatString = "....";
private const string DateTimeOffsetFormatString = DateTimeFormatString + "zzz";
```

It would make it much easier to see that they're pretty much the same.

twsouthwick:
Also, would it make sense to provide a parse function? We want to get a revision version out which we don't expose new APIs, but we could expose it on the next minor version

twsouthwick:
Could we use `XmlConvert.ToString` instead? Wouldn't that be the same, but not require a custom format string

stoiveyp:
@twsouthwick sorry for the delay - day job has worn me out this week! 
Did look at XmlConvert, but it also goes down to too many significant digits in the seconds fraction, so we end up having to customise there too.
I'll make the other changes now though 👍

stoiveyp:
@twsouthwick moved the string conversion to static methods in CellValue and moved the formats to constants
Changelog is also updated


Can't find anything in the OpenXml spec specifically about date time cell values.
So I looked in the W3 Xml Schema documents
https://www.w3.org/TR/xmlschema-2/
```
All ·minimally conforming· processors ·must· support year values with a minimum of 4 digits (i.e., YYYY) and a minimum fractional second precision of milliseconds or three decimal digits (i.e. s.sss). However, ·minimally conforming· processors ·may· set an application-defined limit on the maximum number of digits they are prepared to support in these two cases, in which case that application-defined maximum number ·must· be clearly documented.
```

So without comment within the OpenXml spec or Excel documentation - a processor based on the Xml Schema spec would only have to support the three digits we're now outputting.

twsouthwick:
LGTM! I've opened #752 to add a parsing method so people don't need to know the value to get a strongly typed value. That will have to go into a minor version (which may end up being the next version... there's a few APIs we'll want to expose soon).

PR #757, opened 2020-07-14
Remove unused ElementLookup.Contains method

lindexi:
But now there's no code that uses the `Contains` method.


twsouthwick:
Feel free to remove dead code :)

lindexi:
@twsouthwick Thank you and I will remove it.

PR #758, opened 2020-07-14
Avoid multi-thread creates ElementMetadata object

lindexi:
The origin code is not thread-safe and this is the test code

```csharp
    public class ElementMetadataTests
    {
        [Fact]
        public void Create()
        {
            using ManualResetEvent resetEvent = new ManualResetEvent(false);

            List<Thread> threadList = new List<Thread>();
            var maxCount = 100;
            var elementMetadataList = new ElementMetadata[maxCount];
            for (var i = 0; i < maxCount; i++)
            {
                var n = i;
                threadList.Add(new Thread(() =>
                {
                    resetEvent.WaitOne();
                    var alternateContent = new AlternateContent();
                    var metadata = ElementMetadata.Create(alternateContent);
                    elementMetadataList[n] = metadata;
                })
                {
                    IsBackground = true,
                });
            }

            foreach (var thread in threadList)
            {
                thread.Start();
            }

            resetEvent.Set();
            foreach (var thread in threadList)
            {
                thread.Join();
            }

            ElementMetadata elementMetadata = elementMetadataList[0];
            foreach (var temp in elementMetadataList)
            {
                if (!ReferenceEquals(elementMetadata, temp))
                {
                    // If it enter this branch, means the ElementMetadata object is created more than once
                    break;
                }
            }
        }
    }
```

### DIFF
--- a/src/DocumentFormat.OpenXml/Framework/Metadata/ElementMetadata.cs
+++ b/src/DocumentFormat.OpenXml/Framework/Metadata/ElementMetadata.cs
@@ -56,17 +56,8 @@ namespace DocumentFormat.OpenXml.Framework.Metadata
         {
             var type = element.GetType();
 
-            // Use TryGetValue first for the common case of already existing types to limit number of allocations
-            if (_lookup.TryGetValue(type, out var result))
-            {
-                return result;
-            }
-
-            var metadata = CreateInternal(element);
-
-            _lookup.TryAdd(type, metadata);
-
-            return metadata;
+            // Use GetOrAdd first for the common case of already existing types to limit number of allocations
+            return _lookup.GetOrAdd(type, _ => CreateInternal(element));
         }
 
         public static ElementMetadata Create<TElement>()


twsouthwick:
Is this a correctness issue? I'd imagine that this potentially creates multiple instances of the metadata, but otherwise fine. Can you add a benchmark test to see what kind of allocations/perf changes this causes? The current way doesn't use the GetOrAdd as it would create a closure and since this is potentially a common pathway, I didn't want to do that.

lindexi:
@twsouthwick  Thank you. And the `public static ElementMetadata Create<TElement>()` method is use GetOrAdd 

And I think the both of the `Create` method should use the same logic code.

twsouthwick:
The generic one does not create a closure, so the `GetOrAdd` is fine. The point is to minimize allocations if possible as this is a common pathway anytime an element is made. When I added this in, I found it really helped with performance to do it this way. If you have some numbers to show otherwise, I'm fine accepting the change.

lindexi:
@twsouthwick Thank you and I write the Benchmark code in https://github.com/dotnet-campus/Open-XML-SDK/commit/aaa16caf666e24f0f72a9245318276de466f3c73

And the `Create` method is the origin code and the `Create2` use GetOrAdd 

|  Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
|  Create | 22.19 ns | 0.154 ns | 0.144 ns |      - |     - |     - |         - |
| Create2 | 37.22 ns | 0.337 ns | 0.315 ns | 0.0210 |     - |     - |      88 B |

I can find the `Create2` that use GetOrAdd will allocated 88B

Thank you and I will close this PR

twsouthwick:
Cool. Thanks for the verification. It's also 50% slower. Can you add that to our benchmark suite?

PR #764, opened 2020-07-16
Add Benchmark for create metadata

lindexi:
https://github.com/OfficeDev/Open-XML-SDK/pull/758#issuecomment-658772892


test/DocumentFormat.OpenXml.Benchmarks/ElementMetadataTests.cs
```diff
@@ -0,0 +1,31 @@
...
+        [GlobalSetup]
+        public void Setup()
+        {
+            _element = new AlternateContent();
+        }
+
+        [Benchmark]
+        public void CreateWithInstance()
+        {
+            _ = ElementMetadata.Create(_element);
```
twsouthwick, line 20: Can you have this return the metadata? You can have it return cast as `object` as it's an internal type if needed. The release builds may optimize away the call if not returned.

PR #773, opened 2020-07-19
Add CancellationToken to validation methods

### LINKED ISSUE #761: Integrate cancellation tokens in validation system
There is a setting for max errors, but sometimes the validation can just take a long time. While we can explore perf issues, it would also be good to be able to cancel it using CancellationToken so all the built in ways to manage workflows with them can be used.

Inspired by #760

twsouthwick:
Fixes #761


src/DocumentFormat.OpenXml/Validation/DocumentValidator.cs
```diff
@@ -38,10 +39,11 @@ public DocumentValidator(SchemaValidator schemaValidator, SemanticValidator sema
         /// </summary>
         /// <param name="document">The document to be validated.</param>
         /// <param name="settings">The settings to be used during validation.</param>
+        /// <param name="token"></param>
```
lindexi, line 42: I think we should add comments to tell the developer that the `token` do nothing in .NET 3.5

twsouthwick: Thanks. This method is private in. Net 3.5 so it won't be visible on that platform as there is no cancelation token there. Or are you referring to devs in the open xml codebase?

lindexi: @twsouthwick Yes, for developers.

twsouthwick: Sure Good idea - I'll document that

src/DocumentFormat.OpenXml/Validation/ValidationContext.cs
```diff
@@ -47,7 +49,15 @@ public ValidationContext(ValidationSettings settings, ValidationCache cache)
 
         public bool Valid => Errors.Count == 0;
 
-        public bool IsCancelled => MaxNumberOfErrors > 0 && Errors.Count >= MaxNumberOfErrors;
+        public bool IsCancelled
+        {
+            get
+            {
+                _token.ThrowIfCancellationRequested();
```
lindexi, line 56: I do not recommend throwing an exception in the get method. From the name of this property, we may not expect any exception selling.

twsouthwick: Hmm good point as to not throwing there. I'll change it to a method

Issue #492, opened 2018-04-29
Worksheet class lacks most child element-related properties


ThomasBarnekow:
**Description**

The `DocumentFormat.OpenXml.Spreadsheet.Worksheet` class only has properties for 4 out of 38 child elements. For example, there is no property for the only mandatory child element, `x:sheetData`, which is likely among the most frequently referenced elements.

As per the schema, those child elements have to be in a defined order, which, in the absence of SDK support, needs to be ensured by the caller's code. That increases the size and complexity of the code.

**Information**

- .NET Target: all
- DocumentFormat.OpenXml Version: 3cc26570f2964a5d56a5988f9464b2e4b29812c7

**Observed**

The `Worksheet` class defines the following child elements:
```csharp
    [ChildElementInfo(typeof(SheetProperties))]
    [ChildElementInfo(typeof(SheetDimension))]
    [ChildElementInfo(typeof(SheetViews))]
    [ChildElementInfo(typeof(SheetFormatProperties))]
    [ChildElementInfo(typeof(Columns))]
    [ChildElementInfo(typeof(SheetData))]
    [ChildElementInfo(typeof(SheetCalculationProperties))]
    [ChildElementInfo(typeof(SheetProtection))]
    [ChildElementInfo(typeof(ProtectedRanges))]
    [ChildElementInfo(typeof(Scenarios))]
    [ChildElementInfo(typeof(AutoFilter))]
    [ChildElementInfo(typeof(SortState))]
    [ChildElementInfo(typeof(DataConsolidate))]
    [ChildElementInfo(typeof(CustomSheetViews))]
    [ChildElementInfo(typeof(MergeCells))]
    [ChildElementInfo(typeof(PhoneticProperties))]
    [ChildElementInfo(typeof(ConditionalFormatting))]
    [ChildElementInfo(typeof(DataValidations))]
    [ChildElementInfo(typeof(Hyperlinks))]
    [ChildElementInfo(typeof(PrintOptions))]
    [ChildElementInfo(typeof(PageMargins))]
    [ChildElementInfo(typeof(PageSetup))]
    [ChildElementInfo(typeof(HeaderFooter))]
    [ChildElementInfo(typeof(RowBreaks))]
    [ChildElementInfo(typeof(ColumnBreaks))]
    [ChildElementInfo(typeof(CustomProperties))]
    [ChildElementInfo(typeof(CellWatches))]
    [ChildElementInfo(typeof(IgnoredErrors))]
[...]

PR #774, opened 2020-07-19
Expose `OpenXmlCompositeElement.TryAddChild`

twsouthwick:
This change adds an API to simplify adding child elements in the correct order according to the schema definition.

Part of #492

### DIFF
--- a/src/DocumentFormat.OpenXml/AppendOption.cs
+++ b/src/DocumentFormat.OpenXml/AppendOption.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace DocumentFormat.OpenXml
+{
+    /// <summary>
+    /// Options to define how an element should be appeneded.
+    /// </summary>
+    public enum AppendOption
+    {
+        /// <summary>
+        /// Append the item in list order without regard to schema.
+        /// </summary>
+        List = 0,
+
+        /// <summary>
+        /// Append the item in accordance with what the schema defines.
+        /// </summary>
+        Ordered = 1,
+    }
+}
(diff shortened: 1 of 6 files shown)


twsouthwick:
@ThomasBarnekow I'd love your thoughts on this. Does it help with the issues discussed in #492?

twsouthwick:
@onizet This API should be used instead of what was done for https://github.com/onizet/html2openxml/issues/70. Please note, that due to making the method non-generic, it will be a break for any users of your current version as the reflection code expects a generic method. This will be made available with the release of v2.12.0 (I haven't changed the internal implementation in 2.11.x so it wouldn't hurt your users). However, you'll want to make it work for either both generic and non-generic instances of the method, or quickly support v2.12.0 when it is released.

onizet:
Hi Taylor, this is very good news and finally a good way to conform to valid schema, without hacky ways. I will definitively publish a new version as soon as v2.12.0 points out.
Thanks!

igitur:
Hi,

This PR is a breaking change, specifically the https://github.com/OfficeDev/Open-XML-SDK/pull/774/files#diff-bfc8b437c919c6cfa4a52da72a36b29ac087ebdd0d7c2789043b6ce6bd396ff5R53 block.

```c#
private static void Main(string[] args)
{
    using var ssd = SpreadsheetDocument.Open(@"c:\temp\template.xlsx", false);
    var wsp = ssd.WorkbookPart.WorksheetParts.First(w => w.Uri.ToString().EndsWith("sheet1.xml"));
    var sheetdata = wsp.Worksheet.Elements<SheetData>().First();
    var row = sheetdata.Elements<Row>().First(r => r.RowIndex == 8);
    var cell = row.Elements<Cell>().First(C => C.CellReference == "C8");
    System.Console.WriteLine(cell.CellValue.Text);

    cell.CellValue = null;

    ssd.SaveAs(@"c:\temp\template2.xlsx");
}
```

The `cell.CellValue = null` statement has no effect since v2.12.0. It still worked in v2.11.3.

Should I log a new issue?

Issue #759, opened 2020-07-14
AttributeCollection's IsEmpty will alway be false


lindexi:
**Description**

The IsEmpty property in AttributeCollection will always be false, because the `_data` is never been null

As the Constructor code says, we will set the `_data` field a not null value

https://github.com/OfficeDev/Open-XML-SDK/blob/0977a9430407115ff9ed149f60496fbb2688b5e5/src/DocumentFormat.OpenXml/Framework/Metadata/AttributeCollection.cs#L15-L27

And the `IsEmpty` property will be false only when the `_data` is null

https://github.com/OfficeDev/Open-XML-SDK/blob/0977a9430407115ff9ed149f60496fbb2688b5e5/src/DocumentFormat.OpenXml/Framework/Metadata/AttributeCollection.cs#L29

twsouthwick:
Good catch! Is that used any where? Maybe can delete it or update to check for `.Length == 0`

lindexi:
@twsouthwick I can not find any code use it and I think we can delete it

twsouthwick:
You can add that into the PR you're doing to remove the `Contains` method.

I've been revamping the infrastructure to consolidate concepts and help reason about metadata. Some of these methods were used at different points. Thanks for exploring in here!

lindexi:
> 
> 
> @twsouthwick I can not find any code use it and I think we can delete it



@twsouthwick I find the `ElementState` will use the `AttributeCollection.IsEmpty` property. And I can not delete it.

lindexi:
@twsouthwick  I seem to have found an null reference exception, could you help me?

The `ElementMetadata.Builder.BuildAttributes` will return null when the `_attributes` is null.

```csharp
            private AttributeMetadata[] BuildAttributes()
            {
                if (_attributes is null)
                {
                    return default;
                }
                // Ignore code
            }
```

And it will make the ElementMetadata receive the null argument, and that means the `ElementMetadata.Attributes` may be null.

```csharp
        internal ElementMetadata(
            ReadOnlyArray<AttributeMetadata> attributes,
            ReadOnlyArray<IValidator> validators,
            ReadOnlyArray<IValidator> constraints,
            FileFormatVersions version,
            SchemaAttrAttribute schema,
            CompiledParticle particle,
            Lazy<ElementLookup> lookup)
        {
            Attributes = attributes; // may be null, when the ElementMetadata.Builder.BuildAttributes return null
            Validators = validators;
[...]

PR #781, opened 2020-07-23
Fix the IsEmpty Property

lindexi:
The _data is never null

see https://github.com/OfficeDev/Open-XML-SDK/issues/759


lindexi:
Why we need the IsEmpty property? Because this property is used to indicate that the structure is not initialized.

The structure can be created without going through the constructor.

PR #783, opened 2020-07-25
Use WindowsBase instead of package on .NET Framework

twsouthwick:
Currently we're using System.IO.Packaging NuGet package on .NET 4.6+,
when this isn't really necessary. This ends up just being a bunch of
type forwarders that people don't really need on .NET Framework. We can
just directly reference WindowsBase instead.

This should be a non-breaking change as all the type-forwarders would be
brought in by another dependency if they require them.


twsouthwick:
We'll want to verify this with the pre-release packages.

PR #786, opened 2020-07-30
Rename TryAddChild to AddChild with option for exception

twsouthwick:
This also removes the generic constraint as that is unnecessary and fails to work in the instances where you have just a `OpenXmlElement`.


twsouthwick:
An update to #774

PR #789, opened 2020-08-04
Small performance improvements for AddNamespaceDeclaration

lindexi:
|                     Method |     Mean |     Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------- |---------:|----------:|---------:|------:|------:|------:|----------:|
|    AddNamespaceDeclaration | 3.665 us | 0.5911 us | 1.658 us |     - |     - |     - |         - |
| OldAddNamespaceDeclaration | 4.303 us | 0.8884 us | 2.462 us |     - |     - |     - |         - |

Because the method execution time is too short, it is difficult to observe the intensity of optimization.

Benchmark code: https://github.com/dotnet-campus/Open-XML-SDK/blob/5d5c307a55066fccb3913d28bf945b51e0a6fcde/test/DocumentFormat.OpenXml.Benchmarks/OpenXmlElementAddNamespaceDeclarationTests.cs

### DIFF
--- a/src/DocumentFormat.OpenXml/OpenXmlElement.cs
+++ b/src/DocumentFormat.OpenXml/OpenXmlElement.cs
@@ -752,6 +752,40 @@ namespace DocumentFormat.OpenXml
         }
 
         /// <summary>
+        /// Fx
+        /// </summary>
+        /// <param name="prefix"></param>
+        /// <param name="uri"></param>
+        public void OldAddNamespaceDeclaration(string prefix, string uri)
+        {
+            if (string.IsNullOrEmpty(prefix))
+            {
+                throw new ArgumentNullException(nameof(prefix));
+            }
+
+            if (string.IsNullOrEmpty(uri))
+            {
+                throw new ArgumentNullException(nameof(uri));
+            }
+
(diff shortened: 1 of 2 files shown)


src/DocumentFormat.OpenXml/OpenXmlElement.cs
```diff
@@ -737,12 +737,14 @@ public void AddNamespaceDeclaration(string prefix, string uri)
             {
                 NamespaceDeclField = new List<KeyValuePair<string, string>>();
             }
-
-            foreach (var item in NamespaceDeclField)
+            else
```
twsouthwick, line 740: Why this else? This seems like a functional difference

lindexi: @twsouthwick  Thank you. The first time the code enters this method, the property called "NamespaceDeclField" is null. And the "foreach" will do nothing. That means we should ignore the "foreach" code when the NamespaceDeclField is null.

lindexi:
@twsouthwick  I added the benchmark code.

PR #793, opened 2020-08-17
Add a relationship rewriter to help sanitize malformed URIs

### LINKED ISSUE #715: One illegal uri in Relationship will destroy the document parsing
**Description**

The PPTX document that include an illegal uri in Relationship will make the `System.IO.Packaging.InternalRelationshipCollection.ProcessRelationshipAttributes` throw an exception to  `OpenXmlPart.Load`.

And the `OpenXmlPart.Load` can not catch the exception and it will break the `PresentationDocument.Open`.

**Information**

- .NET Target: All
- DocumentFormat.OpenXml Version: 2.10.1

**Repro**

```csharp
var document = PresentationDocument.Open("hyperlink.pptx", isEditable: false, openSettings)
``` 

Here is the hyperlink.pptx file : https://1drv.ms/p/s!AiKjiQqRWKThlv5zkY4HoRvvJ3Ppdg?e=3kfdNU 

**Observed**

The PresentationDocument.Open throw the UriFormatException exception

```
System.UriFormatException: 'Invalid URI: The hostname could not be parsed.'
```

Because the ppt\slides\_rels\slide1.xml.rels contain this string

```xml
<Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/hyperlink" Target="mailto:!@#$%^&amp;*()_+}{:”?&gt;&lt;,./;’[]=-098766554321" TargetMode="External"/>
```

As you can see, the `Target` is not an uri.

**Expected**

We can design an exception handle API, and we can handle some illegal document.

See #38 #274 #297 #298 

And the #298 only add more information but can not tolerate errors.

And just as @twsouthwick says, we can not fix this in the OpenXML SDK project https://github.com/OfficeDev/Open-XML-SDK/issues/297#issuecomment-345560540 , but I think we can tolerate some errors

twsouthwick:
Since .NET 4.5, there has been issues using the SDK with relationships that have malformed URIs. Specifically, this seems to be seen when docs have hyperlinks that are not valid URIs. This change adds a handler to rewrite this upon opening a document in case they exist. Subsequent user code can handle any other manipulation that is needed (such as removal of the offending hyperlink).

Fixes #715

### DIFF
--- a/src/DocumentFormat.OpenXml/Packaging/OpenSettings.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/OpenSettings.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+
 namespace DocumentFormat.OpenXml.Packaging
 {
     /// <summary>
@@ -24,6 +26,7 @@ namespace DocumentFormat.OpenXml.Packaging
             MarkupCompatibilityProcessSettings.ProcessMode = other.MarkupCompatibilityProcessSettings.ProcessMode;
             MarkupCompatibilityProcessSettings.TargetFileFormatVersions = other.MarkupCompatibilityProcessSettings.TargetFileFormatVersions;
             MaxCharactersInPart = other.MaxCharactersInPart;
+            RelationshipErrorHandlerFactory = other.RelationshipErrorHandlerFactory;
         }
 
         /// <summary>
@@ -64,5 +67,10 @@ namespace DocumentFormat.OpenXml.Packaging
         /// This property allows you to mitigate denial of service attacks where the attacker submits a package with an extremely large Open XML part. By limiting the size of the part, you can detect the attack and recover reliably.
         /// </remarks>
         public long MaxCharactersInPart { get; set; }
+
+        /// <summary>
+        /// Gets or sets a delegate that is used to create a handler to rewrite relationships that are malformed. On platforms after .NET 4.5, <see cref="Uri"/> parsing will fail on malformed strings.
+        /// </summary>
+        public Func<OpenXmlPackage, RelationshipErrorHandler> RelationshipErrorHandlerFactory { get; set; }
     }
 }
--- a/src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
@@ -1,0 +1,137 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.IO.Packaging;
+using System.Xml.Linq;
+
+namespace DocumentFormat.OpenXml.Packaging
+{
+    /// <summary>
+    /// A class that holds information about invalid relationships in <see cref="PackagePart"/> instances.
+    /// </summary>
+    public abstract class RelationshipErrorHandler
+    {
+        private const string RelationshipsTagName = "Relationships";
+        private const string RelationshipTagName = "Relationship";
+        private const string RelationshipNamespaceUri = "http://schemas.openxmlformats.org/package/2006/relationships";
+        private const string TargetAttributeName = "Target";
+        private const string IdAttributeName = "Id";
+
+        /// <summary>
+        /// Create a <see cref="RelationshipErrorHandler"/> that simply rewrites the invalid target Uri.
+        /// </summary>
+        /// <param name="rewriter">The delegate used to rewrite the Uri.</param>
+        /// <returns>A factory function for use in <see cref="OpenSettings.RelationshipErrorHandlerFactory"/>.</returns>
+        public static Func<OpenXmlPackage, RelationshipErrorHandler> CreateRewriterFactory(Rewriter rewriter)
+        {
+            var handler = new DelegateRelationshipErrorHandler(rewriter);
+
+            return _ => handler;
+        }
+
+        /// <summary>
+        /// Delegate to be used for simple rewriting of malformed Uris.
+        /// </summary>
+        /// <param name="partUri">Uri of the part with the invalid relationship.</param>
+        /// <param name="id">Id of relationship</param>
+        /// <param name="uri">Invalid <see cref="Uri"/></param>
+        /// <returns>Rewritten string if available, otherwise <c>null</c>.</returns>
+        public delegate string Rewriter(Uri partUri, string id, string uri);
+
+        /// <summary>
+        /// Rewrites an invalid URI with a valid one in order to correctly open a package.
+        /// </summary>
+        /// <param name="partUri">Uri of the part with the invalid relationship.</param>
+        /// <param name="id">Id of relationship</param>
+        /// <param name="uri">Invalid <see cref="Uri"/></param>
+        /// <returns>Rewritten string if available, otherwise <c>null</c>.</returns>
+        public abstract string Rewrite(Uri partUri, string id, string uri);
+
+        /// <summary>
+        /// Callback for after a package has been completely loaded and all rewritting has occurred.
+        /// </summary>
+        public virtual void OnPackageLoaded()
+        {
+        }
+
+        internal void Handle(Package package)
+        {
+            if (package is null)
+            {
+                throw new ArgumentNullException(nameof(package));
+            }
+
+            if ((package.FileOpenAccess & FileAccess.ReadWrite) != FileAccess.ReadWrite)
+            {
+                throw new OpenXmlPackageException(ExceptionMessages.PackageAccessModeIsReadonly);
+            }
+
+            foreach (var part in package.GetParts())
+            {
+                if (part.Uri.OriginalString.EndsWith(".rels", StringComparison.OrdinalIgnoreCase))
+                {
+                    var doc = WalkRelationships(part);
+
+                    if (doc != null)
+                    {
+                        using var stream = part.GetStream(FileMode.Open, FileAccess.Write);
+
+                        doc.Save(stream);
+                    }
+                }
+            }
+        }
+
+        private XDocument WalkRelationships(PackagePart part)
+        {
+            using var stream = part.GetStream(FileMode.Open, FileAccess.Read);
+            using var reader = new StreamReader(stream);
+
+            var doc = XDocument.Load(reader);
+            var changed = false;
+
+            if (string.Equals(RelationshipsTagName, doc.Root.Name.LocalName, StringComparison.Ordinal)
+                && string.Equals(RelationshipNamespaceUri, doc.Root.Name.Namespace.NamespaceName, StringComparison.Ordinal))
+            {
+                foreach (var child in doc.Root.Descendants(XName.Get(RelationshipTagName, RelationshipNamespaceUri)))
+                {
+                    var id = child.Attribute(IdAttributeName)?.Value;
+                    var target = child.Attribute(TargetAttributeName)?.Value;
+
+                    if (!Uri.TryCreate(target, UriHelper.RelativeOrAbsolute, out _))
+                    {
+                        var updated = Rewrite(part.Uri, id, target);
+
+                        if (updated != null)
+                        {
+                            if (!Uri.TryCreate(updated, UriHelper.RelativeOrAbsolute, out _))
+                            {
+                                throw new InvalidOperationException(ExceptionMessages.InvalidUriProvided);
+                            }
+
+                            child.SetAttributeValue(TargetAttributeName, updated);
+
+                            changed = true;
+                        }
+                    }
+                }
+            }
+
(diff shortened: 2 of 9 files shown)


twsouthwick:
Use of this is shown in the test:

```csharp
[Fact]
public void RewriteMalformedUri()
{
    const string Rewritten = "http://error";
    const string Id = "rId1";

    using (var stream = GetStream(TestFiles.malformed_uri_xlsx, isEditable: true))
    {
        var rewritten = 0;
        var settings = new OpenSettings
        {
            RelationshipErrorRewriter = error =>
            {
                rewritten++;
                Assert.Equal(Id, error.Id);
                Assert.Equal("/xl/worksheets/_rels/sheet1.xml.rels", error.PartUri.OriginalString);
                Assert.Equal("mailto:one@", error.Uri);

                return Rewritten;
            },
        };

        using (var doc = SpreadsheetDocument.Open(stream, true, settings))
        {
            Assert.Equal(1, rewritten);

            var worksheet = Assert.Single(doc.WorkbookPart.WorksheetParts);
            var hyperlink = Assert.Single(worksheet.HyperlinkRelationships);

            Assert.Equal(Id, hyperlink.Id);
            Assert.Equal(Rewritten, hyperlink.Uri.OriginalString);
        }
    }
}
```

twsouthwick:
An alternate API I just pushed up is to use the factory pattern. This allows state to be gathered in a simpler way and ensures anything captured is not retained (since OpenSettings actually holds onto everything):

Here's an example of using it where it automatically removes any offending hyperlinks:

```csharp
[Fact]
public void RemoveHyperlinks()
{
    using (var stream = GetStream(TestFiles.malformed_uri_xlsx, isEditable: true))
    {
        var settings = new OpenSettings
        {
            RelationshipErrorHandlerFactory = p => new RemoveMalformedHyperlinksRelationshipErrorHandler(p),
        };

        using (var doc = SpreadsheetDocument.Open(stream, true, settings))
        {
            var worksheetPart = Assert.Single(doc.WorkbookPart.WorksheetParts);
            Assert.Empty(worksheetPart.HyperlinkRelationships);

            var worksheet = Assert.IsType<Worksheet>(worksheetPart.RootElement);
            Assert.Empty(worksheet.Descendants<Hyperlink>());

            var validator = new OpenXmlValidator();
            Assert.Empty(validator.Validate(doc));
        }
    }
}

private class RemoveMalformedHyperlinksRelationshipErrorHandler : RelationshipErrorHandler
{
    private readonly OpenXmlPackage _package;
    private readonly Dictionary<string, List<string>> _errors;

    public RemoveMalformedHyperlinksRelationshipErrorHandler(OpenXmlPackage package)
    {
        _package = package;
        _errors = new Dictionary<string, List<string>>(StringComparer.Ordinal);
    }

    public override string Rewrite(Uri partUri, string id, string uri)
    {
        var key = partUri.OriginalString
            .Replace("_rels/", string.Empty)
            .Replace(".rels", string.Empty);

        if (!_errors.ContainsKey(key))
        {
            _errors.Add(key, new List<string>());
        }

        _errors[key].Add(id);

        return "http://error";
    }

    public override void OnPackageLoaded()
    {
        foreach (var part in _package.GetAllParts())
        {
            if (_errors.TryGetValue(part.Uri.OriginalString, out var ids))
            {
                foreach (var id in ids)
                {
                    part.DeleteReferenceRelationship(id);

                    if (part is WorksheetPart && part.RootElement is Worksheet ws)
                    {
                        foreach (var h in ws.Descendants<Hyperlink>())
                        {
                            var parent = h.Parent;

                            if (h.Id == id)
                            {
                                h.Remove();
                            }

                            if (!parent.HasChildren)
                            {
                                parent.Remove();
                            }
                        }
                    }
                }
            }
        }
    }
}
```

src/DocumentFormat.OpenXml/Packaging/OpenSettings.cs
```diff
@@ -61,5 +67,10 @@ public MarkupCompatibilityProcessSettings MarkupCompatibilityProcessSettings
         /// This property allows you to mitigate denial of service attacks where the attacker submits a package with an extremely large Open XML part. By limiting the size of the part, you can detect the attack and recover reliably.
         /// </remarks>
         public long MaxCharactersInPart { get; set; }
+
+        /// <summary>
+        /// Gets or sets a delegate that is used to create a handler to rewrite relationships that are malformed. On platforms after .NET 4.5, <see cref="Uri"/> parsing will fail on malformed strings.
+        /// </summary>
+        public Func<OpenXmlPackage, RelationshipErrorHandler> RelationshipErrorHandlerFactory { get; set; }
```
lindexi, line 74: I think using the interface will be better than using Func

src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.IO;
+using System.IO.Packaging;
+using System.Xml.Linq;
+
+namespace DocumentFormat.OpenXml.Packaging
+{
+    /// <summary>
+    /// A class that holds information about invalid relationships in <see cref="PackagePart"/> instances.
+    /// </summary>
+    public abstract class RelationshipErrorHandler
```
lindexi, line 14: I think there should be at least one default implementation in the repository. This means that there is no need for each developer to implement a specific type by himself

src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                if (part.Uri.OriginalString.EndsWith(".rels", StringComparison.OrdinalIgnoreCase))
+                {
+                    var doc = WalkRelationships(part);
+
+                    if (doc != null)
+                    {
+                        using var stream = part.GetStream(FileMode.Open, FileAccess.Write);
+
+                        doc.Save(stream);
```
lindexi, line 59: I am puzzled about a question: If we open a document in read-only mode, will this code change the document?

lindexi:
@twsouthwick Thank you. But sorry, I do not think it is a good idea. On the one hand, this code adds a traversal, which will reduce performance. On the other hand, this code seems to modify the original document.

twsouthwick:
Yes you are right on both accounts, but it does provide a workaround that we can make available ASAP. Can you describe a scenario that our won't work in? Are there times when the relationship needs to remain but is still an invalid uri?

src/DocumentFormat.OpenXml/Packaging/OpenSettings.cs
twsouthwick: Can you explain a bit more what you mean by this? This simply a factory to get a handler. How can an interface help here?

src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
twsouthwick: Maybe. What would you suggest as the default implementation?

src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
twsouthwick: No, it will not. It will throw. If you want it to remain read-only, you'd need to operate on a copy.

twsouthwick:
As for performance, you'd only incur that if you supply a handler; otherwise there's no change. However, performance aside, this provides a simple way to handle an otherwise difficult thing to handle.

src/DocumentFormat.OpenXml/Packaging/OpenSettings.cs
lindexi: @twsouthwick   Thank you. This is just the habitual use of factory design patterns.

An optimization point is as the following code:

       openSettings.RelationshipErrorHandlerFactory  += Foo1;
       openSettings.RelationshipErrorHandlerFactory  += Foo2;

src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
lindexi: @twsouthwick Can we use the `RemoveMalformedHyperlinksRelationshipErrorHandler` code?

src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
lindexi: @twsouthwick If we follow this idea, does our arguments lack the necessary information to support the method of copying file copies? For example, I want to copy files in this code, but I don’t know which file is currently being used.

lindexi:
> 
> 
> As for performance, you'd only incur that if you supply a handler; otherwise there's no change. However, performance aside, this provides a simple way to handle an otherwise difficult thing to handle.

I quite agree with you. After System.IO.Packaging.Package is changed, will we still consider changing this part of the code?

twsouthwick:
I highly doubt we'll get a change into System.IO.Packaging.Package. I'll continue to push for it, but this is a way to mitigate that issue so that people don't get blocked.

src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
twsouthwick: I'd be happy to have an implementation like that live in the project, but it is definitely only test code at the moment. We would want a way to know we are removing all hyperlinks - it is special cased for spreadsheets.

src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
twsouthwick: Can you give an example of what you mean? I'm not sure I follow. From what I understand, you'd do this

```csharp
using var fs = File.OpenRead(....);
using var ms = new MemoryStream();

fs.CopyTo(ms);

using var doc = WordDocument.Open(ms);
```

src/DocumentFormat.OpenXml/Packaging/OpenSettings.cs
twsouthwick: Not sure why this would be an optimization. I'm not personally a fan of events as they are difficult to compose (you kind of can, though, via System.Reactive with some weird hoops initially). If needed, you could create a `RelationshipErrorHandler` that could compose multiple handlers together that hand off to the next one if needed.

src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
lindexi: @twsouthwick Thank you. This method seems to solve the problem. But does this code take up too much memory?

twsouthwick: Potentially, at that point, you'd want to offload it to a file. Depends on the file

src/DocumentFormat.OpenXml/Packaging/RelationshipErrorHandler.cs
twsouthwick: I've got a default implementation that just takes a delegate to rewrite. See `RelationshipErrorHandler.CreateRewriterFactory`

lindexi (approved):

src/DocumentFormat.OpenXml/System/Xml/Linq/XDocumentExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+
+namespace System.Xml.Linq
+{
+    internal static class XDocumentExtensions
+    {
+        public static void Save(this XDocument doc, Stream stream)
```
lindexi, line 10: Should we add the `SaveAsync` by XDocument.SaveAsync ?

twsouthwick: No, this is a polyfill for .NET 3.5. Nothing in the SDK is async. I've wondered if there are scenarios for that; it would be a pretty big undertaking. If you have scenarios where you'd see improvement with async, I'm happy to discuss it on a different issue.

lindexi:
@twsouthwick Any update?

lindexi:
@twsouthwick  Thank you. This will fix the problem. And I wrote a Chinese blog to promote it, see [dotnet OpenXML 修复 Office 文档里面包含格式不正确的 Uri 而无法解析](https://blog.lindexi.com/post/dotnet-OpenXML-%E4%BF%AE%E5%A4%8D-Office-%E6%96%87%E6%A1%A3%E9%87%8C%E9%9D%A2%E5%8C%85%E5%90%AB%E6%A0%BC%E5%BC%8F%E4%B8%8D%E6%AD%A3%E7%A1%AE%E7%9A%84-Uri-%E8%80%8C%E6%97%A0%E6%B3%95%E8%A7%A3%E6%9E%90.html )

PR #804, opened 2020-09-09
Provide IXmlLineInfo from OpenXmlReader when available

### LINKED ISSUE #803: Reader depth or how to get current node's real depth.
Is there any approach to get `depth`, `lineNumber` or `position` of current node in a real XML file?
As it's done using native `XmlTextReader`: https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmltextreader.lineposition?view=netcore-3.1
```csharp
while (reader.Read()){
       switch (reader.NodeType){
         case XmlNodeType.Element:
           Console.Write("{0} {1},{2}  ", reader.Depth, reader.LineNumber, reader.LinePosition);
           Console.WriteLine("<{0}>", reader.Name);
           break;
       }
    }
```

I have tried to get `Depth` of OpenXMLReader, but it seems that it is returning depth of element in some stack which is not real depth of current xml node.

It would be nice to have such functionality.

twsouthwick:
Provides ability to track the line number and position while reading a part. This is only available on implementations that operate on text itself; implementations such as the `OpenXmlDomReader` operate on nodes which don't have line information anymore.

Fixes #803


amanbolat:
Amazing! Thanks you for that pull request. I hope other people will appreciate this functionality too!

PR #809, opened 2020-09-17
Parse exponent for int/uint if present

### LINKED ISSUE #806: System.FormatException on Row.RowIndex.Value once row index hits one million
**Description**

I need to import data from an exceptionally large Excel worksheet (~ 1 million rows),
and I'm using the `OpenXmlReader` class to read the sheet row by row.
However, when calling the `Row.RowIndex.Value` method for the row with index 1 million,
I'm getting a `System.FormatException`.

**Information**

- .NET Target:  .NET Core 3.1
- DocumentFormat.OpenXml Version: 2.11.3
- Partial stack trace:

_System.FormatException: Input string was not in a correct format.
   at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
   at System.Number.ParseUInt32(ReadOnlySpan\`1 value, NumberStyles styles, NumberFormatInfo info)
   at DocumentFormat.OpenXml.UInt32Value.Parse(String input)
   at DocumentFormat.OpenXml.OpenXmlSimpleValue`1.get_Value()_

**Repro**

```csharp
using var reader = OpenXmlReader.Create(worksheetPart);
while (reader.Read())
{
    if (reader.ElementType == typeof(Row))
    {
        var row = (Row)reader.LoadCurrentElement();
        uint rowIndex = row.RowIndex.Value;
    }
}
``` 

```xml
<row r="999999" s="1" customFormat="1" ht="19.7321" customHeight="1">
    <c r="A999999" s="3" t="s">
        <v>3231031</v>
    </c>
    <c r="B999999" s="3" t="s">
        <v>18938</v>
    </c>
</row>
<row r="1e6" s="1" customFormat="1" ht="19.7321" customHeight="1">
    <c r="A1000000" s="3" t="s">
        <v>3231032</v>
    </c>
    <c r="B1000000" s="3" t="s">
        <v>440271</v>
    </c>
</row>
<row r="1.000001e6" s="1" customFormat="1" ht="19.7321" customHeight="1">
    <c r="A1000001" s="3" t="s">
        <v>3231033</v>
    </c>
    <c r="B1000001" s="3" t="s">
        <v>3231034</v>
    </c>
</row>
```

**Observed**

The value that's causing the exception is **1e6**.
What's interesting is that once the row index reaches 1,000,000, the number format of the indexes changes from base 10 to exponential notation.
What's even more curious is that this only happens for rows, because—as you can notice from the example—the values of the cell references are still base 10, even for values larger than 1,000,000.

**Expected**

I have absolutely no idea why the row indexes change format at the 1,000,000 mark, but I would expect the SDK to be able to work with them, nevertheless.

twsouthwick:
Fixes #806

### DIFF
--- a/src/DocumentFormat.OpenXml/SimpleTypes/Int32Value.cs
+++ b/src/DocumentFormat.OpenXml/SimpleTypes/Int32Value.cs
@@ -14,7 +14,7 @@ namespace DocumentFormat.OpenXml
     [DebuggerDisplay("{InnerText}")]
     public class Int32Value : OpenXmlComparableSimpleValue<int>
     {
-        private static readonly NumberStyles ParseStyles = NumberStyles.AllowLeadingSign | NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite | NumberStyles.AllowExponent | NumberStyles.AllowDecimalPoint;
+        private const NumberStyles ParseStyles = NumberStyles.AllowLeadingSign | NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite | NumberStyles.AllowExponent | NumberStyles.AllowDecimalPoint;
 
         /// <summary>
         /// Initializes a new instance of the Int32Value class.
--- a/src/DocumentFormat.OpenXml/SimpleTypes/Int32Value.cs
+++ b/src/DocumentFormat.OpenXml/SimpleTypes/Int32Value.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Globalization;
 using System.Xml;
 
 namespace DocumentFormat.OpenXml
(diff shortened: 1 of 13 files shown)


src/DocumentFormat.OpenXml/SimpleTypes/Int32Value.cs
```diff
@@ -13,6 +14,8 @@ namespace DocumentFormat.OpenXml
     [DebuggerDisplay("{InnerText}")]
     public class Int32Value : OpenXmlComparableSimpleValue<int>
     {
+        private static readonly NumberStyles ParseStyles = NumberStyles.AllowLeadingSign | NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite | NumberStyles.AllowExponent;
```
abalaci, line 17: Hey @twsouthwick, many thanks for taking the time to look into this!

Just a quick point based on what I noticed trying to work around the issue myself—the number styles you have here won't be able to parse something like 1.000001e6 into 1,000,001. Here's a short fiddle: https://dotnetfiddle.net/EXr0Vw.

To cover that, I think you'll probably have to include `NumberSyles.AllowDecimalPoint` as well, although I'm not sure whether this would have any broader implications in terms of parsing.

twsouthwick: Oh interesting. Thanks for catching that. Do you see that in the row count?

abalaci: Yes, I do - see the XML sample I added to the issue. It's actually from the very same file that was causing problems.

twsouthwick: Cool. I've pushed an update for that case.

twsouthwick:
It doesn't look like Office supports opening these files. Closing for now. If desired, open the issue for further discussion.

PR #818, opened 2020-11-19
Fixed NullReferenceException in OpenXmlElement RawOuterXml field

aidilumarov:
The current version of your code:

It sets _rawOuterXml to string.Empty if value is null, but instead of leaving the setter immediately, it resets the field to null at the end of the setter

The changes I made should resolve the issue.

Thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Version 2.12.1 - 2020-12-19
+
+### Fixed
+- Fixed a `System.NullReferenceException` when setting _rawOuterXml field in `DocumentFormat.OpenXml.OpenXml` (#818)
+- Added regression tests in `DocumentFormat.OpenXml.Tests.OpenXmlElementTest2` to check for the mentioned exceptions in the future (#818)
+
 ## Version 2.12.0 - 2020-12-09
 
 ### Added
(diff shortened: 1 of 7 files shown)


twsouthwick:
@aidilumarov Thanks for the catch! Can you add a regression test so that it will be caught in the future?

twsouthwick:
@aidilumarov I'd like to get a 2.12.0 release out this week. If you can get some tests in, we'll have this as part of that release.

Also, can you update the CHANGELOG.md to include this?

aidilumarov:
@twsouthwick Sorry for taking long. I was off for several weeks. I have made the changes you asked for. Kind regards.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Version 2.12.1 - 2020-12-19
+
+### Fixed
+- Fixed a `System.NullReferenceException` when setting _rawOuterXml field in `DocumentFormat.OpenXml.OpenXml` (#818)
+- Added regression tests in `DocumentFormat.OpenXml.Tests.OpenXmlElementTest2` to check for the mentioned exceptions in the future (#818)
```
twsouthwick, line 11: This line is unneeded. I try to keep the changelog focused on observable changes to the library

src/DocumentFormat.OpenXml/Properties/launchSettings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "profiles": {
+    "DocumentFormat.OpenXml": {
```
twsouthwick, line 3: This file and the other one appear unrelated to the change. I believe this is for vs code. If so, happy to take it but would like the rest of the projects supported as well. Best in a separate pr though

twsouthwick (changes requested):
Looks like the build is broken. We have warnings as errors on CI builds

twsouthwick:
I've taken the liberty of updating the branch so it can be merged

twsouthwick (approved):

PR #856, opened 2021-01-13
Replace usage of namespace ids with OpenXmlNamespace

twsouthwick:
This change is fairly large and reworks how namespaces are handled internally. The main benefit of this change is it removes reliance on all inner workings on the namespace id. This was a hard coded value that was essentially a magic number for each namespace. This replaces that with a struct that handles the comparisons (`OpenXmlNamespace`) as well as a type that wraps that to identify a qualified name (`OpenXmlQualifiedName`). This change also makes it clear where a string is a name, a namespace, a prefix, etc.

This does not update the generated files to not use namespace ids. That is still in place, but is the last place where a namespace id is needed.


tomjebo (commented):
what is this?

tomjebo (approved):

PR #857, opened 2021-01-19
Openxmlreader ignorews option

lklein53:
Changes discussed in #848 @twsouthwick

PR #869, opened 2021-01-27
Improve whitespace handling

lklein53:
Added more test cases for the ignoreWhitespace=false option and fixes to handle these cases.

@twsouthwick don't know what you prefer. If you want to remove the changes done in https://github.com/OfficeDev/Open-XML-SDK/pull/857 or if you would like to fix the introduced issues. I wasn't aware that changing the xml reader settings will break the internal state handling of the OpenXMLReader

### DIFF
--- a/src/DocumentFormat.OpenXml/OpenXmlPartReader.cs
+++ b/src/DocumentFormat.OpenXml/OpenXmlPartReader.cs
@@ -451,6 +451,12 @@ namespace DocumentFormat.OpenXml
                 case ElementState.MiscNode:
                     // cursor is end element, pop stack
                     _elementStack.Pop();
+                    if (_elementStack.Count == 0)
+                    {
+                        _elementState = ElementState.EOF;
+                        return false;
+                    }
+
                     break;
 
                 case ElementState.LeafStart:
--- a/test/DocumentFormat.OpenXml.Tests/ofapiTest/OpenXmlReaderTest.cs
+++ b/test/DocumentFormat.OpenXml.Tests/ofapiTest/OpenXmlReaderTest.cs
@@ -426,6 +426,66 @@ namespace DocumentFormat.OpenXml.Tests
         }
 
         /// <summary>
+        /// Test that the OpenXmlReader can handle formatted xml.
+        ///</summary>
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void TestIgnoreWhitespaceWithFormattedXml(bool ignoreWhitespace) {
+            const string PartText = @"
+    <w:document xmlns:v=""urn:schemas-microsoft-com:vml"" xmlns:w=""http://schemas.openxmlformats.org/wordprocessingml/2006/main"">
+      <w:body>
+        <w:p>
+          <w:r>
+            <w:t>First Text</w:t>
+          </w:r>
+          <w:r>
+            <w:t>Second Text</w:t>
+          </w:r>
+        </w:p>
+      </w:body>
+    </w:document>";
+
+            UTF8Encoding utf8Encoding = new UTF8Encoding();
+            var exception = Record.Exception(() => {
+                using var stream = new MemoryStream(utf8Encoding.GetBytes(PartText), false);
+                using var reader = OpenXmlReader.Create(stream, false, ignoreWhitespace);
+                while (reader.Read()) {
+                }
+
+                reader.Close();
+            });
+            Assert.Null(exception);
+        }
+
+        /// <summary>
+        /// Test that the OpenXmlReader can handle a whitespace after the last element.
+        ///</summary>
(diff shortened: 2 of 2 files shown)


twsouthwick:
Is this a new issue uncovered by the change or introduced by the change?

src/DocumentFormat.OpenXml/OpenXmlPartReader.cs
```diff
@@ -451,6 +451,12 @@ private bool MoveToNextElement()
                 case ElementState.MiscNode:
                     // cursor is end element, pop stack
                     _elementStack.Pop();
+                    if (_elementStack.Count == 0)
```
twsouthwick, line 454: This looks like it was a bug that just wasn't caught before :)

test/DocumentFormat.OpenXml.Tests/ofapiTest/OpenXmlReaderTest.cs
```diff
@@ -425,6 +425,66 @@ public void PartReaderIgnoreWhitespaceTest(bool ignoreWhitespace)
...
+        public void TestIgnoreWhitespaceWhitespaceAfterLastElement(bool ignoreWhitespace) {
+            const string PartText = @"
+    <w:document xmlns:v=""urn:schemas-microsoft-com:vml"" xmlns:w=""http://schemas.openxmlformats.org/wordprocessingml/2006/main"">
+      <w:body>
+        <w:p>
+        </w:p>
+      </w:body>
+    </w:document> ";
+
+            UTF8Encoding utf8Encoding = new UTF8Encoding();
```
twsouthwick, line 476: You can just use `Encoding.UTF8` to simplify it a little

test/DocumentFormat.OpenXml.Tests/ofapiTest/OpenXmlReaderTest.cs
```diff
@@ -425,6 +425,66 @@ public void PartReaderIgnoreWhitespaceTest(bool ignoreWhitespace)
...
+            var exception = Record.Exception(() => {
+                using var stream = new MemoryStream(utf8Encoding.GetBytes(PartText), false);
+                using var reader = OpenXmlReader.Create(stream, false, ignoreWhitespace);
+                while (reader.Read()) {
+                }
+
+                reader.Close();
+            });
+            Assert.Null(exception);
```
twsouthwick, line 458: Instead of asserting null, I'd rather see checks that the reader is output things we'd expect. ie after each `Assert.True(reader.Read())` add an assert for the current state

twsouthwick (changes requested):
Thanks for following up with this. A few things to fix up.

lklein53:
> Is this a new issue uncovered by the change or introduced by the change?

I think it is an issue intorduced by the change. With the IgnoreWhitespace set to false new nodes of type Whitespace are returned by the underlying xml reader which wasn't the case before

src/DocumentFormat.OpenXml/OpenXmlPartReader.cs
lklein53: Before the whitespace change this was handeled by 
```csharp
if (_xmlReader.EOF || !_xmlReader.Read())
{
  _elementState = ElementState.EOF;
  return false;
}
```
further down in the function. With the whitespace change the stack can be empty but the reader is not at the end of file as it still hat whitespace to read

twsouthwick:
That makes sense, but was there any way to expose it before? Just trying to understand if it's only an issue that exists now due to the option to ignore whitespace.

lklein53:
I am not sure.

> That makes sense, but was there any way to expose it before? Just trying to understand if it's only an issue that exists now due to the option to ignore whitespace.

I think the following example would have caused the same error without the whitespace change:
```xml
 <w:document xmlns:v=""urn:schemas-microsoft-com:vml"" xmlns:w=""http://schemas.openxmlformats.org/wordprocessingml/2006/main"">
      <w:body>
        <w:p>
        </w:p>
      </w:body>
    </w:document> <!--Comment-->";
```

test/DocumentFormat.OpenXml.Tests/ofapiTest/OpenXmlReaderTest.cs
lklein53: Done

test/DocumentFormat.OpenXml.Tests/ofapiTest/OpenXmlReaderTest.cs
lklein53: Updated the test cases and added one test case that does not depend on the whitespace setting

twsouthwick (approved):
LGTM! Thanks for following up with this!

PR #882, opened 2021-02-10
Office 2019 Updates

tomjebo:
update the generated code from the schema processor (backend)
add code to include new namespaces


src/DocumentFormat.OpenXml/GeneratedCode/schemas_microsoft_com_office_drawing_2016_11_diagram.g.cs
```diff
@@ -16,7 +16,7 @@ namespace DocumentFormat.OpenXml.Office2019.Drawing.Diagram11
 {
     /// <summary>
     /// <para>Defines the NumberDiagramInfoList Class.</para>
-    /// <para>This class is available in Office 2007 or above.</para>
+    /// <para>This class is available in Office 2019 or above.</para>
```
AlfredHellstern, line 19: in the interop documentation we normally use "is supported in 20xx and later" or use "is not supported in 20xx or later". So if supported, use "and later", if not supported, use "or later".

tomjebo: Seems like the meaning is the same but I'll defer to others' opinions. I don't see a really strong need except to match Open Specs docs exact wording. Am I missing anything else?

AlfredHellstern: I saw that you used the usual wording in other files, so would suggest to make this the same.

src/DocumentFormat.OpenXml/GeneratedCode/schemas_microsoft_com_office_spreadsheetml_2017_richdata2.g.cs
```diff
@@ -188,7 +188,7 @@ public StringValue RichSortKey
         }
 
         /// <summary>
-        /// <para>descending</para>
+        /// <para>descending, this property is only available in Office 2010 and later.</para>
```
AndrewDavidoff, line 191: File name suggests that this is schema registered in 2017. Is it correct that property can be available in Office 2010? (multiple cases)

src/DocumentFormat.OpenXml/GeneratedCode/schemas_openxmlformats_org_drawingml_2006_main.g.cs
```diff
@@ -12758,7 +12758,7 @@ public StringValue Name
         }
 
         /// <summary>
-        /// <para>id</para>
+        /// <para>id, this property is only available in Office 2013 and later.</para>
```
AndrewDavidoff, line 12761: File name indicates schema from year 2006. Is it correct that this parameter is "only available in Office 2013 and later"?

src/DocumentFormat.OpenXml/GeneratedCode/schemas_openxmlformats_org_wordprocessingml_2006_main.g.cs
```diff
@@ -32168,7 +32168,7 @@ public PrinterSettingsReference PrinterSettingsReference
         }
 
         /// <summary>
-        /// <para>FootnoteColumns.</para>
+        /// <para>FootnoteColumns, this property is only available in Office 2013 and later..</para>
```
AndrewDavidoff, line 32171: File name indicates schema from year 2006. Is it correct that this parameter is available only starting in Office 2013?

src/DocumentFormat.OpenXml/GeneratedCode/schemas_openxmlformats_org_wordprocessingml_2006_main.g.cs
```diff
@@ -66469,23 +66469,23 @@ public enum CompatSettingNameValues
         ///<summary>
         ///useWord2013TrackBottomHyphenation.
         ///<para>When the item is serialized out as xml, its value is "useWord2013TrackBottomHyphenation".</para>
-        ///<para>This item is only available in Office2016.</para>
+        ///<para>This item is only available in Office 2016 and later.</para>
```
AndrewDavidoff, line 66472: File name indicates schema from year 2006. Is it correct that this parameter is available only starting in Office 2016? (two more cases in this file)

src/DocumentFormat.OpenXml/GeneratedCode/schemas_microsoft_com_office_drawing_2016_11_diagram.g.cs
twsouthwick: This is auto generated and something I actually just put in. We can definitely change it to match standard wordage

twsouthwick (approved):
LGTM

src/DocumentFormat.OpenXml/GeneratedCode/schemas_microsoft_com_office_spreadsheetml_2017_richdata2.g.cs
tomjebo: the reason for this is that the containing complex type is defined in the Office 2019 schema http://schemas.microsoft.com/office/spreadsheetml/2017/richdata2 but the property this comment is describing and which is used by that containing type is actually one that has been around since Office 2010 in the namespace http://schemas.microsoft.com/office/spreadsheetml/2009/9/main.

src/DocumentFormat.OpenXml/GeneratedCode/schemas_openxmlformats_org_wordprocessingml_2006_main.g.cs
tomjebo: This one is a little different. So the ST_CompatSettingName is an enumeration that is defined in most versions of Word format but this enum value was added later in the same xsd file as all previous versions. So it's correct that it is for a later version of Office.

src/DocumentFormat.OpenXml/GeneratedCode/schemas_openxmlformats_org_drawingml_2006_main.g.cs
tomjebo: Same issue here; thm15:id property added to the existing a:theme (2010 or earlier) root element. Hence updating in the 2006  namespace-named generated code.

src/DocumentFormat.OpenXml/GeneratedCode/schemas_openxmlformats_org_wordprocessingml_2006_main.g.cs
tomjebo: Same issue here; w15:footnoteColumns was added directly to the w:secPr (previous section properties) and this caused it to be added to the existing 2006 named generated file.

AndrewDavidoff (approved):

src/DocumentFormat.OpenXml/GeneratedCode/schemas_microsoft_com_office_drawing_2016_11_diagram.g.cs
tomjebo: we agreed to make this change in master first and then merge it to the Office2019 project.

PR #890, opened 2021-02-21
Allow 0/1 for boolean cell values

### LINKED ISSUE #887: OpenXmlValidator not validating boolean values

Is this a:
- [X] Issue with the OpenXml library

If you have answered that this is a question, please ask it on StackOverflow instead of here.
This issue tracker is meant to track product issues while StackOverflow excels at answering questions

---------------- Remove this line and above before posting ----------------

**Description**

This looks like a regression. Values of type boolean, but with values `0` or `1` in the OpenXML document fail to validate.

**Information**

- .NET Target: .NET Core 3.1
- DocumentFormat.OpenXml Version: 2.12.0

This problem doesn't exist in 2.11.3, hence it looks like a regression between 2.11.3 and 2.12.0.

**Repro**

```csharp
var version = typeof(SpreadsheetDocument).Assembly.GetName().Version.ToString();
Console.WriteLine(version);
using var package = SpreadsheetDocument.Open("test.xlsx", isEditable: false);
var validator = new OpenXmlValidator();
var errors = validator.Validate(package);

if (errors.Any())
{
    Console.WriteLine("This file should validate.");
    Console.WriteLine();
    foreach (var e in errors)
    {
        Console.WriteLine($"Part {e.Part.Uri}, Path {e.Path.XPath}: {e.Description}");
    }
}
else
    Console.WriteLine("No validation errors");
``` 
Test file: [test.xlsx](https://github.com/OfficeDev/Open-XML-SDK/files/6017463/test.xlsx)

**Observed**

`OpenXmlValidator.Validate()` produces errors.

> 2.12.0.0
> This file should validate.
> 
> Part /xl/worksheets/sheet1.xml, Path /x:worksheet[1]/x:sheetData[1]/x:row[5]/x:c[3]: Cell contents have invalid value '0' for type 'Boolean'.
> Part /xl/worksheets/sheet1.xml, Path /x:worksheet[1]/x:sheetData[1]/x:row[4]/x:c[3]: Cell contents have invalid value '0' for type 'Boolean'.
> Part /xl/worksheets/sheet1.xml, Path /x:worksheet[1]/x:sheetData[1]/x:row[3]/x:c[3]: Cell contents have invalid value '1' for type 'Boolean'.

**Expected**

`OpenXmlValidator.Validate()` should not produce errors.

twsouthwick:
Fixes #887

### DIFF
--- a/src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
+++ b/src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
@@ -116,25 +116,21 @@ namespace DocumentFormat.OpenXml.Spreadsheet
         /// <param name="value">The result if successful.</param>
         /// <returns>Success or failure</returns>
         public bool TryGetBoolean(out bool value)
-            => bool.TryParse(InnerText, out value) || TryGetBooleanFromInteger(out value);
-
-        private bool TryGetBooleanFromInteger(out bool value)
         {
-            if (int.TryParse(InnerText, out var result))
+            switch (InnerText)
             {
-                switch (result)
-                {
-                    case 0:
-                        value = false;
-                        return true;
-                    case 1:
-                        value = true;
-                        return true;
-                }
+                case "0":
+                case "false":
+                    value = false;
+                    return true;
+                case "1":
+                case "true":
+                    value = true;
+                    return true;
+                default:
+                    value = false;
+                    return false;
             }
-
-            value = false;
-            return false;
         }
 
         private static string ToCellFormat(DateTime dateTime)
--- a/src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
+++ b/src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
@@ -116,7 +116,26 @@ namespace DocumentFormat.OpenXml.Spreadsheet
         /// <param name="value">The result if successful.</param>
         /// <returns>Success or failure</returns>
         public bool TryGetBoolean(out bool value)
-            => bool.TryParse(InnerText, out value);
+            => bool.TryParse(InnerText, out value) || TryGetBooleanFromInteger(out value);
+
+        private bool TryGetBooleanFromInteger(out bool value)
+        {
+            if (int.TryParse(InnerText, out var result))
+            {
+                switch (result)
+                {
+                    case 0:
+                        value = false;
+                        return true;
+                    case 1:
+                        value = true;
+                        return true;
+                }
+            }
+
+            value = false;
+            return false;
+        }
 
         private static string ToCellFormat(DateTime dateTime)
             => dateTime.ToString(DateTimeFormatString, CultureInfo.InvariantCulture);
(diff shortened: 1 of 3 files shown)


twsouthwick:
@tomjebo Can you verify that a `0` or `1` are valid boolean values for cells?

tomjebo:
From MS-OI29500 2.1.659: ...If the cell contains a Boolean, the value shall be 1 to specify true and 0 to specify false.

twsouthwick:
So can it be `true` or `false` as well, or only `1`/`0`?

tomjebo:
Yes, but they MUST be lower case. Excel reads this as lower case "true" or "false" but will only write as "0" or "1". Excel will throw a message dialog and ask to "correct" the underlying XML if the type is "b" and the value is anything but "0", "1", "true", or "false".

src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
```diff
@@ -116,7 +116,26 @@ public bool TryGetDecimal(out decimal value)
         /// <param name="value">The result if successful.</param>
         /// <returns>Success or failure</returns>
         public bool TryGetBoolean(out bool value)
-            => bool.TryParse(InnerText, out value);
+            => bool.TryParse(InnerText, out value) || TryGetBooleanFromInteger(out value);
+
+        private bool TryGetBooleanFromInteger(out bool value)
```
mkorsukov, line 121: This method can be `static`.

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellTests.cs
```diff
@@ -19,6 +19,8 @@ public class CellTests
         [InlineData("false", CellValues.Boolean, true)]
         [InlineData("true", CellValues.Boolean, true)]
         [InlineData("other", CellValues.Boolean, false)]
+        [InlineData("0", CellValues.Boolean, true)]
```
mkorsukov, line 22: Should it provide a `false` result instead?

twsouthwick: These tests are actually checking if validation flags the contents or not. The `true` here means that it is a valid value for a boolean

src/DocumentFormat.OpenXml/Spreadsheet/CellValue.cs
twsouthwick: It was using an instance member `InnerText`

twsouthwick:
Thanks @mkorsukov for the comments!

twsouthwick:
@tomjebo can you take another look and approve if it looks good? I'd like to get this out as a 2.12.3 release

tomjebo (approved):

igitur:
> Yes, but they MUST be lower case. Excel reads this as lower case "true" or "false" but will only write as "0" or "1". Excel will throw a message dialog and ask to "correct" the underlying XML if the type is "b" and the value is anything but "0", "1", "true", or "false".

Thanks, this actually highlighted some issues in ClosedXML. We were writing out Boolean values using propercasing. Nobody every complained, though.

PR #896, opened 2021-02-24
Office 2019 Updates (#882)

tomjebo:
* This commit adds support for the Office 2019 file format types, elements and attributes.


twsouthwick (approved):
LGTM. I don't see a CHANGELOG update. Will that come later?

PR #901, opened 2021-02-26
Allow exponents for numeric cell values

igitur:
Follow-up to #890 . This fixes a validation issue where numbers in exponential format were not recognised.


twsouthwick:
Thanks! Can you add some tests to the parsing tests as well?

igitur:
I did add 2 tests. Do you require any further tests?

twsouthwick:
Those tests are for the validation. There's also tests for CellValue itself for the `TryGet...` methods that it would be good to have these as well. They currently just take the value, so you may need to expand the theories to take in the string and the expected value.

igitur:
Ah, I see what you mean. I created new tests that accept a `string`. Passing a number in exponential format into the existing tests didn't always result in the expected action, as `3E+4` will implicitly convert to `30000`. Well, if they're of type `double` it's the same number, but I think you know what I mean.

twsouthwick (approved):
LGTM! Thanks for the extra tests!

PR #905, opened 2021-03-02
Update README.md to point to new Q&A forums

mikeebowen:
The msdn forums are deprecated and new forum questions should be posted to the "office-addins-dev" tag per https://docs.microsoft.com/en-us/answers/products/m365 under Office Client development


tomjebo (approved):
Thanks Mike!

PR #906, opened 2021-03-05
Tests for double and decimal CellValue ctors

tomjebo:
* adding tests for double and decimal ctors for CellValue


test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
```diff
@@ -74,6 +76,26 @@ public void CellDoubleTest(double num)
             Assert.Equal(num, result);
         }
 
+        [Fact]
+        public void CellDoubleCultureTest()
+        {
+            // Change current culture
+            CultureInfo culture, oldculture;
```
twsouthwick, line 83: can you create a helper for this? Something like this:

```
internal readonly struct CultureInfoTester:IDisposable
{
private readonly CultureInfo _old;

public CultureInfoTester(CultureInfo desired){
_old = Thread.CurrentThread.CurrentCulture;
Thread.CurrentThread.CurrentCulture = desired;
}

public void Dispose(){
Thread.CurrentThread.CurrentCulture = _old;
}
}
```

This can then be used to wrap the methods

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
```diff
@@ -74,6 +76,26 @@ public void CellDoubleTest(double num)
...
+            oldculture = Thread.CurrentThread.CurrentCulture;
+            culture = CultureInfo.CreateSpecificCulture("fr-FR");
```
twsouthwick, line 85: It would be good to have a few cultures to verify this: ie en-US, fr-FR, ru-RU, zh-CH, tr-TR, ar-EG, etc (get in different ways of handling commas, left-to-right, etc)

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
```diff
@@ -74,6 +76,26 @@ public void CellDoubleTest(double num)
             Assert.Equal(num, result);
         }
 
+        [Fact]
+        public void CellDoubleCultureTest()
```
twsouthwick, line 80: what about for int/dates/etc? May not be as big of a deal, but with exponential support added in, it may

twsouthwick (changes requested):
Can you update the title of the PR title with details?

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
tomjebo: Yes, I'll spend some more time on this to see what makes sense.

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
tomjebo: Dates definitely could be included. Not sure about int but I'll take some time to check them out.

twsouthwick:
Any movement here, @tomjebo?

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
tomjebo: So something like this? I just pushed the helper.

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
```diff
@@ -74,6 +77,28 @@ public void CellDoubleTest(double num)
...
+        [InlineData("zh-CH")]
+        [InlineData("ru-RU")]
+        [InlineData("en-US")]
+        [InlineData("tr-TR")]
+        [InlineData("ar-EG")]
+        [Theory]
+        public void CellDoubleCultureTest(string culture)
+        {
+            // Change current culture
+            _ = new CultureInfoTester(new CultureInfo(culture));
```
twsouthwick, line 91: you'll want to do this in a `using`:

```cs
using(new CultureInfoTester(culture))
{
 ...
}
```

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
```diff
@@ -180,6 +205,26 @@ public void CellDecimalTestNegative(string input)
             Assert.False(value.TryGetDecimal(out _));
         }
 
+        [Fact]
+        public void CellDecimalCultureTest()
+        {
+            // Change current culture
+            CultureInfo culture, oldculture;
```
twsouthwick, line 212: Why not use the new helper?

test/DocumentFormat.OpenXml.Tests/Common/CultureHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace DocumentFormat.OpenXml.Tests.Common
+{
+    internal readonly struct CultureInfoTester : IDisposable
+    {
+        private readonly CultureInfo _old;
+
+        public CultureInfoTester(CultureInfo desired)
```
twsouthwick, line 15: You could provide a string overload that creates a cultureinfo to minimize `new CultureInfo(...)` at the call site

twsouthwick (changes requested):
Yeah, that's the right idea. Forgot to put usings around it though

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
tomjebo: I checked dates and they don't work the same way as decimal and double numbers. I tested a couple different dates but Excel seems to pay attention to the incoming format, unlike decimal and double.

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
tomjebo: just hadn't applied it, was waiting for your confirmation on the approach. it's applied now with your suggestions.

test/DocumentFormat.OpenXml.Tests/Common/CultureHelper.cs
tomjebo: Done

twsouthwick (approved):

test/DocumentFormat.OpenXml.Tests/Spreadsheet/CellValueTests.cs
tomjebo: done also

twsouthwick (approved):

PR #928, opened 2021-04-28
Add commentsids part to maindocumentpart relationship

tomjebo:


test/DocumentFormat.OpenXml.Packaging.Tests/PartConstraintRuleTests.cs
```diff
@@ -50,6 +50,20 @@ public void AllExpectedParts()
                 .OrderBy(v => v, StringComparer.Ordinal)
                 .ToList();
 
+#if DEBUG
+            var tmp = Path.GetTempFileName();
```
twsouthwick, line 54: This seems to be duplicated a few times. Is this to help with generating the parts tests?

Could you consolidate any of this to a centralized extension method or something?

tomjebo: Yes, it's for the part constraint testing updates after adding parts. And yes, I probably can turn this into an extension so I'll add something to this commit.

lastthyme (approved):

test/DocumentFormat.OpenXml.Tests/Common/ITestOutputHelperExtenstions.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Xunit.Abstractions;
+
+namespace DocumentFormat.OpenXml.Tests.Common.OutputHelperExtensions
+{
+    public static class ITestOutputHelperExtenstion
+    {
+        static public void WriteObjectToTempFile(this ITestOutputHelper output, object obj)
```
twsouthwick, line 16: You can mark this with the attribute `[Conditional("DEBUG")]` and then you don't need to if/def - it'll only happen on debug

twsouthwick (approved):
One thought, but LGTM

PR #952, opened 2021-06-23
Provide direct support for Linq to XML

ThomasBarnekow:
With this PR, I would like to introduce direct support for Linq to XML in the Open XML SDK as discussed during the Microsoft DevDays Redmond 2021 Online.

The first commit adds the first pieces (some of which are only examples) required for providing at least basic support for Linq to XML without the need for any further library. Those pieces are neither complete nor the final word on how we should integrate this functionality.

- The namespace-related classes (`W`, `W10`, and `W14`) are only a small subset of the classes required to fully cover all schemas. They only serve as an example to show what those classes would look like. We need to discuss whether it makes sense to generate those classes, e.g., with a view to future enhancements of the Open XML schemas.

- There are alternatives to the methods provided by the `LinqExtensions` class that would provide for a better developer experience. At the moment, the `LinqExtensions` class simply replicates the functionality currently provided by the Open XML PowerTools (however, rewritten and with further enhancements by myself). An alternative would be to add methods to the `OpenXmlPart` class that would also take care of the fact that users might be using Linq to XML and the strongly typed classes together. The Open XML PowerTools provide classes like `PowerToolsBlock`, `StronglyTypedBlock`, and `PowerToolsBlockExtensions` (all contributed by myself) to deal with the coexistence. They would no longer be required if we integrate Linq to XML support more directly.

So, this is to get the discussion started. I'm looking forward to your comments.


twsouthwick:
Can you include some tests to show what the expected use is? Many of these things don't feel right exposing as public endpoints - I'm interested in what is the API experience you want to enable.

src/DocumentFormat.OpenXml/Linq/LinqExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            using Stream stream = part.GetStream();
+            if (stream.Length > 0)
+            {
+                using XmlReader xmlReader = XmlReader.Create(stream);
+                rootDocument = XDocument.Load(xmlReader);
+            }
+            else
+            {
+                if (LinqUtility.OpenXmlPartInfos.TryGetValue(part.GetType(), out OpenXmlPartInfo? partInfo))
```
twsouthwick, line 42: Check out the `part.Metadata` API - it should have much of what you're looking for (can't remember the exact name, though)

ThomasBarnekow: Do you mean the classes in the `Framework\Metadata` folder? There is a class called `ElementLookup` which seems to provide that information. It seems to require a deeper look, though.

ThomasBarnekow:
> Can you share a high level view of how you expect the APIs to be used?
> 
> Many of these things don't feel right exposing as public APIs (ie the SchemaAttr is obsoleted for a reason - there are better APIs available internally that we could expose externally).

A very high-level view would be that only the `LinqExtensions` and the namespace-related classes (e.g., `W`) would have to be public. This is in support of the use cases like the one I presented. You can also find tons of examples in the Open XML PowerTools repo.

We should talk about whether reading and writing of `XDocument` and/or `XElement` instances should be provided through extension methods (of `LinqExtensions`) or rather instance methods of `OpenXmlPart`. 

Regarding the "better APIs available internally", I would appreciate your guidance. The comment regarding `SchemaAttrAttribute` did not mention any specific other API, so I just noted that it would have to be replaced when the other API (whatever it is) is available. If there is already an API that can be used instead, I'd be happy to use that.

ThomasBarnekow:
@twsouthwick, here is a quick example from my talk. It showcases the `GetXElement()` extension method and the `W` class.

```csharp
[Fact]
public void CanRemoveSdtsUsingLinqToXml()
{
    // Arrange.
    using WordprocessingDocument wordDocument = WordprocessingDocumentFromResource(Path);
    XElement document = wordDocument.MainDocumentPart.GetXElement();

    // Act.
    var transformedDocument = (XElement) RemoveSdtsTransform(document);

    // Assert.
    Assert.NotEmpty(document.Descendants(W.sdt));
    Assert.Empty(transformedDocument.Descendants(W.sdt));
}

private static object RemoveSdtsTransform(XNode node)
{
    return node switch
    {
        XElement element when element.Name == W.sdt =>
            element.Element(W.sdtContent)?
                   .Nodes()
                   .Select(RemoveSdtsTransform),

        XElement element =>
            new XElement(element.Name,
                         element.Attributes(),
                         element.Nodes()
                                .Select(RemoveSdtsTransform)),
        _ => node
    };
}
```

This is in contrast to the following example, which uses the strongly typed classes:

```csharp
[Fact]
public void CanRemoveSdtsUsingStronglyTypedClasses()
{
    // Arrange.
    using WordprocessingDocument wordDocument = WordprocessingDocumentFromResource(Path);
    Document document = wordDocument.MainDocumentPart.Document;

    // Act.
    var transformedDocument = (Document) RemoveSdtsTransform(document);

    // Assert.
    Assert.NotEmpty(document.Descendants<SdtBlock>());
    Assert.NotEmpty(document.Descendants<SdtRun>());

    Assert.Empty(transformedDocument.Descendants<SdtBlock>());
    Assert.Empty(transformedDocument.Descendants<SdtRun>());
}

private static object RemoveSdtsTransform(OpenXmlElement element)
{
    return element switch
    {
        SdtBlock sdt => sdt.SdtContentBlock.Elements().AggregateResultsOf(RemoveSdtsTransform),
        SdtRun sdt => sdt.SdtContentRun.Elements().AggregateResultsOf(RemoveSdtsTransform),
        SdtRow sdt => sdt.SdtContentRow.Elements().AggregateResultsOf(RemoveSdtsTransform),
        SdtCell sdt => sdt.SdtContentCell.Elements().AggregateResultsOf(RemoveSdtsTransform),

        _ => element.TransformAggregating(RemoveSdtsTransform)
    };
}
```

twsouthwick:
Ah ok. I'm thinking this will be more appropriate in a separate assembly. For v3, I'd love to separate out the core package reading from the strongly typed classes. This would allow a set up such as the following:

```
DocumentFormat.OpenXml.Framework
                               |
        |--------------------------|
DF.Oox.Schema             DF.Oox.Linq
```

This is me thinking aloud, but this set of functionality is potentially a good justification for a separation. If someone just wants to go about it the LINQ way, it would be *much* smaller of an app.

For now, could you create this in a new project? It'll be easy to combine them, but potentially harder to separate later.

ThomasBarnekow:
@twsouthwick, do you have more information on the demarcation line between the `Framework` and `Schema` assemblies? Would the `Schema` contain everything that is currently contained in the `GeneratedCode` folder and the `Framework` everything else? What about the internal `Framework` classes, properties, and methods? Should the `Schema` and `Linq` assemblies have access to those or would you make many more elements publicly accessible?

ThomasBarnekow:
@twsouthwick, I've thought about this a bit more in the context of the "alternatives to the methods provided by the `LinqExtensions` class that would provide for a better developer experience" that I mentioned when opening the PR. As one alternative (which I would likely prefer) would be to extend the `OpenXmlPart` class, separating the assemblies would mean having one part in the `Framework` (i.e., the enhancement of the `OpenXmlPart`) and another one in `Linq` (i.e., the namespace-related classes such as `W`). Further, the namespace-related classes are really small compared to the strongly typed classes and would not warrant a separation due to size.

Would you be open to having the `Linq`-related features in the `Framework` while potentially separating the `Schema` (which is likely a significant assembly.)

ThomasBarnekow:
@twsouthwick and @tomjebo, the latest commit shows how I would actually want to integrate the Linq to XML support, i.e., by enhancing the `OpenXmlPart` and `OpenXmlPartRootElement` classes.

Most of the feature is implemented in the `OpenXmlPart` class. The changes made to the `OpenXmlPartRootElement` class only make sure the the `RootXElement` (which would replace the `GetXElement()` extension method) is properly synchronized with the `RootElement` (or `InternalRootElement`).

Have a look at `OpenXmlPartTests` (added in the last commit) to see how this can be used. In my eyes, the beauty of this is that developers can use the strongly typed classes and Linq to XML pretty much at the same time without running into issues. This would provide for a great developer experience.

Using the enhanced `OpenXmlPart` and `OpenXmlPartRootElement` classes, we would only need the namespace-related classes. All other classes I added to demonstrate the "extension way of integrating Linq to XML" would no longer be required.

ThomasBarnekow:
@twsouthwick and @tomjebo, I removed the extension-related classes to avoid any confusion as to how I propose to integrate the Linq to XML support. With this, the `DocumentFormat.OpenXml.Linq` namespace only contains examples of namespace-related classes.

We should talk about the enhancements made to the `OpenXmlPart` and `OpenXmlPartRootElement` classes. Another question is whether we want to generate the namespace-related classes.

twsouthwick:
Let's keep this focused on the updates for OpenXmlElement/Part - I kind of want to think a bit more about how we want to expose all the `XName`s. I really like the easy access for the `XElement` root, though!

ThomasBarnekow:
Yes, we can focus on the updates for `OpenXmlPart` and `OpenXmlPartRootElement`.

To provide a data point for "all the `XName`s", the Open XML PowerTools provide those in about 3,800 lines of code in `PtOpenXmlUtil.cs`. Looking at the strongly typed classes, `schemas_openxmlformats_org_wordprocessingml_2006_main.g.cs` alone has more than 66,000 lines of code. Thus, my concern would not be the size but rather how we'd keep it in sync with an evolving schema (which is why code generation is an option).

ThomasBarnekow:
I fixed the netstandard1.3 build, which failed because:

- the `System.Xml.Reader` and other classes required an additional NuGet package for that target and
- the `using System.Xml` statement was excluded by a preprocessor statement (because the `FEATURE_XML_SCHEMA` variable was not set for the netstandard1.3 target).

There are still warnings related to one or two of the sample namespace-related classes, which have fields names beginning with an underscore, which is not CLS-compliant. We can either accept this to be compatible with the Open XML PowerTools (from where those examples are taken) or change it and accept breaking changes (which could be fine because developers would have to switch to the `OpenXmlPart` instance methods anyhow.

twsouthwick:
Can you remove the XName classes? Once we ship it, we can't move it, so I'd like to review how we'll expose them. I'm not so worried about the number of lines but rather what the API shape looks like.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -3,14 +3,17 @@
 
 using System;
 using System.Collections.Generic;
+#if FEATURE_XML_SCHEMA
```
twsouthwick, line 6: Can you move this down to the lower #if/dev (and keep the extra newlin?

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -141,11 +142,13 @@ internal sealed override OpenXmlPart NewPart(string relationshipType, string con
             {
                 return TargetPathOfWord;
             }
-            else if (TargetPathOfExcel is not null && package.ApplicationType == ApplicationType.Excel)
+
+            if (TargetPathOfExcel is not null && package.ApplicationType == ApplicationType.Excel)
```
twsouthwick, line 146: I'd like to keep the else if to group things logically that we expect it to be one or the other here

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -380,15 +348,13 @@ public void FeedData(Stream sourceStream)
...
-            using (Stream targetStream = GetStream(FileMode.Create))
-            {
-                sourceStream.CopyTo(targetStream);
-            }
+            using Stream targetStream = GetStream(FileMode.Create);
+            sourceStream.CopyTo(targetStream);
         }
 
-        #endregion
+#endregion
```
twsouthwick, line 355: can you revert the spacing changes here?

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -438,10 +404,7 @@ public void ValidateXml(XmlSchemaSet schemas, ValidationEventHandler validationE
 
             using (var partStream = GetStream())
             {
-                if (validationEventHandler is not null)
-                {
-                    xmlReaderSettings.ValidationEventHandler += validationEventHandler;
-                }
+                xmlReaderSettings.ValidationEventHandler += validationEventHandler;
```
twsouthwick, line 407: why this change?

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -480,17 +443,154 @@ public void ValidateXml(string schemaFile, ValidationEventHandler validationEven
         /// Gets the root element of the current part.
         /// Returns null when the current part is empty or is not an XML content type.
         /// </summary>
-        public OpenXmlPartRootElement? RootElement
+        public OpenXmlPartRootElement? RootElement => PartRootElement;
+
+#endregion
+
+#region Linq to XML
```
twsouthwick, line 450: Could you create a new file for a partial class of OpenXmlPart.Linq? These class are getting too large and I'm not a fan of regions

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -480,17 +443,154 @@ public void ValidateXml(string schemaFile, ValidationEventHandler validationEven
...
+
+        /// <summary>
+        /// Gets or sets
+        /// </summary>
+        public XElement? RootXElement
```
twsouthwick, line 455: Should this be exposed via the `OpenXmlPartRootElement` instead? That way it would always be in sync

twsouthwick (changes requested):

twsouthwick:
Sorry for the delays - I've been on vacation and am back now :)

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
ThomasBarnekow: The `OpenXmlPart` class has not been split into partial classes. I added the Linq region because there were already 9 other regions in there. The Linq feature accounts for 140 of 908 line items. Do you really want to do this?

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
ThomasBarnekow: It should not be exposed that way, because the `OpenXmlPartRootElement` only plays a role if both the Linq to XML approach and the strongly typed classes are used at the same time.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
ThomasBarnekow: Reinstated.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
ThomasBarnekow: Done. I've done that for all `#region` and `#endregion` occurrences. I don't remember changing those.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
ThomasBarnekow: `validationEventHandler` is not nullable, so the condition is always true.

ThomasBarnekow:
No worries. Just made most of the changes you suggested. Please consider my comment on the partial class.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
ThomasBarnekow: Done

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
twsouthwick: I'm fine with that

test/DocumentFormat.OpenXml.Tests/Linq/OpenXmlPartTests.cs
```diff
@@ -0,0 +1,175 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using DocumentFormat.OpenXml.Packaging;
+using DocumentFormat.OpenXml.Wordprocessing;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+using Xunit;
+
+#nullable enable
```
twsouthwick, line 11: Should we enable nullable on tests? That probably would be nice to verify things...

test/DocumentFormat.OpenXml.Tests/Linq/W.cs
```diff
@@ -0,0 +1,842 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Xml.Linq;
+
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+
+namespace DocumentFormat.OpenXml.Tests.Linq
```
twsouthwick, line 8: Cool - let's keep this here for now and we can merge it into the main project at some point

twsouthwick (approved):
The LGTM. We'll want to hold off on merging it for now so we can get a 2.13.1 with some behavior fixes

test/DocumentFormat.OpenXml.Tests/Linq/W.cs
ThomasBarnekow: Yes, I needed to retain that here for unit testing.

test/DocumentFormat.OpenXml.Tests/Linq/OpenXmlPartTests.cs
ThomasBarnekow: If the assembly under test has nullability enabled, it would make sense to do that for the corresponding unit tests, too. This way, you would more likely identify cases where nullability is incorrect. For example, see my other issue on nullabity of the MainDocunentPart.Document property, for example.

twsouthwick: Created an issue to track that

ThomasBarnekow:
@twsouthwick, agree. We also need to discuss how we want to create those namespace-related classes (e.g., `W`, `W14`).

ThomasBarnekow:
@twsouthwick, I added a commit for the namespace-related classes generated from the `ElementMetadata` and `AttributeMetadata` provided by the SDK. You will see several generated classes in the `Linq\GeneratedCode` folder within the `DocumentFormat.OpenXml` project. The code generation is currently performed in a unit test (see `Linq\CodeGenerationTests.cs` with the `DocumentFormat.OpenXml.Tests` project).

The generated code adds 199 kilobytes (3.6%) to the `DocumentFormat.OpenXml.dll` assembly (in Release configuration), increasing the size from 5,520 to 5,719 kilobytes.

It would also be possible to play with the XML comments generated for fields. For example, we could:

- entirely remove XML comments;
- add remarks on strongly-typed classes and/or properties corresponding to the field; and/or
- add remarks on parent and child elements.

twsouthwick:
@ThomasBarnekow Can you open a separate PR to keep this one just the XElement changes?

twsouthwick:
A couple of thoughts:

- Can you create a separate assembly for this? Maybe DocumentFormat.OpenXml.Linq
- What about using C# 9.0 Source Generators?

ThomasBarnekow:
@twsouthwick, regarding the **separate assembly**, I would like to make it easy for developers. If you separated the core framework from both (1) the strongly-typed classes and (2) the Linq classes I just added, then having separate assemblies would be a natural way. Developers would have to make a choice in any case. So, are you definitely going to "separate out the core package reading from the strongly typed classes" in v3 as described in a previous comment?

> Ah ok. I'm thinking this will be more appropriate in a separate assembly. For v3, I'd love to separate out the core package reading from the strongly typed classes. This would allow a set up such as the following:
> 
> ```
>         DocumentFormat.OpenXml.Framework
>                       |
>         |--------------------------|
> DF.Oox.Schema                 DF.Oox.Linq
> ```
> 
> This is me thinking aloud, but this set of functionality is potentially a good justification for a separation. If someone just wants to go about it the LINQ way, it would be _much_ smaller of an app.
> 
> For now, could you create this in a new project? It'll be easy to combine them, but potentially harder to separate later.

Separating only that one feature, which adds 199 kilobytes (3.6%) to the 5,520 kilobytes of the current `DocumentFormat.OpenXml.dll` assembly (in Release configuration), would not make too much sense in my mind.

Regarding the **C# 9.0 Source Generators**, I don't have any experience with those yet. What would be the value from your perspective?

Regarding the **separate PR**, that can be done.

ThomasBarnekow:
@twsouthwick, this PR should now be where it was before adding the generated classes.

I will open a separate PR for the generated code. However, there is one question: How often do you change the generated strongly-typed classes? The C# Source Generators must target netstandard2.0. My source generator would have a dependency on the DocumentFormat.OpenXml.dll, which can either be a project reference or a direct DLL reference.

When using a project reference, we would have to multi-target in many cases. When the target framework is only net5.0, for example, the generator can't depend on the DocumentFormat.OpenXml project. In that case, we'll have to target bot net5.0 and netstandard2.0. If we wanted to avoid that, we'd have to depend on some reference DocumentFormat.OpenXml.dll that targets netstandard2.0. This could be an option if the generated strongly-typed classes don't change too often.

ThomasBarnekow:
@twsouthwick, there might be another potential issue with the C# Code Generators. It looks like they might only be useable when targeting a framework that can consume netstandard2.0. This means we could not support the legacy .NET Frameworks 3.5, 4.0, and 4.6.

Would it be a problem to only offer a netstandard2.0 assembly? This could be consumed from .NET Framework 4.6.1 and later.

twsouthwick:
Let's discuss this further in the new pr

ThomasBarnekow:
@twsouthwick, I just opened the new PR. Please have a look.

twsouthwick:
FYI - We're open for merges for 2.14.0 :)

twsouthwick (approved):

PR #953, opened 2021-06-23
Fix nullability annotations of OpenXmlComparableSimpleReference

### LINKED ISSUE #933: Incorrect nullability annotations for OpenXmlComparableSimpleReference operator ==/!=
**Description**

https://github.com/OfficeDev/Open-XML-SDK/blob/main/src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableSimpleReference.cs#L131
https://github.com/OfficeDev/Open-XML-SDK/blob/main/src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableSimpleReference.cs#L147

Operators parameters not marked as nullable which leads to following error when checking derived instances (e.g. StringValue) for null.

Also other operators seems to accept nulls, but I don't use them, so there is no breaking scenario for them from me.

**Information**

- .NET Target: net48, net5.0
- DocumentFormat.OpenXml Version: 2.13.0

**Repro**

```cs
// CS8604	Possible null reference argument for parameter 'left' in 'bool OpenXmlComparableSimpleReference<string>.operator !=(OpenXmlComparableSimpleReference<string> left, OpenXmlComparableSimpleReference<string> right)'.
if (cell.CellReference != null)
```

**Observed**

error generated

**Expected**

no error generated

ThomasBarnekow:
This PR fixes #933.

### DIFF
--- a/src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableSimpleReference.cs
+++ b/src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableSimpleReference.cs
@@ -94,26 +94,28 @@ namespace DocumentFormat.OpenXml
             {
                 return false;
             }
-            else if (Value is null)
+
+            if (Value is null)
             {
                 return false;
             }
-            else if (ReferenceEquals(this, obj))
+
+            if (ReferenceEquals(this, obj))
             {
                 return true;
             }
-            else if (obj is OpenXmlComparableSimpleReference<T> other)
+
+            if (obj is OpenXmlComparableSimpleReference<T> other)
             {
                 return Equals(other);
             }
-            else if (obj is T otherValue)
+
+            if (obj is T otherValue)
             {
                 return otherValue.Equals(Value);
             }
-            else
-            {
-                return false;
-            }
+
+            return false;
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableSimpleReference.cs
```diff
@@ -94,26 +94,28 @@ public override bool Equals(object? obj)
             {
                 return false;
             }
-            else if (Value is null)
+
+            if (Value is null)
```
twsouthwick, line 98: why remove the else?

twsouthwick (dismissed):

twsouthwick:
image

src/DocumentFormat.OpenXml/SimpleTypes/OpenXmlComparableSimpleReference.cs
ThomasBarnekow: It's not required (see the `return false` above).

ThomasBarnekow:
@twsouthwick, that's interesting. I did not get those errors when running the unit tests on my machine. How did you run those unit tests against those legacy frameworks? They are not run when using the Test Explorer or ReSharper.

ThomasBarnekow:
@twsouthwick, please re-run the tests. I hope I fixed them but couldn't verify that it works with .NET Framework 3.5, for example.

twsouthwick:
@ThomasBarnekow  Check out https://github.com/OfficeDev/Open-XML-SDK/blob/main/CONTRIBUTING.md#building for details on running tests for all platforms. It causes memory issues to load all versions of the project by default, so I separated out a while back. Maybe with VS 2022 being 64-bit that won't be as big of a deal.

twsouthwick (approved):

PR #955, opened 2021-06-24
Fix various nullability problems in OpenXmlCompositeElement and OpenXmlElement

sorensenmatias:
The MaybeNull attribute is redundant in these cases - the nullability is well covered by the NotNullIfNotNull attribute solely.

### DIFF
--- a/src/DocumentFormat.OpenXml/OpenXmlCompositeElement.cs
+++ b/src/DocumentFormat.OpenXml/OpenXmlCompositeElement.cs
@@ -228,9 +228,9 @@ namespace DocumentFormat.OpenXml
         }
 
         /// <inheritdoc/>
-        [return: MaybeNull]
         [return: NotNullIfNotNull("newChild")]
-        public override T AppendChild<T>([AllowNull] T newChild)
+        public override T? AppendChild<T>(T? newChild)
+            where T : class
         {
             if (newChild is null)
             {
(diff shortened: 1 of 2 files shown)


src/DocumentFormat.OpenXml/OpenXmlCompositeElement.cs
```diff
@@ -228,9 +228,9 @@ public bool AddChild(OpenXmlElement newChild, bool throwOnError = true)
         }
 
         /// <inheritdoc/>
-        [return: MaybeNull]
         [return: NotNullIfNotNull("newChild")]
-        public override T AppendChild<T>([AllowNull] T newChild)
+        public override T? AppendChild<T>(T? newChild)
+            where T : class
```
twsouthwick, line 233: This is a breaking change - we can't add a generic constraint where there wasn't already

twsouthwick (changes requested):

twsouthwick:
Happy to explore different ways of annotating this stuff, but we can't introduce breaking changes.

src/DocumentFormat.OpenXml/OpenXmlCompositeElement.cs
sorensenmatias: I understand that it looks like a breaking change, but in fact it is not. The generic constraint of the base method is inherited, and the base restricts T to be of type OpenXmlElement. 
Since OpenXmlElement is a class, this new constraint should not matter.

I am however not sure why we have to add the constraint, but it is the only way I could make it compile.
As I said, T is already restricted to OpenXmlElement, so why would you have to specify this constraint? I can try to research more.

sorensenmatias: More information can be found here: https://devblogs.microsoft.com/dotnet/try-out-nullable-reference-types/, see "The issue with T?".

twsouthwick: Thanks for the clarification!I'm on vacation but will look when I get back

twsouthwick (approved):
Lgtm. Thanks for the clarification!

PR #956, opened 2021-06-25
Fix CellDoubleTest

ThomasBarnekow:
On machines (e.g., German) that do not use the invariant culture, the CellDoubleTest() unit test fails because it computes incorrect expected string values for 1.5 and -1.5 (e.g., "1,5" and "-1,5"). This PR fixes that.


twsouthwick (approved):
LGTM!  Thanks!

Issue #962, opened 2021-07-13
Consider providing w14:paraId and w14:textId generators


ThomasBarnekow:
**Description**

`w:p` and `w:tr` elements can have the `w14:paraId` and `w14:textId` attributes, which are defined in [MS-DOCX](https://docs.microsoft.com/en-us/openspecs/office_standards/ms-docx/b839fe1f-e1ca-4fa6-8c26-5954d0abbccd) as `ST_LongHexNumber` values that are unique within the document part as well as greater than `0` and less than `0x80000000`.

Microsoft Word uses a random number generator to generate the values (noting that is not a requirement).

At the moment, the `Paragraph` (`w:p`) and `TableRow` (`w:tr`) classes do not generate values for the `ParagraphId` (`w14:paraId`) and `TextId` (`w14:textId`) attributes. There are also no utility classes or methods for producing compliant values.

Therefore, the question is whether we want to offer any functionality for generating or validating those attribute values.

Providing utility methods would be very straightforward. For example, here is the code (taken from two classes) that I am using in my codebase for creating random `ST_LongHexNumber` values (optionally making sure they are less than `0x80000000` while always guaranteeing they are greater than `0`):

```csharp
private static readonly RNGCryptoServiceProvider Generator = new RNGCryptoServiceProvider();

/// <summary>
/// Creates an ST_LongHexNumber value, masking the most significant byte with
/// the given <paramref name="msbMask" />.
/// </summary>
/// <param name="msbMask">The most significant byte mask.</param>
public static string CreateRandomLongHexNumber(byte msbMask = 0xff)
{
    // Create a four-byte random number, noting that the first byte (data[0])
    // will become the most significant byte in the string value created by
    // the ToHexString() method.
    var data = new byte[4];
    Generator.GetNonZeroBytes(data);
    data[0] &= msbMask;

    return data.ToHexString();
}

/// <summary>
/// Converts the given value into a hexadecimal string, with the first
/// byte in the list being the most significant byte in the resulting
/// string.
/// </summary>
/// <param name="source">The list of bytes to be converted.</param>
/// <returns>A hexadecimal string.</returns>
public static string ToHexString(this IReadOnlyList<byte> source)
{
    var dest = new char[source.Count * 2];

    var i = 0;
    var j = 0;

    while (i < source.Count)
    {
        byte b = source[i++];
        dest[j++] = ToCharUpper(b >> 4);
        dest[j++] = ToCharUpper(b);
    }

[...]

PR #997, opened 2021-07-24
Generate unique w14:paraId values

ThomasBarnekow:
This PR addresses issue #962 (Consider providing w14:paraId and w14:textId generators).

At the moment, this is an incomplete implementation that uses the `Paragraph` class to demonstrate the potential API and usage of the feature. Corresponding enhancements will have to be made for the `TableRow` class.

The only public API change so far is the addition of two instance methods to the `Paragraph` class:

```csharp
/// <summary>
/// For Paragraph instances that have already been added to the DOM tree,
/// sets the <see cref="ParagraphId" /> to a random and unique value.
/// </summary>
/// <exception cref="InvalidOperationException">If the paragraph has not been added to a DOM tree.</exception>
public void SetUniqueParagraphId()
{
    Document? document = Document;

    if (document is null)
    {
        throw new InvalidOperationException("Paragraph must be added to the DOM tree before calling this method.");
    }

    ParsedState.Attributes.GetProperty(nameof(ParagraphId)).Value = document.CreateUniqueParagraphId();
}

/// <summary>
/// For Paragraph instances that have not yet been added to the DOM tree,
/// sets the <see cref="ParagraphId" /> to a random and unique value.
/// </summary>
/// <param name="document">The <see cref="Document" /> used to create the random and unique value.</param>
/// <returns>This <see cref="Paragraph" /> instance.</returns>
public Paragraph WithUniqueParagraphId(Document document)
{
    ParsedState.Attributes.GetProperty(nameof(ParagraphId)).Value = document.CreateUniqueParagraphId();
    return this;
}
```

As we need to ensure uniqueness within the main document part, A `Document` instance is used to generate random **and** unique `w14:paraId` values. Existing `w14:paraId` attribute values are retrieved and registered when the DOM tree is loaded. Generated `w14:paraId` values are registered in the same way, allowing us to check for and avoid collisions/conflicts.

Have a look at the unit test (test/DocumentFormat.OpenXml.Tests/ParaId/ParaIdTests.cs) to see how this is used.


twsouthwick:
I would prefer to have this be something configurable by the user. Can it be a service set in the OpenSettings when creating a package?

ThomasBarnekow:
> I would prefer to have this be something configurable by the user. Can it be a service set in the OpenSettings when creating a package?

Not sure I fully understand this. What exactly would you make configurable? The registration of the `w14:paraId` values that happens when opening a package? 

The user would have to explicitly call the `SetUniqueParagraphId()` and `WithUniqueParagraphId(Document document)` methods to assign unique `w14:paraId` values. This does not happen "magically". The only "magic" that is done as a prerequisite for ensuring those values are actually unique is the tracking of existing `w14:paraId` values as and when the DOM tree is loaded. If you are never going to call those other methods, you could avoid any overhead (which should be `O(1)`) by also requiring the user to opt into the tracking "magic". However, you would then also have to throw exceptions in case the user calls those methods without previously having opted in. 

A question would be whether you want to allow users to create invalid Open XML by assigning duplicate `w14:paraId` values.

ThomasBarnekow:
@twsouthwick, do you have any feedback on my questions above?

tomjebo:
@ThomasBarnekow 
>>A question would be whether you want to allow users to create invalid Open XML by assigning duplicate w14:paraId values.
No, we don't (want to allow users to create invalid documents). I haven't reviewed the code yet (sorry) but if your description is accurate, then I'm fine with the approach. I will review the code and run some tests as well.

ThomasBarnekow:
@tomjebo, thanks! If you don't want to allow users to create duplicate `w14:paraId` values (and thereby invalid documents), we should not only register those `w14:paraId` values but also throw exceptions as suggested in my comment in the `OpenXmlElement.cs` file (where I also noted that we might want to override the property in the `Paragraph` and `TableRow` classes).

tomjebo (dismissed):
This looks good to me.

tomjebo:
I approved the code. We should of course add table code. I ran tests timing the new code using the same test code Thomas provide but in a console app.
With the paraId generation I get 9.2-9.8 second runs. 
With static paraIds and not using the generation code at all, I get 8.2-8.6 second runs. 
Without adding the paraIds, just a text run, I get 6.8-7.1 second runs.

@ThomasBarnekow @twsouthwick what do you both think?

src/DocumentFormat.OpenXml/Wordprocessing/Document.cs
```diff
@@ -3,20 +3,122 @@
...
+        /// were prevented by our explicit uniqueness check rather than the secure
+        /// random number generator.
+        /// </remarks>
+        internal int PreventedParagraphIdCollisions { get; private set; }
+
         private protected override bool StrictTranslateAttribute(in OpenXmlQualifiedName qname, string? value)
         {
-            if (string.Equals("conformance", qname.Name, StringComparison.Ordinal)
-                && string.Equals("strict", value, StringComparison.Ordinal))
+            if (string.Equals("conformance", qname.Name, StringComparison.Ordinal) &&
```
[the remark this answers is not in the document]
ThomasBarnekow: Correct, this is code beautification.

ThomasBarnekow:
> I approved the code. We should of course add table code. I ran tests timing the new code using the same test code Thomas provide but in a console app.
> With the paraId generation I get 9.2-9.8 second runs.
> With static paraIds and not using the generation code at all, I get 8.2-8.6 second runs.
> Without adding the paraIds, just a text run, I get 6.8-7.1 second runs.
> 
> @ThomasBarnekow @twsouthwick what do you both think?

Thanks, @tomjebo. Yes, we should add code related to the `TableRow` class. As it stands now, the code only demonstrates the idea and principle, using the `Paragraph` class as an example.

Further, I see optimization potential, e.g., by **not** overriding the following method of the `OpenXmlElement` class **but** rather changing the `ParagraphId` properties in the `Paragraph` and `TableRow` classes.

```csharp
private protected void SetAttribute<TSimpleType>(TSimpleType? value, [CallerMemberName] string propertyName = null!)
```

The above method is called by all strongly-typed setters, which means that the following check is performed in all cases:

```csharp
if (this is Paragraph p && propertyName is nameof(Paragraph.ParagraphId) && value is HexBinaryValue hexBinaryValue)
{
    // TODO: Consider throwing an exception in case the value is not unique.
    p.Document?.RegisterParagraphId(hexBinaryValue);
}
```

The relevant code of the `Paragraph` and `TableRow` classes is generated, so I think the code generation would have to take care of that. To determine the extent to which this improves the performance, we could certainly play with the generated code. However, I did not want to confuse anybody by doing that when first creating the PR.

@tomjebo, could you share your benchmark code or at least provide some more detail on what exactly and how you tested? If I read that correctly, we are looking at an order of 20-21% run time increase for adding "static paraIds" (without any checks, right?) and another 12-14% for generating unique paraIds. Right?

tomjebo:
@ThomasBarnekow Here's the code, it's just your test turned into a console app: 

```csharp
using DocumentFormat.OpenXml.Wordprocessing;
using System.IO;
using System.Linq;
using System;

#nullable enable

namespace DocumentFormat.OpenXml.Tests
{
    class Program
    {
        {
            PerformTest();

        }
        private static void CreateTestDocument(Stream stream, int start, int end)
        {
            using var wordDocument = WordprocessingDocument.Create(stream, WordprocessingDocumentType.Document);
            MainDocumentPart part = wordDocument.AddMainDocumentPart();

            var body = new Body();
            var document = new Document(body);

            for (int i = start; i < end; i++)
            {
                // Use WithUniqueParagraphId(document) to create random and unique w14:paraId.
                Paragraph paragraph = new Paragraph(); //.WithUniqueParagraphId(document);

                var run = new Run(new Text("Run with no paraid")); // paragraph.ParagraphId!.InnerText!));
                //paragraph.ParagraphId = "7F000FFF";
                paragraph.AppendChild(run);

                body.AppendChild(paragraph);
            }

            part.Document = document;
        }

        private static void UpdateTestDocument(Stream stream, int start, int end)
        {
            using WordprocessingDocument wordDocument = WordprocessingDocument.Open(stream, true);
            MainDocumentPart part = wordDocument.MainDocumentPart!;

            Document document = part.Document;
            Body body = document.Body!;

            for (int i = start; i < end; i++)
            {
                // Use WithUniqueParagraphId(document) to create random and unique w14:paraId.
                Paragraph paragraph = new Paragraph(); //.WithUniqueParagraphId(document);

                var run = new Run(new Text("Added run with no paraid")); // paragraph.ParagraphId!.InnerText!));
                //paragraph.ParagraphId = "7F000FFF";

                paragraph.AppendChild(run);

                body.AppendChild(paragraph);
            }
        }

        public static void PerformTest()
        {
            var watch = new System.Diagnostics.Stopwatch();

            watch.Start();

            const int iterations = 10;
            const int count = 10000;

            Console.WriteLine($"Performing {iterations} iterations with {count} paragraphs each ...");

            CreateTestDocument(stream, 0, count);

            for (var i = 1; i < iterations; i++)
            {
                UpdateTestDocument(stream, i * count, (i + 1) * count);
            }

            using WordprocessingDocument wordDocument = WordprocessingDocument.Open(stream, true);
            Body body = wordDocument.MainDocumentPart!.Document!.Body!;
            //List<string> paragraphIds = body.Elements<Paragraph>().Select(p => p.ParagraphId!.InnerText!).ToList();
            //var uniqueParagraphIds = new HashSet<string>(paragraphIds);

            //Console.WriteLine($"Have {uniqueParagraphIds.Count} different w14:paraId values.");

            watch.Stop();
            Console.WriteLine($"Execution Time: {watch.ElapsedMilliseconds} ms");
            Console.WriteLine($"Press any key to continue...");
            Console.ReadKey();
        }
    }
}
```

ThomasBarnekow:
Ah, OK. Thinking out loud, I don't know whether the run time results are very meaningful.

The purpose of the unit test is to demonstrate two things:

- Firstly, even for large Word documents, collisions are extremely unlikely, even if no check is performed (owing to the secure random number generator). In a comment to the corresponding issue (see #962), I argued that a novel like War and Peace (561,304 words, 1,225 pages) should have ca. 5.000 to 8.000 paragraphs, which is well below the 10.000 paragraphs created per iteration. In the first few iterations, no collisions were noted in my tests.

- Secondly, the check guarantees that there are no collisions. For example, within 100.000 paragraphs (8 to 20 times War and Peace), a few collisions are likely to occur. The check prevents all of those.

Now, in a typical Word document, you would likely have in the order of hundreds (or even just tens) of paragraphs. Thousands of paragraphs would mean hundreds of pages. So we are looking at far fewer paragraphs, each of which will have more complex markup on average. This then means the code added for this feature will have a proportionally lower run time impact.

Regarding my other questions, would it be possible to change the code generated for the `Paragraph` and `TableRow` classes?

tomjebo:
@ThomasBarnekow 
>>we should not only register those w14:paraId values but also throw exceptions ...

Hm... well, (thinking out loud also :) ) we could be a nanny to the programmer or put up "guard-rails" to keep them from hurting themselves. But the problem here is that this leads us to look at all the other properties in Office Open XML that we map to a class or class member and which should be unique (honestly I don't know how many there are, but willing to be there are more than a few).  Maybe an option would be to disallow direct assignment of the paraId property in the Paragraph class, forcing the programmer to use the unique, non-conflicting generation function. If the programmer really wants to set the paraId in a non-safe way, they can always go down to the XML (via System.XML or some other library).

ThomasBarnekow:
@tomjebo, I think we just need to decide whether or not we must guarantee uniqueness of the paraId values. Your earlier conment suggested that we must guarantee uniqueness. If we don't have to guarantee uniqueness, we can allow programmers to assign duplicate values via the ParagraphId property. 

If you want to guarantee uniqueness without throwing, we should not provide a means to set the ParagraphId via the property, which can currently be used to assign duplicates by accident.

With a view to other attributes having to be unique, I am only aware of the w:id attribute. Would you have another example?

tomjebo:
@ThomasBarnekow there are actually lots of unique ids and names and other attributes (i.e. smtId, uniqueId, name, lots of id, etc...) throughout the standard and our extensions. They may or may not cause problems when Office opens the documents but I'm sure that Office will not generate a document with duplicates. If we provide code to generate an attribute value that is supposed to be unique and Office ensures this is unique, then we must also. This may not be the first attribute or element value that has been under discussion for generation by the SDK code but it's the first I've taken part in. The question of whether we should check ids provided by a programmer is interesting and I'd be ok with that as well. It could ensure that the SDK is not part of corrupt document emission at least for paraId's.

src/DocumentFormat.OpenXml/OpenXmlElement.cs
```diff
@@ -197,7 +198,18 @@ internal virtual void ConfigureMetadata(ElementMetadata.Builder builder)
 
         private protected void SetAttribute<TSimpleType>(TSimpleType? value, [CallerMemberName] string propertyName = null!)
             where TSimpleType : OpenXmlSimpleType
-            => ParsedState.Attributes.GetProperty(propertyName).Value = value;
+        {
+            // TODO: Revisit and add corresponding code for TableRow.
+            // Consider adding the following code to the generated code in schemas_openxmlformats_org_wordprocessingml_2006_main.g.cs.
+            // That way, the checks will only have to be performed for Paragraph and TableRow instances.
+            if (this is Paragraph p && propertyName is nameof(Paragraph.ParagraphId) && value is HexBinaryValue hexBinaryValue)
```
twsouthwick, line 205: This is not an ideal place for this. Now `OpenXmlElement` has specific knowledge of these things.

Maybe add a virtual method (for now make it private protected) that can handle this. It may make sense to make this `SetAttribute` virtual but I'm not certain

src/DocumentFormat.OpenXml/Wordprocessing/Document.cs
```diff
@@ -3,20 +3,122 @@
...
 using DocumentFormat.OpenXml.Framework;
 using System;
+using System.Collections.Generic;
+using System.Security.Cryptography;
 
 namespace DocumentFormat.OpenXml.Wordprocessing
 {
     public partial class Document
     {
+        private static readonly RNGCryptoServiceProvider Generator = new();
```
twsouthwick, line 13: Can create a service that abstracts these things? It should handle the generation, the duplicate checking, etc

twsouthwick (changes requested):

src/DocumentFormat.OpenXml/OpenXmlElement.cs
ThomasBarnekow: I could not agree more, which is why I added the comment on adding the code to the generated `Paragraph` and `TableRow` classes. Would that be a possibility?

src/DocumentFormat.OpenXml/Wordprocessing/Document.cs
ThomasBarnekow: What exactly do you have in mind? How should `Paragraph` and `TableRow` instances access that service? Wouldn't that add a potentially unnecessary layer?

Why not have the `Document` instance provide that service? It defines the scope within which `w14:paraId` values must be unique. Thus, we'd need one service instance per `Document` instance anyhow.

ThomasBarnekow:
Why did the check fail?

twsouthwick:
```
##[error]test\DocumentFormat.OpenXml.Tests\ParaId\ParaIdTests.cs(4,1): Error SA1210: Using directives should be ordered alphabetically by the namespaces
```

src/DocumentFormat.OpenXml/OpenXmlElement.cs
twsouthwick: You can already do what's needed here (Document at least already has a partial class for this kind of stuff: https://github.com/OfficeDev/Open-XML-SDK/blob/main/src/DocumentFormat.OpenXml/Wordprocessing/Document.cs)

src/DocumentFormat.OpenXml/Wordprocessing/Document.cs
twsouthwick: The Document instance can provide it, but the implementation for it would be much easier to test if it were in its own service. The Document class (and all the strongly-typed classes provided) should have limited additional logic.

src/DocumentFormat.OpenXml/Wordprocessing/Paragraph.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        /// <summary>
+        /// For Paragraph instances that have not yet been added to the DOM tree,
+        /// sets the <see cref="ParagraphId" /> to a random and unique value.
+        /// </summary>
+        /// <param name="document">The <see cref="Document" /> used to create the random and unique value.</param>
+        /// <returns>This <see cref="Paragraph" /> instance.</returns>
+        public Paragraph WithUniqueParagraphId(Document document)
+        {
+            ParsedState.Attributes.GetProperty(nameof(ParagraphId)).Value = document.CreateUniqueParagraphId();
```
twsouthwick, line 99: If this is not an exposes property, we can expose it

src/DocumentFormat.OpenXml/Wordprocessing/Document.cs
```diff
@@ -3,20 +3,122 @@
...
+        /// </summary>
+        /// <remarks>
+        /// The previously existing w14:paraId values are registered when reading the DOM tree
+        /// from the OpenXmlPart, using the RegisterParagraphId() methods.
+        /// </remarks>
+        /// <returns>A random and unique w14:paraId value.</returns>
+        internal HexBinaryValue CreateUniqueParagraphId()
+        {
+            HexBinaryValue paragraphId = CreateRandomParagraphId();
+            while (!ParagraphIds.Add(paragraphId.InnerText!))
```
twsouthwick, line 115: Check if not null or make the ParagraphIds take `string?`

src/DocumentFormat.OpenXml/OpenXmlElement.cs
ThomasBarnekow: The `SetAttribute` method is called by the `ParagraphId` properties of the `Paragraph` and `TableRow` classes. It would be best to change those properties to perform the check that is currently performed by this `SetAttribute` method. However, those properties are **generated**. To my knowledge, they cannot be implemented (again) in a partial class unless they are no longer generated.

src/DocumentFormat.OpenXml/Wordprocessing/Document.cs
ThomasBarnekow: What exactly do you mean?

The `HexBinaryValue` instances created by `CreateRandomParagraphId()` cannot be `null`. Visual Studio just thinks `InnerText` can be `null` because it is nullable. I'm not sure whether that makes sense (because a non-null `HexBinaryValue` should contain a valid, non-null value and, therefore, non-null `InnerText`).

The method creates a `HexBinaryValue` because the downstream code expects a `HexBinaryValue`.

twsouthwick: Yes, it *should* contain a non-null value at this point, but nothing is guaranteeing that since `HexBinaryValue` *may* contain a null `InnerText`. Can you verify it's null and throw if not, or allow the hashset to take a null? If you just need the innertext, have the `CreateRandomParagraphId` create that instead of a HexBinaryValue

src/DocumentFormat.OpenXml/Wordprocessing/Document.cs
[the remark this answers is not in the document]
twsouthwick: The pattern that existed is the current one most places in the project. It does make it a bit difficult to review changes that mix random style changes with functional changes.

src/DocumentFormat.OpenXml/Wordprocessing/Document.cs
ThomasBarnekow: **<w:rant>**

It _will definitely_ contain a nun-noll value at this point. The `HexBinaryValue` returned as follows is guaranteed to contain a non-null `TextValue` (assigned by the constructor) and, therefore, non-null `InnerText` (and `Value`):

```csharp
return HexStringFactory.Create(bytes);
```

The corresponding unit test in `HexBinaryValueTests` establishes that fact (for empty byte arrays and those with one or more elements), although it is not easy to see because the unit test methods use either the `Value` or the `InnerText` property, both of which delegate to `TextValue`, though. So, to be honest, I was confused as to which property I'm supposed to use (i.e., `Value` or `InnerText`; the documentation tells you to not use `TextValue`).

A `HexBinaryValue` instance _may only_ contain a null `InnerText` (or `TextValue` or `Value`) if I am passed a `HexBinaryValue` that was created outside of my control. In this case, however, I am fully in control. The `HexBinaryValue` value is created by a utility method of the same class. Having to check for `null` in this case would be both weird and either:

- consume CPU cycles for no good reason or
- hopefully, be optimized away by the compiler.

Can you explain to me why the `HexBinaryValue` constructor even takes a `null` value? Given an empty `byte` array or `ReadOnlySpan<byte>`, the two `HexBinaryValue.Create()` methods will `return new HexBinaryValue(string.Empty)`. This is either inconsistent or the way to go unless you go one step further and demand a `HexBinaryValue` be an actual hex binary value, which would be `00` at least (i.e., same rules as for `int`, `uint`, etc.).

**</w:rant>**

**<w:wayForward>**

Having said that, I have taken a different route and indeed created a `ParagraphIdService` (as you suggested) that does not make use of the `HexBinaryValue` class anymore (to be reusable also for the Linq to XML use case). Since that service still requires code that produces a hex string from a byte array, I've carved out the code used by the `HexBinaryValue` class and created a `HexStringFactory` class with `Create()` methods that create `string` rather than `HexBinaryValue` instances. Both the `HexBinaryValue.Create(ReadOnlySpan<byte> bytes)` method and the `ParagraphIdService` class use the `HexStringFactory` to create hex strings. The (`public` or `internal`) `HexBinaryValue.Create(ReadOnlySpan<byte> bytes)` then simply looks like this:

```csharp
static HexBinaryValue Create(ReadOnlySpan<byte> bytes)
{
    return new(HexStringFactory.Create(bytes));
}
```

I will soon push new commits with those changes.

**</w:wayForward>**

ThomasBarnekow:
@tomjebo and @twsouthwick, please have a look at the refactored approach. I've created a separate service interface and implementation. The `WordprocessingDocument` class implements the interface, delegating to the concrete service implementation. The latter derives from an abstract base class that can be overridden by users to implement their own services if desired (e.g., to create w14:paraId values in a different way).

If you are fine with the approach and architecture, I would complete the implementation by adding the corresponding code required for the `TableRow` class (see the `Paragraph` class for what this means).

Building and testing with `ProjectLoadStyle=All` works on my machine. However, I had to jump through some hoops to make it work for .NET 3.5 and 4.0. If you find the time, I'd be interesting to understand why we are still building the SDK for .NET 3.5, 4.0, and 4.6 (instead of just 4.6.1) and netstandard 1.3 (instead of just netstandard 2.0).

ThomasBarnekow:
@tomjebo and @twsouthwick, I've done more refactoring and added some additional functionality. At this point, the feature looks like this:

- All feature-related classes are in subfolders called `ParaId` in the `DocumentFormat.OpenXml` and `DocumentFormat.OpenXml.Tests` projects.
- The `ParagraphIdHolder` interface is currently implemented by the `Paragraph` class. If we agree that this is the way to go, I'll create a `public partial class TableRow` that implements that interface and overrides the same methods that are also overridden by the `Paragraph` class at the moment. Using that interface, we abstract away the difference between `Paragraph` and `TableRow` instances when dealing with `w14:paraId`-related matters.
- The `IParagraphIdService` defines the operations for the `w14:paraId` service. It is implemented by the abstract class `ParagraphIdService`, which delegates the actual creation of the next `w14:paraId` value to its concrete subclasses. For that task, I've implemented the `RandomParagraphIdService` which uses a strong random number generator to do that. An `IParagraphIdHolder` accepts an `IParagraphIdService` when assigning a unique `w14:paraId` value to a `Paragraph` instance that is not yet associated with a `WordprocessingDocument` by having been added to the DOM tree of a part root element that is associated with an `OpenXmlPart` that in turn is associated with a `WordprocessingDocument`. 
- The `WordprocessingDocument` class implements the `IParagraphIdService` interface and delegates the implementation to the `RandomParagraphIdService`. Thus, one can pass a reference to a `WordprocessingDocument` where an `IParagraphIdService` is required (e.g., to the method `IParagraphIdHolder.SetUniqueParagraphId(IParagraphIdService)`.
- For the more recent frameworks, the `WordprocessingDocument` further implements the following methods:
  - `public IReadOnlyCollection<string> AssignUniqueParagraphIds()` assigns unique `w14:paraId` values if an `IParagraphIdHolder` does not have a `w14:paraId` value or if that value is not unique.
  - `public IReadOnlyCollection<string> RegisterAllParagraphIds()` registers all `w14:paraId` values existing in the `WordprocessingDocument`.
- For .NET Frameworks 3.5 and 4.0, the above methods use the following signature:
  - `public ICollection<string> AssignUniqueParagraphIds()`
  - `public ICollection<string> RegisterAllParagraphIds()`
- To be able to return read-only `ICollection<T>` instances (when targeting .NET Frameworks 3.5 and 4.0), I created the `ReadOnlyCollectionWrapper<T>` class, which wraps `ICollection<T>` instances, makes the `IsReadOnly` property return `true`, and throws `NotSupportedException`s when calling `Add(T)`, `Clear()`, and `Remove(T)`.

The more generic `HexStringFactory` class is still located next to the `HexBinaryValue` class in the `SimpleTypes` folder.

Further, we have the following unit test files:
- `ParagraphTests.ParaId.cs`: Feature-related unit tests for the `Paragraph` class. The test class can be made `partial`.
- `RandomParagraphIdServiceTests.cs`: As the name suggests. Also covers the abstract `ParagraphIdService` class.
- `ReadOnlyCollectionWrapperTests.cs`: As the name suggests.
- `WordprocessingDocumentTests.ParaId.cs`: Feature-related unit tests for the `WordprocessingDocument` class. The test class can be made `partial`.

As a side note, it seems there is no real standardization in the unit tests. The namespaces are sometimes `DocumentFormat.OpenXml.Tests` and sometimes `DocumentFormat.OpenXml.Wordprocessing`, for example. This is why my `WordprocessingDocumentTests` class does not clash with an existing class with the same name, which is in a different namespace, though.

Building and testing with ProjectLoadStyle=All works on my machine.

ThomasBarnekow:
@tomjebo and @twsouthwick, I just rebased the `para-id` branch on `main` and fixed a comment.

src/DocumentFormat.OpenXml/Wordprocessing/Document.cs
twsouthwick: My point is nothing about behavior, but that I'd rather not go around the null reference checks. Maybe the annotations can be updated to something that will tell the compiler that it would be non-null at that point to *fix* this.

As for why it takes a null - it was like that from before I got involved. Potentially a reasonable breaking change at somepoint, though. I'll create an issue to track that.

twsouthwick: See #1009

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -187,13 +188,19 @@ internal void CreateInternal(OpenXmlPackage? openXmlPackage, OpenXmlPart? parent
 
             //OpenXmlPart parentPart = this._ownerPart;
 
-            //Uri is auto generated to make sure it's unique
-            var targetPath = GetTargetPath(openXmlPackage, TargetPath) ?? ".";
+            // Uri is auto generated to make sure it's unique
+            string targetPath = GetTargetPath(openXmlPackage, TargetPath) ?? ".";
+
+            if (targetPath is null)
```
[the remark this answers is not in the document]
twsouthwick: I tend to prefer `var` and let the signature of methods dictate and types I care about. It appears we don't enforce anything at the moment, so feel free to use whatever. We can update that with an analyzers at some point.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -769,31 +775,28 @@ internal void LoadDomTree<T>()
...
 
-                    if (rootElement.LoadFromPart(this, stream))
-                    {
-                        // set this part to the root Element
-                        rootElement.OpenXmlPart = this;
+            try
+            {
+                // set OpenXmlPart before loading from part to be able to access
+                // OpenXmlPart and OpenXmlPackage while loading.
+                var rootElement = new T { OpenXmlPart = this };
```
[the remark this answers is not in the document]
twsouthwick: Hmm interesting. Created #1010 to track further exploring that

src/DocumentFormat.OpenXml/Packaging/WordprocessingDocument.cs
```diff
@@ -236,7 +236,7 @@ public static WordprocessingDocument CreateFromTemplate(string path, bool isTemp
 
             if (extension != ".docx" && extension != ".docm" && extension != ".dotx" && extension != ".dotm")
             {
-                throw new ArgumentException($"Illegal template file: {path}", nameof(path));
+                throw new ArgumentException($@"Illegal template file: {path}", nameof(path));
```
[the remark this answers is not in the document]
twsouthwick: Hmm ideally would be moved to LocalizedStrings.resx

src/DocumentFormat.OpenXml/ParaId/IParagraphIdService.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace DocumentFormat.OpenXml.Wordprocessing
+{
+    /// <summary>
+    /// Defines the operations of the w14:paraId (ParagraphId) service.
+    /// </summary>
+    public interface IParagraphIdService
```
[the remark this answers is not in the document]
twsouthwick: I would like to see this interface be available to handle the other types as well, such as

```csharp
interface IUniqueIdGenerator<TSimpleType> : IEnumerable<string>
{
  int Count { get; }

  bool TryRegisterId(string value);

  bool TryRegisterId(T value);

  string CreateId();
}
```

src/DocumentFormat.OpenXml/ParaId/IParagraphIdHolder.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+
+namespace DocumentFormat.OpenXml.Wordprocessing
+{
+    /// <summary>
+    /// Defines operations for elements holding w14:paraId (ParagraphId) values,
+    /// i.e., <see cref="Paragraph" /> and <see cref="TableRow" /> instances.
+    /// </summary>
+    public interface IParagraphIdHolder
```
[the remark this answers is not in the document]
twsouthwick: I generally am very cautious of exposing public surface changes unless I'm very happy with their shape. Once exposed, they will not be removed without being a breaking change. For me, the bar tends to be fairly high to actually expose something as public.

This feels like it could be made more generic and used elsewhere, but not in its current form. If there's a use case to make this public, we'll need to iterate on the design of it a bit more.

src/DocumentFormat.OpenXml/Wordprocessing/Document.cs
ThomasBarnekow: Thanks, @twsouthwick. You asked exactly the right question in #1009. If it's a breaking change, let's go for it in the next major release.

twsouthwick (commented):
Thanks @ThomasBarnekow for working with us on this PR. Due to the number of public APIs being exposed here, a lot of my feedback is on their shape and the why of exposing them. Things may be able to merged more quickly by making only the required subset of those APIs public as public APIs are forever (unless we consciously want to break people) but we can iterate on design more quickly internally.

src/DocumentFormat.OpenXml/ParaId/Paragraph.ParaId.cs
```diff
@@ -0,0 +1,128 @@
...
+
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        /// <inheritdoc />
+        public string SetUniqueParagraphId()
```
twsouthwick, line 109: Can we change this to a `bool TrySetUniqueId...` pattern rather than throw inside?

src/DocumentFormat.OpenXml/ParaId/Paragraph.ParaId.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (wordprocessingDocument is null)
+            {
+                throw new InvalidOperationException("Paragraph must be added to the DOM tree before calling this method.");
+            }
+
+            return SetUniqueParagraphId(wordprocessingDocument);
+        }
+
+        /// <inheritdoc />
+        public string SetUniqueParagraphId(IParagraphIdService paragraphIdService)
```
twsouthwick, line 121: let's make this internal for now - I've got some ideas I'll play around with how to make it configureable for the end user

src/DocumentFormat.OpenXml/ParaId/IParagraphIdService.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        /// <summary>
+        /// Gets the w14:paraId values that have been registered with the
+        /// <see cref="IParagraphIdService" /> instance.
+        /// </summary>
+#if NET35 || NET40
```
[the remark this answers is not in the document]
ThomasBarnekow: I had thought about a common API. Wanting to use an `IReadOnlyCollection<string>`, I realized that only `net35` and `net40` do not support that interface, and the corresponding interface in the SDK is declared internal (why?).

All frameworks released since 2012 (net45 and later) support the interface. Therefore, I went with `IReadOnlyCollection<string>` for anybody not stuck with `net35`. Those currently stuck with `net35` will anyhow be stuck with that framework forever (and it seems it's supported until 2029, which could be deemed a form of cruelty ;-). .NET Framework 4.0 is no longer supported since 2016, so I don't see why I should bother.

src/DocumentFormat.OpenXml/ParaId/Paragraph.ParaId.cs
ThomasBarnekow: Yes, that would be possible.

ThomasBarnekow:
@twsouthwick, thanks for your suggestions. I could not respond to your points on the `IParagraphIdService` and `IParagraphIdHolder` interfaces directly, so I'll address them here.

Regarding the `IParagraphIdService` interface: 

- Do you want to replace that interface with the `IUniqueIdGenerator<TSimpleType>` interface?
- Do you want to derive from `IEnumerable<string>` to avoid the usage of `IReadOnlyCollection<string>` for returning the registered ids? If so, we'd be doing that because legacy frameworks like .NET Framework 3.5 and 4.0 do not support that interface. The suggested shape somehow doesn't feel natural to me so I'd rather not provide the service for those frameworks than make compromises for everybody else.
- What is the contract for the `CreateId()` method?

Regarding the `IParagraphIdHolder` interface:

- What do you mean with "can be made more generic and used elsewhere"? There are exactly two classes that would implement this interface: `Paragraph` and `TableRow`. The interface only abstracts away differences between those two classes as holders of `w14:paraId` attribute values.
- While I would find it useful to expose the interface, if you have reservations, we can make it internal (at least for now).

src/DocumentFormat.OpenXml/ParaId/Paragraph.ParaId.cs
ThomasBarnekow: This method would be useful in use cases where you would like to provide your own `IParagraphIdService` instance when setting those values. For example, I have a specific use case in which I would like to ensure the uniqueness of `w14:paraId` values across a set of `WordprocessingDocument` instances rather than "just" within an individual `WordprocessingDocument`.

What kind of configuration are you thinking about?

twsouthwick:
For the paragraph Id service:

> Do you want to replace that interface with the IUniqueIdGenerator<TSimpleType> interface?
Yes, this makes it a generalized generator for simple types

> Do you want to derive from IEnumerable<string> to avoid the usage of IReadOnlyCollection<string>
Yes, that is a constraint of the SDK we must work with. I'm fine with the the interface also implementing `IReadOnlyCollection,string>` as it already does (ie IEnumerable<string> and providing a Count), which would "light up" on platforms where it's available

> What is the contract for the CreateId() method?
The same as `CreateUniqueParagraphId();` right now

For the paragraph id service:

Let's just keep it internal for now so we can punt naming/shape discussions for later

src/DocumentFormat.OpenXml/ParaId/Paragraph.ParaId.cs
twsouthwick: I agree with the purpose of it, but I'm not happy with the shape of the API at the moment.

I've been toying with some ideas of services and how a user might want to replace/configure things. ie have a basic IOC container at the root (such as the OpenXmlPartContainer) that exposes `IServiceProvider` so we can provide services without coupling ourselves to specific implementations. This fits in with that goal and would like to explore a more generalized infrastructure. Keeping it internal allows us to play with that without having to decide on that shape right now.

twsouthwick:
@ThomasBarnekow I chatted with Tom and I'm now convinced not to over generalize too early (that's a habit of mine :)). Let's just make as much internal as possible (probably just exposing the method to actually generate the unique idea on the element). Then after we get the equivalent hook up for the other elements that need it, we can see if it makes sense to generalize it.

ThomasBarnekow:
@twsouthwick, I made changes based on your feedback:

- `IParagraphIdHolder` is now internal. However, `Paragraph` (and in the next step `TableRow`) still implements all `IParagraphIdHolder` methods, including `SetUniqueParagraphId(IParagraphIdGenerator)`. The latter is required to set unique `w14:paraId` values for `Paragraph` instances that are not yet added to the DOM tree. If you wanted to make that method internal, we'd have to make further changes.
- `IParagraphIdService` renamed to `IParagraphIdGenerator`. The `RegisterParagraphId()` methods are now internal.
- `WordprocessingDocument` no longer implements `IParagraphIdGenerator` interface (previously `IParagraphIdService`). 
 
I'm not happy with the shape of this feature right now. It is less elegant to use than it was before. So please have a look and let's talk about how we can improve this. I've already thought about implementing a completely separate service that does not change the interface of the SDK classes at all.

twsouthwick:
Great! I agree the shape is weird and we can figure out what that should look like. I think it will be more clear when the functionality gets added to the other classes. 

As for the interface implementation you can do it explicitly to keep the methods from being public.

I'm heading out for a camping trip but will review when I can.

ThomasBarnekow:
@twsouthwick, I've made more changes to essentially avoid all SDK API changes. Now, the only change to the existing SDK classes is that the `Paragraph` and `TableRow` class implement an internal interface `IIdentifiableByParaId` (previously `IParagraphIdHolder`).

A user would now construct and use a `RandomParagraphIdGenerator` instance to generate unique `w14:paraId` values and assign those to the `ParagraphId` properties of the `Paragraph` and `TableRow` instances. Have a look at the unit tests to see usage scenarios.

On the one hand, this is less integrated and has less internal magic (e.g., does not register `w14:paraId` values as DOM trees are loaded). On the other hand, it does not change the public API at all. I like this much better than the previous "compromise".

twsouthwick (approved):
LGTM. Let's get this in and continue iterating on the public APIs to support. Can you update the original issue with the missing features you feel aren't in yet?

PR #1000, opened 2021-07-28
Add DocumentFormat.OpenXml.Linq project

ThomasBarnekow:
As discussed in #952 (Provide direct support for Linq to XML), this PR adds the `DocumentFormat.OpenXml.Linq` project with code generated by the `DocumentFormat.OpenXml.CodeGeneration.Linq` project. It enables code generation by also enhancing the `ElementMetadata` class.

I made an effort to implement the `DocumentFormat.OpenXml.CodeGeneration.Linq` project as a C# Source Generator. However, it turned out to be either impossible or at least too complex (and therefore not worth the effort). A C# Sourcing Generator **must** target `netstandard2.0` and cannot depend on other projects in the same solution unless those are also analyzers. In particular, a dependency on the `DocumentFormat.OpenXml` project is not possible. However, that is required to generate the namespace-related classes for Linq to XML.

We'll have to discuss the following questions:

- How should we integrate the code generation performed by `DocumentFormat.OpenXml.CodeGeneration.Linq` into the build process? It is a console project that currently targets `net5.0` and can be invoked easily.
- Which frameworks should `DocumentFormat.OpenXml.Linq` target? In this initial commit, it only targets `netstandard2.0`. Should we add `net461` like other Microsoft NuGet packages targeting `netstandard2.0` (which do not support any .NET framework older than 4.6.1)? Or do we have to support all the legacy frameworks that are also targeted by `DocumentFormat.OpenXml`?


twsouthwick:
Thanks for separating this out! I'm fine not having it a source generator for now (we can play around with that after getting the main structure in.

As for what to target, best to target the same as the samples for now probably

ThomasBarnekow:
> As for what to target, best to target the same as the samples for now probably

What exactly do you mean by "samples"? I would see the following options:

1. `netstandard2.0`, if we only wanted to support recent .NET frameworks (`net472`, `net48`), `netstandard2.0` and later, `netcoreapp2.1` and later, and `net5.0` and later, for example; 
2. `netstandard2.0;net461`, to improve the developer experience for .NET frameworks 4.6.1 to 4.7.1; or
3. whatever `'$(ProjectLoadStyle)' == 'All'` would target as the `ProductTargetFrameworks` (which does not include `net5.0` although it is an individual target).

ThomasBarnekow:
@twsouthwick, could you tell me why the build failed? And do you have any further feedback?

ThomasBarnekow:
@twsouthwick, it seems I had not pulled the most recent commits from upstream when creating the branch from main. Therefore, the reason for the failed check was a failed unit test that was unrelated to this PR.

 I pulled from upstream and rebased my PR branch. At least on my local machine, all unit tests now pass. So please run the check again.

ThomasBarnekow:
@twsouthwick, why did this check fail?

ThomasBarnekow:
Just to be sure, I checked again, running the following on my local machine for both the Debug and Release configurations:

```
set ProjectLoadStyle=All
dotnet build --configuration [Debug|Release]
dotnet test --configuration [Debug|Release]
```

For both Debug and Release, it builds without any warnings or errors and all unit tests pass.

Is there any other check performed on the build server beyond what we can check on our machines? Would it be possible to get access to the build server outputs at least? Not knowing why stuff fails is a bit unsatisfying.

ThomasBarnekow:
@twsouthwick, do you have any feedback on my questions above? Firstly:

> What exactly do you mean by "samples"? I would see the following options:
> 
> 1. `netstandard2.0`, if we only wanted to support recent .NET frameworks (`net472`, `net48`), `netstandard2.0` and later, `netcoreapp2.1` and later, and `net5.0` and later, for example;
> 2. `netstandard2.0;net461`, to improve the developer experience for .NET frameworks 4.6.1 to 4.7.1; or
> 3. whatever `'$(ProjectLoadStyle)' == 'All'` would target as the `ProductTargetFrameworks` (which does not include `net5.0` although it is an individual target).

Secondly, why did the check fail? 

> Just to be sure, I checked again, running the following on my local machine for both the Debug and Release configurations:
> 
> ```
> set ProjectLoadStyle=All
> dotnet build --configuration [Debug|Release]
> dotnet test --configuration [Debug|Release]
> ```
> 
> For both Debug and Release, it builds without any warnings or errors and all unit tests pass.
> 
> Is there any other check performed on the build server beyond what we can check on our machines? Would it be possible to get access to the build server outputs at least? Not knowing why stuff fails is a bit unsatisfying.

src/DocumentFormat.OpenXml.CodeGeneration.Linq/Program.cs
```diff
@@ -0,0 +1,536 @@
...
+using System.IO;
+using System.Linq;
+
+namespace DocumentFormat.OpenXml.CodeGeneration.Linq
+{
+    public static class Program
+    {
+        private const string EmptyNamespace = "NoNamespace";
+
+        public static void Main()
```
lindexi, line 18: Is it a demo code?

ThomasBarnekow: No, this is not meant to be demo code. This is a code generator that can be invoked on the command line or during the build process. It generates (most of) the classes in the `DocumentFormat.OpenXml.Linq` project.

lindexi: @ThomasBarnekow But I hard to find the directory: `const string directoryName = "..\\..\\..\\..\\src\\DocumentFormat.OpenXml.Linq\\GeneratedCode";`

ThomasBarnekow: Do you mean that it is hard to find?

This might be a temporary way to specify the output path, which is the `GeneratedCode` folder within the `DocumentFormat.OpenXml.Linq` project. We need to first decide how the code generation is invoked and then determine how the path must be specified. But that would require some feedback from @twsouthwick.

src/DocumentFormat.OpenXml.Linq/DocumentFormat.OpenXml.Linq.csproj
```diff
@@ -0,0 +1,51 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), DocumentFormat.OpenXml.Package.props))\DocumentFormat.OpenXml.Package.props" />
+
+  <PropertyGroup>
+    <Nullable>enable</Nullable>
+    <IsShipped>true</IsShipped>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
twsouthwick, line 8: For now, set this to `$(ProductTargetFrameworks)`. The build ends up building a matrix of various permutations and sometimes the output is just .NET 3.5 or .NET 4.0, which would not be consumable here

twsouthwick:
Sorry for the slow responses - between some busy projects and vacations, I've had little time to review here.

This is shaping up fine and once we get the 2.13.1 release out (hopefully today), we can then get these merged in.

Let's not worry about the code generation at this point and leave what you have in. After that, I have some ideas as to how to make the code generators work, but what you're doing now is fine.

ThomasBarnekow:
Thanks, Taylor. I've added a commit that changes the .csproj file so that we are targeting the `$(ProductTargetFrameworks)`. I've built and tested on my machine with `ProjectLoadStyle=All`.

src/DocumentFormat.OpenXml.CodeGeneration.Linq/Program.cs
twsouthwick: Yeah, we'll be playing around with this and hopefully be able to use a source generator soon. This is one of the big problems that source generators aim to solve, and I know who to bug if we have trouble getting them to work :)

src/DocumentFormat.OpenXml/Framework/Metadata/ElementMetadata.cs
```diff
@@ -39,10 +43,13 @@ internal class ElementMetadata
             _children = lookup;
         }
 
-        private ElementMetadata()
+        private ElementMetadata(Type type)
         {
+            Type = type;
         }
 
+        public Type Type { get; }
```
twsouthwick, line 51: What's the purpose of this? It's probably fine, but does add some size to the data structure which we create many, many, many of so I'm interested if there's an additional way to do it.

src/DocumentFormat.OpenXml/Properties/Properties.cs
```diff
@@ -18,6 +18,9 @@
...
 [assembly: InternalsVisibleTo("DocumentFormat.OpenXml.CodeGeneration.Schematron, Publickey=002400000480000094000000060200000024000052534131000400000100010061d8931836c82bf25ca6b773dfd6e7b3ab4e43fba60cf4a86347170373415a165ccc40da3da4a52163822db9fa91f15828236d32d6a9fe754859f10d1f8262646c1f3fb6b4348123f14d733db0ff11c3198b7cf56caaebbf14563990446a6c32aff36d5a7097194294c127fe3cdf9f2609daae5f4daf26f8b6227f203d2a8bbf")]
 
+// For Linq namespace class genertion
+[assembly: InternalsVisibleTo("DocumentFormat.OpenXml.CodeGeneration.Linq, Publickey=002400000480000094000000060200000024000052534131000400000100010061d8931836c82bf25ca6b773dfd6e7b3ab4e43fba60cf4a86347170373415a165ccc40da3da4a52163822db9fa91f15828236d32d6a9fe754859f10d1f8262646c1f3fb6b4348123f14d733db0ff11c3198b7cf56caaebbf14563990446a6c32aff36d5a7097194294c127fe3cdf9f2609daae5f4daf26f8b6227f203d2a8bbf")]
```
twsouthwick, line 22: I would love to see some of the internal types start to be made public, so we can stop needing so many internalsvisibleto. If the shape of the APIs being used are working well, we can look at removing that for this project at least.

twsouthwick (approved):

src/DocumentFormat.OpenXml.CodeGeneration.Linq/Program.cs
```diff
@@ -0,0 +1,536 @@
...
+        {
+            var visitedElementTypes = new HashSet<Type>();
+            var elementMetadataCollection = new List<ElementMetadata>();
+
+            foreach (ElementLookup.ElementChild elementChild in ElementLookup.Parts.Elements)
+            {
+                AssembleElementMetatata(ElementMetadata.None, elementChild, visitedElementTypes, elementMetadataCollection, fieldInfos);
+            }
+
+            return elementMetadataCollection;
```
twsouthwick, line 293: >  Error CS0266: Cannot implicitly convert type 'System.Collections.Generic.List<DocumentFormat.OpenXml.Framework.Metadata.ElementMetadata>' to 'DocumentFormat.OpenXml.IReadOnlyCollection<DocumentFormat.OpenXml.Framework.Metadata.ElementMetadata>'

This breaks on .net 3.5/.net 4.0

ThomasBarnekow: It does not have to work on .net 3.5/.net 4.0, does it? This is just a build tool that generates code and can thus target any appropriate .net framework or .net 5.0. You would only run this **once** to generate the namespace-related classes.

twsouthwick: The way the CI is currently set up is to cross-compile various permutations in order to get some test coverage on the .net 3.5/.net 4.0 builds. It needs to build for those targets at the moment, but you can do a compile remove or something so that those builds are no-op.

With VS 2022 supporting larger projects, I'd like to see a simplification of the build process....

ThomasBarnekow: I've changed the project to target the `$(SamplesFrameworks)`, i.e., `net46`, `netcoreapp2.1`, `netcoreapp3.1`, and `net5.0`. I would hope that this is enough.

ThomasBarnekow:
@twsouthwick, please run the checks to see whether it works. It builds successfully on my machine with `set ProjectLoadStyle=All`. As I said above, it targets whatever is in $(SamplesFrameworks), i.e., `net46`, `netcoreapp2.1`, `netcoreapp3.1`, and `net5.0` at the moment.

I've noted that the unit tests target `net452` and all of the above. However, while the product does not target `net452`, it targets `net35` and `net40` on top of `net46` (plus `netstandard1.3` and `netstandard2.0`). Can you explain why `net35´ and `net40` are required? Why can't the SDK not simply target `netstandard2.0` and `net461` like several other Microsoft libraries? Another observation is that it targets `net5.0` by default, which is however not included in the `$(ProductTargetFrameworks)` when `ProjectLoadStyle` is set to `All`.

src/DocumentFormat.OpenXml/Framework/Metadata/ElementMetadata.cs
ThomasBarnekow: The purpose of this is to generate XML documentation in the following form (using the `w:abstractNumId` element and attribute as an example:

```csharp
/// <summary>
/// Represents the w:abstractNumId XML element and attribute.
/// </summary>
/// <remarks>
/// <para>As an XML element, it:</para>
/// <list type="bullet">
/// <item><description>has the following parent XML elements: <see cref="num" />.</description></item>
/// <item><description>has the following XML attributes: <see cref="val" />.</description></item>
/// <item><description>corresponds to the following strongly-typed classes: AbstractNumId.</description></item>
/// </list>
/// <para>As an XML attribute, it:</para>
/// <list type="bullet">
/// <item><description>is contained in the following XML elements: <see cref="abstractNum" />.</description></item>
/// <item><description>corresponds to the following strongly-typed properties: AbstractNum.AbstractNumberId.</description></item>
/// </list>
/// </remarks>
public static readonly XName abstractNumId = w + "abstractNumId";
```

In the above XML documentation, you'll see

- "corresponds to the following strongly-typed classes: **AbstractNumId**" and
- "corresponds to the following strongly-typed properties: **AbstractNum**.AbstractNumberId"

The class names (`AbstractNumId` and `AbstractNum`) are retrieved from the `Type` property. While the corresponding property exists in the `AttributeMetadata` class, it did not exist in the `ElementMetadata` class, so I had to add it.

src/DocumentFormat.OpenXml/Properties/Properties.cs
ThomasBarnekow: Agree. I only had to add this because the relevant types are internal at the moment.

twsouthwick:
I'll try to answer all the questions:

- A number of people use .NET 3.5 and .NET 4.0 for things like sharepoint integrations. Since these are still supported, I keep the targets around
- A common misconception is that it's unsupported to compile an application targeting .NET 4.0 and run it on .NET 4.8. That's supported and does potentially opt you into various quirks that are required by some systems
- .NET 5.0 target is there so we can start experimenting with things like Span<> but haven't decided to pull the trigger with that TFM yet

twsouthwick:
I pushed a change that should work - it will only compile for .NET 5.0, but on the other sample projects, it will just output a message saying it is unsupported on that target.

Long term we can hopefully streamline this, but the build matrix on the server is done so that we can run tests against the .NET 3.5 builds.

PR #1013, opened 2021-08-31
More public XmlPath members

rmboggs:
This makes the constructors and a couple of static methods in the XmlPath class public.


twsouthwick:
Do we need both the constructor and the factory?

rmboggs:
> Do we need both the constructor and the factory?

Honestly, I think the constructors would be sufficient.  I only did that because @ThomasBarnekow suggested it.  If you need me to revert the factory, let me know.

twsouthwick:
Let's just do the constructors

rmboggs:
> Let's just do the constructors

Fair enough, let me revert the non-constructor related changes

rmboggs:
That should do it.

src/DocumentFormat.OpenXml/XmlPath.cs
```diff
@@ -21,9 +21,13 @@ namespace DocumentFormat.OpenXml
     public class XmlPath
     {
         /// <summary>
-        /// Initializes a new instance of the XmlPath.
+        /// Initializes a new instance of the XmlPath from the specified
```
twsouthwick, line 24: can you use `<see href=""/>` annotations for types here?

src/DocumentFormat.OpenXml/XmlPath.cs
```diff
@@ -54,7 +58,7 @@ internal XmlPath(OpenXmlElement element)
         /// Initializes a new instance of the XmlPath from the specified OpenXmlPart.
```
twsouthwick, line 58: same as above

src/DocumentFormat.OpenXml/XmlPath.cs
rmboggs: You want me to use `<see href=""/>` here or `<see cref=""/>`?  If the former, what url should I be using for the classes?

twsouthwick: haha yeah cref

rmboggs: Cool, just making sure.  I'll work on that.

rmboggs: Should be updated as requested now.

twsouthwick (approved):

rmboggs:
Hi,

Since this is approved, do you know when it can be merged into Main?

twsouthwick:
Done

PR #1021, opened 2021-09-10
Move XLinq related functionality to DocumentFormat.OpenXml.Linq assembly

twsouthwick:
This creates the XLinq related functionality as a feature that can be opted into if desired.


twsouthwick:
@ThomasBarnekow This preserves the behavior all the tests were covering. I'm not sure what the annotations were and nothing was checking that, so let me know what you think.

test/DocumentFormat.OpenXml.Linq.Tests/OpenXmlPartTests.cs
```diff
@@ -36,10 +36,13 @@ public void RootXElement_Get_NewPart_NullAndSynchronized()
             // Arrange.
             using var stream = new MemoryStream();
             using var wordDocument = WordprocessingDocument.Create(stream, WordprocessingDocumentType.Document);
+
+            wordDocument.TrySetPartRootXElementFeature();
```
ThomasBarnekow, line 40: @twsouthwick, will this call be required before the feature can be used? Or can this be avoided?

twsouthwick: Right now, it's required because it needs to register to events which need to be unhooked and require disposal. The feature to provide disposal right now requires to be done at the package level

ThomasBarnekow:
@twsouthwick, unfortunately, this changes the public API and breaks the code that I have written against that API. But I'm obviously only one user.

I am more worried about this feature suddenly no longer being a "first-class citizen", as the `OpenXmlPart` support for strongly-typed classes. In particular, having to opt-in by calling the `TrySetPartRootXElementFeature()` method makes it less elegant than the API offered by the Open XML PowerTools already. Prior to this PR, the `OpenXmlPart` API for the Linq-to-XML feature was similar to the corresponding API for the strongly-typed classes.

There is no cost associated with the addition of the feature to the `OpenXmlPart`. Therefore, I don't see why this change is necessary and why users would have to opt-in. Thus, I would kindly ask you to not make this change.

test/DocumentFormat.OpenXml.Linq.Tests/OpenXmlPartTests.cs
ThomasBarnekow: Unfortunately, having to call this method explicitly would render the API pretty much unusable. For example, when passing a `WordprocessingDocument` to multiple methods that transform the `WordprocessingDocument`, each method would have to ensure the feature is enabled. Countless calls would have to be added to the codebase.

twsouthwick:
Thanks Thomas for the feedback. 

> I am more worried about this feature suddenly no longer being a "first-class citizen"

I believe we have different understandings of what "first-class citizen" means.  My view is that “first-class citizen” would mean that the SDK allows for composing functionality in a way so a developer may customize their needs. That’s the aim of the “Features” functionality I merged in. My plan is, as time goes on, more of the inner workings of the SDK will become “Features” so that they can be more easily tested/maintained/customized/etc. The XLinq feature seemed like a logical first step as it builds on top of the OpenXmlElement and doesn’t have any dependencies we have to worry about (remember, backwards compatibility has always been a key issue and so changes to the core library must meet a high bar).

> having to opt-in by calling the TrySetPartRootXElementFeature()

This is up for design of course, like any functionality. The challenge (as I pointed out) is that in order to set this up, we need to be able to dispose the events we hook up to listen for changes to a document.  Happy to explore other options here.

I imagine people using this by creating their own extension method that would add any features they are interested in enabling when they first create a document so it would end up being available if added there. It may be that for this pattern to work, we should have some samples of "best practices" instead of what tests do (which are not always best practice).

> There is no cost associated with the addition of the feature to the OpenXmlPart.

Of course there is a cost. Everything has a cost, either in development, test, performance, maintainability, etc. I see a non-trivial cost for maintainability with the current implementation as `OpenXmlElement` already feels like spaghetti code.

test/DocumentFormat.OpenXml.Linq.Tests/OpenXmlPartTests.cs
twsouthwick: Found a workaround. This is no longer needed (it will now verify it has the feature, and if not add it to the parent).

twsouthwick:
FYI The requirement to initialize package wide has been removed and now it automatically registers necessary features on demand if not already registered.

ThomasBarnekow:
@twsouthwick, that sounds good! We are now only talking about an API change, e.g., from 
```csharp
public XElement? RootXElement { get; set; }
```
to
```csharp
public static XElement? GetXElement(this OpenXmlPart part);
public static void SetXElement(this OpenXmlPart part, XElement? xElement);
```
I'm generally fine with that. So let's talk about optimizing the design. 

Assigning `null` to `RootXElement` threw an `ArgumentNullException`, so the `SetXElement()` method should take an `XElement` rather than `XElement?`. Or do we want to allow `null`, meaning that an `XDocument` without a root element would have to be written to the part?

Next, I was thinking about whether we'd need to make changes to the API to accommodate Open XML PowerTools users. For example, the `PtOpenXmlExtensions` class defines the following extension methods (and others):

```csharp
public static XDocument GetXDocument(this OpenXmlPart part);
public static void PutXDocument(this OpenXmlPart part);
public static void PutXDocument(this OpenXmlPart part, XDocument document);
```
I have created my own, corresponding extensions that work with an `XElement` rather than an `XDocument`. Our methods also work with `XElement` instances, which is good and more practical than `XDocument` instances.

I'm now thinking that the fact that our API is different from the Open XML PowerTools extensions is actually a good thing. Once the new release is published, I would create a PR for the Open XML PowerTools, which changes the above-mentioned extension methods so that they will use our API and, thus, benefit from the feature (e.g., the synchronization of the strongly-typed and Linq-to-XML classes). This could even be done as a non-breaking change. Thus, I think I'm happy with the API as it stands (noting that we need to talk about the nullability question above).

Remembering your question regarding the annotations, the Open XML PowerTools methods I listed above leverage annotations of the `OpenXmlPart` instances. For example, `GetXDocument()` stores the `XDocument` instance as an annotation. Calling the method again returns the annotation rather than reading from the part. `part.PutXDocument()` writes the annotation to the part. Removing those annotations was a safety measure that would no longer be necessary if the methods are changed to use our feature. Annotations will then no longer be necessary (to be confirmed).

src/DocumentFormat.OpenXml.Linq/OpenXmlLinqExtensions.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <param name="part">The part to get the contents of.</param>
+        /// <returns>An <see cref="XElement"/>.</returns>
+        public static XElement? GetXElement(this OpenXmlPart part)
+            => part.GetPartRootXElementFeature().Root;
+
+        /// <summary>
+        /// Sets an <see cref="XElement"/> to the part. This will force a save to the underlying stream.
+        /// </summary>
+        /// <param name="part">The part to set to.</param>
+        /// <param name="xElement">The element to set.</param>
```
ThomasBarnekow, line 54:
```suggestion
        /// <param name="element">The <see cref="XElement"/>.</param>
```

src/DocumentFormat.OpenXml.Linq/OpenXmlLinqExtensions.cs
```diff
@@ -0,0 +1,169 @@
...
+        public static void SetXElement(this OpenXmlPart part, XElement? xElement)
+            => part.GetPartRootXElementFeature().Root = xElement;
```
ThomasBarnekow, line 56:
```suggestion
            => part.GetPartRootXElementFeature().Root = element;
```

src/DocumentFormat.OpenXml.Linq/OpenXmlLinqExtensions.cs
```diff
@@ -0,0 +1,169 @@
...
+            /// <summary>
+            /// Gets or sets the root LINQ to XML element.
+            /// </summary>
+            public XElement? Root
+            {
+                get
+                {
+                    if (_rootXElement is null)
+                    {
+                        if (_part.RootElement is null)
```
ThomasBarnekow, line 97:
```suggestion
                        if (!_part.IsRootElementLoaded)
```

@twsouthwick, see my longer comment below.

It would be better to invert the boolean expression and swap the `if` and `else` blocks below.

src/DocumentFormat.OpenXml.Linq/OpenXmlLinqExtensions.cs
```diff
@@ -0,0 +1,169 @@
...
+            private readonly IPartRootEventsFeature _events;
+
+            private XElement? _rootXElement;
+
+            public RootXElementFeature(OpenXmlPart part)
+            {
+                _part = part;
+                _events = _part.Features.GetRequired<IPartRootEventsFeature>();
+
+                _events.Change += OnPartRootChange;
```
ThomasBarnekow, line 77: @twsouthwick, I'm not suggesting a change here but wanted to note that only `Saving` and `Saved` events are raised by the `OpenXmlPartRootElement`. However, I'm thinking we might also want to have `OpenXmlPartRootElement` raise the `Reloading` and `Reloaded` events that you have already defined. For example, consider the case where the `_rootXElement` was deserialized from an `OpenXmlPartRootElement` instance that was changed after loading it from the `OpenXmlPart`. If the user discards those changes by calling the `Reload()` method, the `_rootXElement` should also be discarded and loaded again once it is accessed.

ThomasBarnekow (changes requested):
@twsouthwick, I've made a couple of suggestions for improvements.

twsouthwick:
Thanks for the feedback! One thought I had about the API is to not have the `Set` and `Get...` methods but just return the feature so the methods could be called directly there and accessed via the property. i.e.:

```csharp
public static IXElementFeature AsXElement(this OpenXmlPart);

public static Test()
{
  part.AsXElement().Root = ....
}
```

Thoughts?

src/DocumentFormat.OpenXml.Linq/OpenXmlLinqExtensions.cs
twsouthwick: I've created an issue to track that for a future change.

twsouthwick:
Good call on the nullability - I've updated that and addressed your feedback.

twsouthwick:
As far as XDocument/XElement support - if we want to support XDocument, then I'd rather just expose that and have people operate on `XDocument.Root` if they want the XElement. I'm fine with either (the document doesn't really add much, so I'll leave it up to you since you've been using this pattern)

twsouthwick:
> Would there be a way to combine the best of two worlds?

One goal I'm pursuing with the SDK is to make the size more minimal (it's currently 5kb+). The LINQ functionality (with the static class that are there right now) would add 291KB. It makes sense to me to keep it all together as one deliverable unit people can opt into fairly easily by just installing it and using the extension methods but not burden others with the extra size if they won't ever use it.

Let's just keep the `Get...` /`Set...` pattern and people can create their own extension methods if there would prefer a `ToXElement()` call.

twsouthwick:
btw - sounds like you'd be a fan of the "Extension Everything" goal of C# especially the extension properties. If only it were available!

ThomasBarnekow:
> As far as XDocument/XElement support - if we want to support XDocument, then I'd rather just expose that and have people operate on `XDocument.Root` if they want the XElement. I'm fine with either (the document doesn't really add much, so I'll leave it up to you since you've been using this pattern)

Let's not go down that `XDocument` route! That's not very usable (which is why I added my own extension methods that take or return an `XElement`). You don't need an `XDocument` but an `XElement`. It is possible to change the Open XML PowerTools extension methods to work with the `XElement` returned by our API and simply wrap it in an `XDocument` for compatibility.

I just mentioned that `XDocument`-based API to make my thinking transparent and noted that I like the `XElement`-based API.

ThomasBarnekow:
> > Would there be a way to combine the best of two worlds?
> 
> One goal I'm pursuing with the SDK is to make the size more minimal (it's currently 5kb+). The LINQ functionality (with the static class that are there right now) would add 291KB. It makes sense to me to keep it all together as one deliverable unit people can opt into fairly easily by just installing it and using the extension methods but not burden others with the extra size if they won't ever use it.
> 
> Let's just keep the `Get...` /`Set...` pattern and people can create their own extension methods if there would prefer a `ToXElement()` call.

I'm not talking about the full feature with all the generated namespace-related classes (e.g., `W`, `W10`, `W14`). I'm just talking about the `IPartRootXElementFeature` interface (26 lines), the `OpenXmlLinqExtensions` class (41 lines, since we'd not need the `Get`, `Set`, and `Save` extension methods), and the `RootXElementFeature` class (118 lines), and the following 8 lines in the `OpenXmlPart` class:

```csharp
public XElement? RootXElement
{
    get => this.GetPartRootXElementFeature().Root;
    set => this.GetPartRootXElementFeature().Root = value;
}

public void SaveRootXElement()
    => this.GetPartRootXElementFeature().Save();
```

We'd be talking about less than 200 lines in total (incl. comments and blank lines). That should not have any noteworthy impact on the size of the assembly.

ThomasBarnekow:
> btw - sounds like you'd be a fan of the "Extension Everything" goal of C# especially the extension properties. If only it were available!

We'd not be talking about my comment above if we had extension properties ;-)

twsouthwick:
For now, I'd like to keep the "LINQ" related behavior together. If it proves to be super common and installing the LINQ package with the method calls instead of properties is too problematic, we can explore moving it into the main SDK.

ThomasBarnekow:
> For now, I'd like to keep the "LINQ" related behavior together. If it proves to be super common and installing the LINQ package with the method calls instead of properties is too problematic, we can explore moving it into the main SDK.

As mentioned above, the "LINQ" related behavior consists of less than 200 lines of code. There is no dependency (in either direction) between that code and the generated classes, which don't provide any actual behavior (just `public static readonly` fields). While those classes are useful and helpful, you don't need them to make use of the LINQ-related behavior that could be provided very elegantly (and with less than 10 lines of code) by the `OpenXmlPart`. By keeping the feature and the generated classes together in a separate assembly, you are demanding that developers take a dependency on another assembly, the size of which is driven by generated classes they might not even need.

I buy the argument that you don't want to increase the size of the assembly by 291 KB (from ca. 5,000 KB)., which is why I agreed to create the separate assembly. If the size is a problem, we should not be asking developers to add 291 KB to their codebase unless they explicitly need what drives that size, i.e., the generated classes.

Wouldn't you agree that an API with a `RootXElement` property (the counterpart of the strongly-typed `RootElement`) and the `SaveRootXElement()` method, that could be provided by the `OpenXmlPart`, is better than the extension methods (which are not bad but not quite as elegant)?

@tomjebo, I'd be interested in your thoughts as well.

twsouthwick:
Another reason (besides viewing the XLinq as a distributed feature) is that it relies on the `IDisposableFeature`, which is also in a separate assembly. That assembly is intended as a collection of additional "quality of life" improvements that may help but are not central to the file-format abstraction of the core SDK.

ThomasBarnekow:
@twsouthwick, I believe you are going too far. You have created a separate assembly for an `IDisposableFeature` implemented in less than 200 lines of code. And this is a reason for not including less than 200 lines of code for the LINQ-related feature that I always wanted to be part of the core DocumentFormat.OpenXml library, which is why I created my original PR (based on @tomjebo 's request) in the first place.

tomjebo:
@ThomasBarnekow I've been monitoring this conversation as I saw it as design discussion. However, I would like to interject just a little in support of separate assemblies. Stepping back and looking at the paraId requirement for uniqueness, the question of whether we should include such code in the SDK was always somewhat grey for me. When you proposed to provide this, I was ok with investigating but how to do it was in question so I requested that you submit a PR to demonstrate your concept and let us iterate a bit. After all this discussion, I'm still ok with providing the code but as soon as we started talking about "opt-in", which I now believe to be important as other consumers of the SDK *may* want to use different schemes for paraId content and this gives them the freedom to do just that, it became much more apparent to me that this should be an add-on feature.  One reason is that there are other things we talk about in our extension and implementation note specifications that could be implemented for added Office value and yet may not be desirable in the SDK's base or non-extended functionality to other consumers.  For that reason, I think Taylor's suggestion, whether perfect or needing refinement, is wise in principle. I would like to continue to explore and refine that approach but would be open to listen to any real problems you see with that approach.

ThomasBarnekow:
@tomjebo, thanks for chiming in. Note that this discussion is not about the `w14:paraId` feature, which can be opt-in. This is about providing direct support for Linq to XML in the Open XML SDK as discussed during the Microsoft DevDays Redmond 2021 Online (see PR #952). After my talk on the DevDays, you asked me to open a PR for this.

"Direct support" did not mean "separate library". However, I understood @twsouthwick's request to create a separate library (`DocumentFormat.OpenXml.Linq`) for the generated namespace-related classes, because those would add ca. 290 KB (ca. 5-6%) to the ca. 5,000 KB `DocumentFormat.OpenXml` library. The current discussion is about adding ca. 200 lines of code to the `DocumentFormat.OpenXml` library (which are currently also in the `DocumentFormat.OpenXml.Linq` library), including the following public API of `OpenXmlPart`, which is what got merged already (with a different implementation, which was replaced with the `Feature´ approach, which is an improvement in terms of maintainability):

```csharp
public XElement? RootXElement
{
    get => this.GetPartRootXElementFeature().Root;
    set => this.GetPartRootXElementFeature().Root = value;
}

public void SaveRootXElement()
    => this.GetPartRootXElementFeature().Save();
```

I would like to avoid having to use a separate library that would then have to use extension methods to provide essentially the same feature. The reason is that not everybody who would want to use the API above would also need the generated namespace-related classes (e.g., `W`, `W10`, `W14`). An example would be all the users of the Open XML PowerTools, which already provide their version of the namespace-related classes (not as complete and without any documentation). It would be a minor, non-breaking change of the Open XML PowerTools to use the above API (implementations of two extension methods). It would be a breaking change to use the namespace-related classes provided by the `DocumentFormat.OpenXml.Linq` library.

In other words, the API above, with a different implementation, has been added to the `OpenXmlPart` by merging PR #952. This current PR is about removing that API and its implementation again and replacing it with extension methods and a feature implementation provided by a separate library (`DocumentFormat.OpenXml.Linq`). I would like to have this small API and the feature implementation as a core feature of the `DocumentFormat.OpenXml` library. I am fine with having the generated namespace-related classes in the separate `DocumentFormat.OpenXml.Linq` library. Those classes can be opt-in.

twsouthwick:
@ThomasBarnekow I feel like this conversation is becoming unproductive by going in circles. I understand you are not fully on board with the direction this PR is taking this feature, but there are reasons behind it that I have enumerated many times above. To reiterate, the main driving force behind moving things to features is to decouple things from the core SDK project to make it smaller, not to add more features into it.  And yes, this is in direct conflict with your desire for the feature to be in the core library.

I'm sorry that you had misunderstood what "direct support" meant, but the feature is still being provided albeit in a different assembly. There is a large discussion we are having about what the scope of this project is and should be as we move forward and these changes are some of the main driving forces there. We see the value of being able to build things like what you have provided, but historically there would have been a hard "no" to a feature like this at all as it is not directly involved with accessing the parts/elements/etc. As @tomjebo mentioned, this and the paraid feature has got us really thinking about what this should look like - how to meet our goals as well as enable the community to have access to better SDK. The goal is to provide foundations where things like this are easily built on top of the core assembly (which is what is being done).  I understand that for you this may seem like a distinction without a difference, but for now, we're going to move forward with this. We may be wrong, but we're trying to balance a number of factors here.

Remember: once something is in the SDK (and released in a stable release which this feature has not), it's there for good, but these new libraries we can play around with. We can measure how many people use them and then potentially pull things in that see high usage. We can break things (as we will most likely not release them as stable for a bit to get wider feedback on) and make sure it works well for the wider audience. These ideas that we requested you provide PRs for are potentially very powerful and this doesn't change that fact. We appreciate the effort you put in to provide them. However, we are the ones responsible for maintaining them and ensuring they work well all together and with the future of the project.

tomjebo:
@ThomasBarnekow I was aware this thread is about Linq but was exposing my thought process which for me initiated with the paraId feature.  I also acknowledge that perhaps we didn't focus on the meaning of "direct support", but to me it is no less direct having it offered as a feature by the SDK.

tomjebo (approved):

ThomasBarnekow:
There is (hopefully) one last issue to be addressed if you put the `OpenXmlLinqExtensions` class and the namespace-related classes (e.g., `W`, `W10`, `W14`) in the same assembly and, as it stands at the moment, same namespace. Consider the following simplified example:

### Example

First, the `DocumentFormat.OpenXml.Linq` namespace contains the extension methods provided by the feature:

```csharp
// DocumentFormat.OpenXml.Linq: OpenXmlLinqExtensions.cs

namespace DocumentFormat.OpenXml.Linq
{
    public static class OpenXmlLinqExtensions
    {
        public static XElement? GetXElement(this OpenXmlPart part)
            => part.GetPartRootXElementFeature().Root;

        public static void SetXElement(this OpenXmlPart part, XElement element)
            => part.GetPartRootXElementFeature().Root = element;
    }
}
```

Second, the `DocumentFormat.OpenXml.Linq` namespace contains the generated namespace-related classes (I've removed all comments to keep it short):

```csharp
// DocumentFormat.OpenXml.Linq: W.g.cs

namespace DocumentFormat.OpenXml.Linq
{
    public static class W
    {
        public static readonly XNamespace w = "http://schemas.openxmlformats.org/wordprocessingml/2006/main";
        public static readonly XName abstractNum = w + "p";
    }
}
```

Third, the `OpenXmlPowerTools` namespace contains extension methods (similar but not identical to the ones provided by this PR), many namespace-related classes, and a lot of useful functionality (not shown in the example).

```csharp
// OpenXmlPowerTools: PtOpenXmlUtil.cs

using DocumentFormat.OpenXml.Linq;

namespace OpenXmlPowerTools
{
    // The following class contains the extension methods that could be changed to use the new API.
    public static class PtOpenXmlExtensions
    {
        public static XDocument GetXDocument(this OpenXmlPart part)
        {
            // Leverage new Linq feature provided by DocumentFormat.OpenXml.Linq library.
            return new XDocument(new XDeclaration("1.0", "UTF-8", "yes"), part.GetXElement());
        }

        public static void PutXDocument(this OpenXmlPart part, XDocument document)
        {
            // Leverage new Linq feature provided by DocumentFormat.OpenXml.Linq library.
            part.SetXElement(document.Root);
        }
    }

    // This is one example of a namespace-related class that already exists in the OpenXmlPowerTools library.
    public static class W
    {
        public static readonly XNamespace w = "http://schemas.openxmlformats.org/wordprocessingml/2006/main";
        public static readonly XName abstractNum = w + "p";
    }
}
```

Fourth, when using both libraries from your own project, you'd do something like this:

```csharp
// My own project

using DocumentFormat.OpenXml.Linq;
using OpenXmlPowerTools;

namespace MyOwnNamespace
{
    public static class MyOwnClass
    {
        public static void MyOwnMethod(MainDocumentPart part)
        {
            // Get the root element, using the extension method existing in the DocumentFormat.OpenXml.Linq namespace.
            // You might have used part.GetXDocument().Root before, but you now want to use the new feature.
            // There is no issue because the same name does not exist in the OpenXmlPowerTools namespace.
            var document = part.GetXElement();

            // Assume you are using some other feature provided by the OpenXmlPowerTools, which is why
            // you've imported the namespace OpenXmlPowerTools.

            // Do something else for which you need one or more of the namespace-related classes.
            // If we are not specifically including it from either namespace, this is where we'll get a CS0104:
            // 'W' is an ambiguous reference between 'DocumentFormat.OpenXml.Linq.W' and 'OpenXmlPowerTools.W'.
            // The solution here is to say either:
            // using W = DocumentFormat.OpenXml.Linq.W; or
            // using W = OpenXmlPowerTools.W;
            var paragraphs = document.Descendants(W.p);
        }
    }
}
```

### Discussion

The example above shows a situation where developers might want to use the feature (and its API) but not (yet) the namespace-related classes. While the feature provides additional functionality or behavior, the namespace-related classes don't add value immediately, unless you require names that are not provided by the Open XML PowerTools.

If they've used the Open XML PowerTools, using just the API is not possible without explicitly declaring which namespace-related class they want to use. 

I understand that you do not want to add the API to the `OpenXmlPart`, which would immediately solve this problem, so we need to find another way to avoid the naming conflict.

### Solution

To be able to separately opt-in to:

1. the feature with the extension methods as its API (`OpenXmlLinqExtensions` etc.) and
2. the namespace-related classes (e.g., `W`),

we should put both (1) and (2) into separate namespaces. For example:

1. `DocumentFormat.OpenXml.Linq` for the feature etc. and
2. `DocumentFormat.OpenXml.Linq.Names` for the namespace-related classes.

Any other sensible combination or names would work as well.

What are your thoughts?

twsouthwick:
This seems like it would be ideal as a separate discussion. I've created an issue to discuss that further. It has become apparent in this discussion that apparently there's also a constraint that wasn't part of the initial design of integrating with the power tools. I'm fine with that, but it would have been nice to have the set forth initially. Let's discuss it in that issue and see what we can do.

PR #1063, opened 2021-11-04
Use multiple PublicAPI.Shipped.txt files

twsouthwick:
A recent update to the PublicAPI allows to merge multiple files into a single data source. Since the majority of the APIs in the SDK are shared, with only a few different due to framework issues, this greatly simplifies API management.


twsouthwick:
This is blocked by the library not being on NuGet

Issue #1069, opened 2021-11-15
Can't open presentations containing 3D objects (MIME type mismatch)


andrewmolyneux:
**Description**

After updating to SDK version 2.14.0 from 2.10.1, I noticed that opening presentations containing 3D objects causes an exception. Error message is similar to "The document cannot be opened because there is an invalid part with an unexpected content type. [Part Uri=/ppt/media/model3d1.glb] [Content Type=model/gltf.binary] [Expected Content Type=model/gltf-binary]."

**Information**

- .NET Target: .NET Framework
- DocumentFormat.OpenXml Version: 2.14.0

**Repro**

Using this test file created with PowerPoint 2019:
[3d.pptx](https://github.com/OfficeDev/Open-XML-SDK/files/7538059/3d.pptx)

```csharp
using System;
using DocumentFormat.OpenXml.Packaging;

namespace ReproOxml3dBug
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var presDoc = PresentationDocument.Open("3d.pptx", false))
            {
                Console.WriteLine("Opened successfully");
            }
        }
    }
}
``` 

**Observed**

An exception occurs as shown below:
```Unhandled Exception: DocumentFormat.OpenXml.Packaging.OpenXmlPackageException: The document cannot be opened because there is an invalid part with an unexpected content type.
[Part Uri=/ppt/media/model3d1.glb],
[Content Type=model/gltf.binary],
[Expected Content Type=model/gltf-binary].
   at DocumentFormat.OpenXml.Packaging.OpenXmlPart.Load(OpenXmlPackage openXmlPackage, OpenXmlPart parent, Uri uriTarget, String id, Dictionary`2 loadedParts)
   at DocumentFormat.OpenXml.Packaging.OpenXmlPartContainer.LoadReferencedPartsAndRelationships(OpenXmlPackage openXmlPackage, OpenXmlPart sourcePart, RelationshipCollection relationshipCollection, Dictionary`2 loadedParts)
   at DocumentFormat.OpenXml.Packaging.OpenXmlPart.Load(OpenXmlPackage openXmlPackage, OpenXmlPart parent, Uri uriTarget, String id, Dictionary`2 loadedParts)
[...]

PR #1075, opened 2021-11-19
Issue1069 check for model/gltf.binary content type and accept

tomjebo:
This fix checks to see if the part content type is "model/gltf.binary" and the expected part is "model/gltf-binary". If so, then the SDK accepts the part as is. Both will be consumed but the SDK will continue to write out "model/gltf-binary" as this is the current standard and is accepted by PowerPoint.


tomjebo:
Fixes #1069

sergey-tihon:
@tomjebo would be nice to have test for this case in the repo ;)

I build SDK from this branch and confirm that it works on my files.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -78,13 +78,16 @@ internal void Load(OpenXmlPackage? openXmlPackage, OpenXmlPart? parent, Uri uriT
 
             if (IsContentTypeFixed && metroPart.ContentType != ContentType)
             {
-                var errorMessage = SR.Format(
+                if (!((metroPart.ContentType == "model/gltf.binary") && (ContentType == "model/gltf-binary")))
```
twsouthwick, line 81: can you not special case this? Maybe add a virtual method to OpenXmlPart that `IsValidContentType` that you can override in the type that allows this contenttype to also check that

twsouthwick (changes requested):

twsouthwick:
Don't forget to update the changelog

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
tomjebo: @twsouthwick  thanks for the suggestion, I added the virtual method for this and fixed it in the Model3DReferenceRelationshipPart class. please review.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [2.14.x] - 2021-11-23
+
+### Added
+- Virtual content type checking for OpenXmlPart and derived types and model/gltf.binary fix (#1069)
```
twsouthwick, line 10:
```suggestion
- Added virtual method for content type checking in `OpenXmlPart` and fixed issue for `model/gltf.binary` (#1069)
```

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO.Packaging;
+
+namespace DocumentFormat.OpenXml.Packaging
+{
+    /// <summary>
+    /// Defines the Model3DReferenceRelationshipPart
+    /// </summary>
+    public partial class Model3DReferenceRelationshipPart : OpenXmlPart, IFixedContentTypePart
```
twsouthwick, line 16: nit: no need for the base types here

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// Determines if the content type agrees with this part's constraints.
+        /// </summary>
+        /// <param name="metroPart"></param>
+        /// <returns></returns>
```
twsouthwick, line 22:
```suggestion
        /// <inheritdoc />
```

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
```diff
@@ -0,0 +1,33 @@
...
+        private protected override bool IsValidContentType(PackagePart metroPart)
+        {
+            if (IsContentTypeFixed && metroPart.ContentType != ContentType)
+            {
+                return metroPart.ContentType == "model/gltf.binary";
+            }
+
+            return true;
```
twsouthwick, line 30: what about it's default type? I'd imagine you'd want to call `base.IsValidContentType(metroPart)` at some point

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// Defines the Model3DReferenceRelationshipPart
+    /// </summary>
+    public partial class Model3DReferenceRelationshipPart : OpenXmlPart, IFixedContentTypePart
+    {
+        /// <summary>
+        /// Determines if the content type agrees with this part's constraints.
+        /// </summary>
+        /// <param name="metroPart"></param>
+        /// <returns></returns>
+        private protected override bool IsValidContentType(PackagePart metroPart)
```
twsouthwick, line 23: nit: why is it `metroPart`? Can you change it to just `part`?

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            if (IsContentTypeFixed && metroPart.ContentType != ContentType)
```
twsouthwick, line 25: I don't believe we need the `IsContentTypeFixed` check - we *know* this type is. That check should be done before calling this method

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -76,19 +76,19 @@ internal void Load(OpenXmlPackage? openXmlPackage, OpenXmlPart? parent, Uri uriT
             // TODO: should we delay load?
             var metroPart = openXmlPackage.Package.GetPart(uriTarget);
 
-            if (IsContentTypeFixed && metroPart.ContentType != ContentType)
+            _packagePart = metroPart;
+
+            if (!IsValidContentType(metroPart))
```
twsouthwick, line 81: I think you'll want to keep the `IsContentTypeFixed` check here before calling out to the method

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -76,19 +76,19 @@ internal void Load(OpenXmlPackage? openXmlPackage, OpenXmlPart? parent, Uri uriT
...
             {
                 var errorMessage = SR.Format(
-                    ExceptionMessages.InvalidPartContentType,
-                    metroPart.Uri.OriginalString,
-                    metroPart.ContentType,
-                    ContentType);
+                ExceptionMessages.InvalidPartContentType,
```
twsouthwick, line 84: nit: can you keep the indentation?

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -496,6 +496,16 @@ internal long MaxCharactersInPart
         /// </summary>
         internal bool IsContentTypeFixed => this is IFixedContentTypePart;
 
+        /// <summary>
+        /// Determines if the content type agrees with this part's constraints.
+        /// </summary>
+        /// <param name="metroPart"></param>
+        /// <returns></returns>
+        private protected virtual bool IsValidContentType(PackagePart metroPart)
```
twsouthwick, line 504: nit: can you rename to `part` from `metroPart`?

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
```diff
@@ -496,6 +496,16 @@ internal long MaxCharactersInPart
         /// </summary>
         internal bool IsContentTypeFixed => this is IFixedContentTypePart;
 
+        /// <summary>
+        /// Determines if the content type agrees with this part's constraints.
+        /// </summary>
+        /// <param name="metroPart"></param>
+        /// <returns></returns>
```
twsouthwick, line 503: Can you fill these in?

twsouthwick (changes requested):

twsouthwick:
Can you add a test for this?

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
tomjebo: Do I need the /// <summary> block or can I also use /// <inheritdoc/> on it like below?

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
tomjebo: I think you're right. I initially wrote it like the base so it wouldn't need to call the base but really we should use a pattern of calling the base first, IMO.

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
tomjebo: yes, correct.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
tomjebo: yes

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
tomjebo: fixed

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
tomjebo: I changed it to call the base first then do the special case.

twsouthwick: As-is, the base method isn't doing anything....

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        /// <inheritdoc/>
+        private protected override bool IsValidContentType(PackagePart part)
+        {
+            base.IsValidContentType(part);
+            if (part.ContentType != ContentType)
+            {
+                // model/gltf.binary is an acceptable MIME for this part, PowerPoint writes this on save.
+                return part.ContentType == "model/gltf.binary";
+            }
```
twsouthwick, line 26:
```suggestion
            if (base.IsValidContentType(part))
            {
                return true;
            }

             // model/gltf.binary is an acceptable MIME for this part. It's not in the standard, but PowerPoint writes this on save.
             return part.ContentType == "model/gltf.binary";
```

twsouthwick (commented):
Not sure the current override will do what you're expecting. It would be great to have a test for this.

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
```
twsouthwick, line 5:
```suggestion
```

All files in this project already have nullable enabled, so we don't need to manually enable it

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
twsouthwick: Nope, actually neither are needed. It's a partial class, so the original part will have the comments. I'm not certain how comments for partial classes work - they may be combined, or just the first one is used. Either way, you can just remove these.

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
tomjebo: Oh, missed that. I'll remove it.

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
```diff
@@ -261,5 +261,27 @@ public void TestDataReferenceRelationshipsAreClonedCorrectly()
...
+                {
+                    Assert.NotNull(presDoc);
+                }
+            }
+            catch (OpenXmlPackageException)
+            {
+                validationResult = false;
+            }
+
+            Assert.True(validationResult);
```
twsouthwick, line 284:
```suggestion
              using (var presDoc = PresentationDocument.Open(TestFiles._3DTest, false))
              {
                  Assert.NotNull(presDoc);
              }
```

src/DocumentFormat.OpenXml/Packaging/Model3DReferenceRelationshipPart.cs
tomjebo: removed

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
```diff
@@ -261,5 +261,24 @@ public void TestDataReferenceRelationshipsAreClonedCorrectly()
...
+        [Fact]
+        public void TestOpenModel3DWrittenByPowerPoint()
+        {
+            using (var testFile = GetStream(TestFiles._3DTestDash, false))
+            using (var presDoc = PresentationDocument.Open(testFile, false))
+            {
+                Assert.NotNull(presDoc);
+            }
+
+            using (var testFile = GetStream(TestFiles._3DTestDot, false))
```
twsouthwick, line 277: Can these be separate tests? Also, it would nice to assert on the content type of the package

twsouthwick (approved):
A couple comments, but do with them as you want. Once you've addressed the warnings, feel free to merge!

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
tomjebo: @twsouthwick not sure what you mean by "content type of the package". Do you mean assert on the model/gltf-binary content type (the part) or do you mean something else?  The test is pretty much about validating the model/gltf(-|.)binary content type so I don't know why it should explicitly assert this. Maybe you mean assert that the glb part is not something else?

twsouthwick: Because then when you look at the test, you know **that** is what you're testing here. In a year, or 5 years, you may not remember exactly what the test is supposed to be doing

tomjebo: what i meant was, there is no content type of the package. do you mean assert on the content type of the .glb part? If so, I can probably add code to find the part and check that it's either model/gltf-binary or model/gltf.binary.  But we say that in the comments and the code will assert that naturally.

tomjebo: But I can see what you mean. It's not clear what the test is actually looking for other than a clean open with a non-null package object. I'll add some code to look for the part and see if it's one of the two.

tomjebo: Actually, it will have to be model/gltf-binary. The part class (Model3DReferenceRelationshipPart) will always show as model/gltf-binary internally and will write this. But will just look the other way when loading model/glft.binary. I added the assert for it.

PR #1080, opened 2021-11-24
Add Examples

mikeebowen:
This adds examples for Animated Model 3D, Named Sheet Views, Document Tasks, and SVG support.


samples/AnimatedModel3DExample/Program.cs
```diff
@@ -0,0 +1,408 @@
...
+using System.IO;
+using System.Linq;
+
+namespace AnimatedModel3DExample
+{
+    internal class Program
+    {
+        static void Main(string[] args)
+        {
+            if (args.Length < 3)
```
twsouthwick, line 16: Can you make a common file for this stuff that we can include in all the samples?

samples/SVGExample/SVGExample.csproj
```diff
@@ -0,0 +1,20 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="DocumentFormat.OpenXml" Version="2.14.0" />
+    <PackageReference Include="Svg" Version="3.3.0" />
+    <PackageReference Include="System.IO.Packaging" Version="6.0.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="DocumentFormat.OpenXml">
```
twsouthwick, line 15: These should be project references and built as part of the solution. You'll want to update the project file to be similar to the SunburstChartExample so things build on CI correctly

twsouthwick (changes requested):

twsouthwick:
The samples themselves are fine, it's more making sure they fit in well with the rest of the project

Issue #1081, opened 2021-11-28
Feature Proposal: method to retrieve ImagePartType from file extension


JaimeStill:
**Description**

In working through some abstractions for working with images in wordprocessing documents, I wrote a method for getting the `ImagePartType` based on the provided extension. I'm trying to see the utility of incorporating this into the API, and if so, where I might incorporate it?

```cs
// proposed method
static ImagePartType GetImagePartType(string ext) =>
    ext.ToLower() switch
    {
        ".bmp"  => ImagePartType.Bmp,
        ".emf"  => ImagePartType.Emf,
        ".ico"  => ImagePartType.Icon,
        ".jpg"  => ImagePartType.Jpeg,
        ".jpeg" => ImagePartType.Jpeg,
        ".pcx"  => ImagePartType.Pcx,
        ".png"  => ImagePartType.Png,
        ".svg"  => ImagePartType.Svg,
        ".tiff" => ImagePartType.Tiff,
        ".wmf"  => ImagePartType.Wmf,
        _ => throw new NotSupportedException($"{ext} is not supported")
    };

// example usage
static ImagePart LoadImageIntoDoc(MainDocumentPart main, FileInfo file)
{
    var imagePart = main.AddImagePart(GetImagePartType(file.Extension));
    // remaining method excluded for brevity
}
```

rmboggs:
I think this is a good idea.  I'm not sure where this would fit in this project though.

JaimeStill:
@rmboggs capabilities similar to this (but inverse) exist in an [`ImagePartTypeInfo`](https://github.com/OfficeDev/Open-XML-SDK/blob/main/src/DocumentFormat.OpenXml/Packaging/ImagePartTypeInfo.cs) class, but it is not part of the public facing API.

tomjebo:
@JaimeStill That looks very doable, please feel free to submit a PR and we can iterate on it if needed.

JaimeStill:
Closed by #1082

Asbjoedt:
Hi
This is interesting. Of course the list of image extensions are way longer than the one here, and you can't add everyone, but I would like this one added.
```".tif" => ImagePartType.Tiff,```

I also saw .gif is missing.
```".gif" => ImagePartType.Gif,```

I can make a PR myself (my first) that adds these two line.

PR #1082, opened 2021-11-29
AddImagePart extension method to support GetImagePartType feature

JaimeStill:
I was unsure of where to add the functionality described in #1081

I ultimately decided that it might be best to create an `ImagePartExtensions` class that:

1. Defines a `GetImagePartType` method given an associated file extension string.
2. Overloads `AddImagePart` with a version that takes the file extension string and adds the `ImagePart` using the native method, but retrieving the `ImagePartType` via the above method.

If there is a better place / approach for this feature, please let me know so that I have a better grasp on where to accomplish similar features in the future should they arise. Thanks!


src/DocumentFormat.OpenXml/Packaging/ImagePartTypeExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <param name="main">The source MainDocumentPart that is adding the ImagePart.</param>
+        /// <param name="extension">The file extension of the image.</param>
+        public static ImagePart AddImagePart(this MainDocumentPart main, string extension) =>
+            main.AddImagePart(GetImagePartType(extension));
+
+        /// <summary>
+        /// Given a file extension, retrieve the associated ImagePartType value.
+        /// </summary>
+        /// <param name="extension">A file extension supported by ImagePartType.</param>    
+        public static ImagePartType GetImagePartType(this string extension)
```
twsouthwick, line 27: can you move this to the IMagePartTypeInfo class as it would be symmetric operations? Also, I'd prefer this not be an extension method off of string

Does it need to be public?

twsouthwick (changes requested):

src/DocumentFormat.OpenXml/Packaging/ImagePartTypeExtensions.cs
JaimeStill: Moved into `ImagePartTypeInfo` and no longer an extension method. Adjusted `AddImagePart` extension accordingly. Please see 94f53b3d12843f9c014adcbdf1c25106ae73f9b3

src/DocumentFormat.OpenXml/Packaging/ImagePartTypeExtensions.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
```
twsouthwick, line 3: nit: there will be warnings when compiling with whitespace like this. We error out on CI builds if there are any warnings

src/DocumentFormat.OpenXml/Packaging/ImagePartTypeExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace DocumentFormat.OpenXml.Packaging
+{
+    /// <summary>
+    /// Provides extensions to working with ImagePart and ImagePartType
+    /// </summary>
+    public static class ImagePartExtensions
+    {
+        /// <summary>
+        /// Extends an overloaded AddImagePart method that
```
twsouthwick, line 14: can you use `<see cref="..."/>` tags where there actual types/methods?

twsouthwick (changes requested):
Looking good. Thanks @JaimeStill  for the PR. A couple more things, then we can kick off the build and verify it.

Make sure to try it locally with an environment variable `ProjectLoadStyle=All` (by default we don't build all configurations due to size). You can also change the setting in Directory.Build.props (or targets - can't remember)

src/DocumentFormat.OpenXml/Packaging/ImagePartTypeExtensions.cs
JaimeStill: Feedback handled in 828ffa141a9684d12e63e719ff4c013ac56f435c. Additionally, cleared all build warnings. Thanks for helping me get this straight!

twsouthwick:
I believe you should still be getting a warning about public APIs. If not, let's see if the build passes here

JaimeStill:
What do I need to do in order to add ImageTypeExtensions and AddImagePart to the public API?

twsouthwick:
You should get a warning in VS that it's not part of the public API and be offered to add it to the list

JaimeStill:
Sorry, I was using VS Code and didn't have any way of adding to the public API. Should be good to go now with the exception that it is giving end of life warnings for .NET Core 2.1. I merged with the latest from main here, and I'm still getting EOL warnings.

twsouthwick:
those warnings are fine. Thanks for the heads up about VS code. Do you have the csharp extension added in there?

twsouthwick (approved):
LGTM! Thanks

JaimeStill:
> LGTM! Thanks

No worries, thanks for the help!

> those warnings are fine. Thanks for the heads up about VS code. Do you have the csharp extension added in there?

I sure do. It didn't underline the class name or the extension method for me to pull up suggested actions the way it did in VS.

I primarily work from a monorepo where I have a single Web API, an Angular library, and multiple Angular apps, so VS Code works a lot better for me on the day to day. [Example repo](https://github.com/JaimeStill/OpenXmlDocs/tree/main/doc-builder) if interested.

PR #1088, opened 2021-12-05
Enhance Linq feature

ThomasBarnekow:
This PR enhances the `RootXElementFeature`, `IPartRootXElementFeature`, and `OpenXmlLinqExtensions` by adding the capability to work with both `XElement` and `XDocument` representations. It adds the following properties and extension methods to the public API:

```csharp
public interface IPartRootXElementFeature
{
    XDocument Document { get; set; }
}

public static class OpenXmlLinqExtensions
{
    XDocument GetXDocument(this OpenXmlPart part);
    void SetXDocument(this OpenXmlPart part, XDocument document);
    void SaveXDocument(this OpenXmlPart part);
}
```

The `SaveXDocument` extension method is only added for naming consistency with the `GetXDocument` and `SetXDocument` methods. It has the exact same effect as calling `part.SaveXElement()`.

Calling `part.GetXElement()` has the same effect as `part.GetXDocument().Root`. Calling `part.SetXElement(element)` either adds or replaces the `XDocument` instance's root element. In other words, we can freely use either subset of Get, Set, and Save methods or mix and match as desired.

Depending on the use case, the `GetXDocument` method has the benefit of always returning the same instance (unless you change it via `SetXDocument`). For example, you can get the instance from an initially empty part and change the root `XElement` via `SetXElement`. Further, changing and saving the strongly typed `OpenXmlPartRootElement` will only remove the `XDocument` instance's root element but not replace the instance. Thus, you could hold on to that instance in your code if desired.

`XDocument` representations are extensively used by the [Open XML PowerTools](https://github.com/EricWhiteDev/Open-Xml-PowerTools). To leverage the Linq feature in the Open XML PowerTools without lots of breaking changes, we need the ability to get, set, and save `XDocument` representations as well.

### DIFF
--- a/src/DocumentFormat.OpenXml.Linq/Feature/IPartRootXElementFeature.cs
+++ b/src/DocumentFormat.OpenXml.Linq/Feature/IPartRootXElementFeature.cs
@@ -12,6 +12,11 @@ namespace DocumentFormat.OpenXml.Packaging
     public interface IPartRootXElementFeature
     {
         /// <summary>
+        /// Gets or sets the associated <see cref="XDocument" />. Setting will save the underlying <see cref="OpenXmlPart"/>.
+        /// </summary>
+        XDocument Document { get; set; }
+
+        /// <summary>
         /// Gets or sets the associated <see cref="XElement"/>. Setting will save the underlying <see cref="OpenXmlPart"/>.
         /// </summary>
         [DisallowNull]
--- a/src/DocumentFormat.OpenXml.Linq/Feature/RootXElementFeature.cs
+++ b/src/DocumentFormat.OpenXml.Linq/Feature/RootXElementFeature.cs
@@ -15,7 +15,7 @@ namespace DocumentFormat.OpenXml.Packaging
         private readonly OpenXmlPart _part;
         private readonly IPartRootEventsFeature _events;
 
-        private XElement? _rootXElement;
+        private XDocument? _partXDocument;
 
         public RootXElementFeature(OpenXmlPart part)
         {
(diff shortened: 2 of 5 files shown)


ThomasBarnekow:
@twsouthwick, could you please run the build pipeline? 

Will NuGet packages be available in the CI feed already before merging or only after merging into main?

src/DocumentFormat.OpenXml.Linq/Feature/IPartRootXElementFeature.cs
```diff
@@ -11,6 +11,11 @@ namespace DocumentFormat.OpenXml.Packaging
     /// </summary>
     public interface IPartRootXElementFeature
```
twsouthwick, line 12: Maybe we should change this to `IPartRootXLinqFeature`

twsouthwick (approved):
LGTM!

src/DocumentFormat.OpenXml.Linq/Feature/RootXElementFeature.cs
```diff
@@ -25,67 +25,115 @@ public RootXElementFeature(OpenXmlPart part)
...
+
+                _partXDocument = value;
 
                 Save();
             }
         }
 
+        /// <inheritdoc />
+        [DisallowNull]
+        public XElement? Root
```
twsouthwick, line 74: Do we need this on the feature itself? Or would an extension method be sufficient?

ThomasBarnekow: For a public interface, I would want to have it in the interface. If we made the interface internal, an extension method would be sufficient. Here is why:

I currently see two congruent sets of properties and methods in the public API:

1. `Root`, `GetXElement`, `SetXElement`, and `SaveXElement`; and
2. `Document`, `GetXDocument`, `SetXDocument`, `SaveXDocument`.

Removing just the `Root` property would break that congruence. If we made the interface internal, the public API would only have the following methods:

1. `GetXElement`, `SetXElement`, and `SaveXElement`; and
2. `GetXDocument`, `SetXDocument`, `SaveXDocument`.

That would also be totally fine and possibly even better than the first set. In any case, we should make sure the public API provides two congruent sets of properties (optional) and methods (required).

Should we want to make the `IPartRootXElementFeature` interface (current naming) internal, we would also make the `OpenXmlLinqExtensions.GetPartRootXElementFeature` extension method internal.

src/DocumentFormat.OpenXml.Linq/Feature/IPartRootXElementFeature.cs
twsouthwick: Once we release it, adding members to an interface is breaking, so my thoughts with these have been to keep them fairly scoped and then add additional interfaces for new features. 

Yes, the main goal is to make it more generic mostly because now it's not *just* the XElement, but the XDocument as well.

src/DocumentFormat.OpenXml.Linq/Feature/RootXElementFeature.cs
twsouthwick: I'm fine either way here. Just thinking of what is really being abstracted behind the interface and what would ideally be a default implementation if we could use that with the SDK.

twsouthwick:
@ThomasBarnekow I'm going to merge this in as-is and let's think about the naming to make sure we're good with it. It'll probably be released next week, so we'll need to make the changes before then.

I'll be on vacation the rest of the week, so I'll merge this in so you can play with it and make sure it's all how you'd expect.

src/DocumentFormat.OpenXml.Linq/Feature/IPartRootXElementFeature.cs
ThomasBarnekow: Can we make the interface internal?

src/DocumentFormat.OpenXml.Linq/Feature/RootXElementFeature.cs
ThomasBarnekow: See my question above. I suggest we make the interface internal.

PR #1090, opened 2021-12-07
fixes a couple historical namespace issues

tomjebo:
DocumentFormat.OpenXml.Office.Drawing.ShapeTree is available in Office 2010 and above, not 2007.
DocumentFormat.OpenXml.Office2010.ExcelAc changed to DocumentFormat.OpenXml.Office2013.ExcelAc

### DIFF
--- a/src/DocumentFormat.OpenXml/GeneratedCode/schemas_microsoft_com_office_spreadsheetml_2011_1_ac.g.cs
+++ b/src/DocumentFormat.OpenXml/GeneratedCode/schemas_microsoft_com_office_spreadsheetml_2011_1_ac.g.cs
@@ -13,11 +13,11 @@ using System;
 using System.Collections.Generic;
 using System.IO.Packaging;
 
-namespace DocumentFormat.OpenXml.Office2013.ExcelAc
+namespace DocumentFormat.OpenXml.Office2010.ExcelAc
 {
     /// <summary>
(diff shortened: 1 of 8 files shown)


src/DocumentFormat.OpenXml/GeneratedCode/schemas_microsoft_com_office_spreadsheetml_2011_1_ac.g.cs
```diff
@@ -13,7 +13,7 @@
 using System.Collections.Generic;
 using System.IO.Packaging;
 
-namespace DocumentFormat.OpenXml.Office2010.ExcelAc
+namespace DocumentFormat.OpenXml.Office2013.ExcelAc
```
twsouthwick, line 16: you cannot change this - it's set for now (that's a breaking change)

twsouthwick (changes requested):
API surface area on released types cannot be changed in a pre-v3.0 release.

src/DocumentFormat.OpenXml/GeneratedCode/schemas_microsoft_com_office_spreadsheetml_2011_1_ac.g.cs
tomjebo: reverted.

PR #1091, opened 2021-12-08
Fix and enhance Linq feature

ThomasBarnekow:
Fix and enhance Linq feature

This PR makes the following changes:

- Make interface internal
- Fix Save() method by not saving if XDocument.Root is null
- Add IsRootXElementLoaded(this OpenXmlPart part) extension method
- Rename interfaces and classes
- Add unit tests

The new naming is based on the OpenXmlPartRootElement class. We are essentially dealing with XElement representations of OpenXmlPartRootElement instances that can be accessed directly via GetXElement() or indirectly via the containing XDocument that can be accessed via GetXDocument().

### DIFF
--- a/src/DocumentFormat.OpenXml.Linq/Feature/IOpenXmlPartRootXElementFeature.cs
+++ b/src/DocumentFormat.OpenXml.Linq/Feature/IOpenXmlPartRootXElementFeature.cs
@@ -9,7 +9,7 @@ namespace DocumentFormat.OpenXml.Packaging
     /// <summary>
     /// Feature to allow access to <see cref="XElement"/> representation of a <see cref="OpenXmlPart"/>.
     /// </summary>
-    internal interface IPartRootXElementFeature
+    internal interface IOpenXmlPartRootXElementFeature
     {
         /// <summary>
         /// Gets or sets the associated <see cref="XDocument" />. Setting will save the underlying <see cref="OpenXmlPart"/>.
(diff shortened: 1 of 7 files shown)


src/DocumentFormat.OpenXml.Linq/Feature/IOpenXmlPartRootXElementFeature.cs
```diff
@@ -9,7 +9,7 @@ namespace DocumentFormat.OpenXml.Packaging
     /// <summary>
     /// Feature to allow access to <see cref="XElement"/> representation of a <see cref="OpenXmlPart"/>.
     /// </summary>
-    public interface IPartRootXElementFeature
```
twsouthwick, line 12: The value of keeping this public allows a contract that could be implemented differently for testing/custom implementation unknown to the library itself.

twsouthwick (commented):
This change looks fine, but I'm leaning towards keeping the interface public. If we're not sure exactly of the shape, we can make it internal for the current release, but I'd like to keep it public for extensibility purposes.

src/DocumentFormat.OpenXml.Linq/Feature/IOpenXmlPartRootXElementFeature.cs
ThomasBarnekow: I would not have any problem with a public interface. I just made it internal in case you wanted to sort out naming issues, for example.

I am fine with this name specifically (and am hoping that you agree), so that is not the issue. If we could make breaking changes in the next major release (should that be necessary at all), we could make it public right now. However, if it does not have to be public in the short term, we could make it public in the next release and "let it sink in". I'd leave that up to you.

twsouthwick: That all sounds good to me!

twsouthwick (approved):

PR #1092, opened 2021-12-13
Add namespace generator to SDK

twsouthwick:
This adds a source generator that takes a set of namespaces and the associated version it appeared in and generates methods that augments OpenXmlNamespaceResolver to understand these versions.

An additional class is also added to handle backwards compatibility usage of namespace ids, but doesn't add any new namespace ids for future ids. This is added purely to retain any support needed for users that may have stored the id instead of the prefix.


tomjebo (approved):

tomjebo:
ok, yeah, looks like I need this. merging...

PR #1087, opened 2021-12-03
Examples

mikeebowen:
Update examples to use local references and work with updated API.

PR #1101, opened 2021-12-15
Add Linq-to-XML example

ThomasBarnekow:
Based on the SVGExample currently added in PR #1087, this pull request adds corresponding examples that demonstrate two approaches:

1. the Linq-to-XML way of creating markup, using the Linq-to-XML feature provided by `DocumentFormat.OpenXml.Linq` and
2. a strongly-typed way of creating markup that is closer to how you would do it in the Linq-to-XML way.

Let's discuss how we deal with the strongly-typed example. We can do multiple things. Firstly, we could demonstrate different code styles developers can use. Secondly, we could replace the original one with this one.

By the way, I am still using Visual Studio 2019, which is the reason for the initial changes in the solution file. I'm not sure whether that is an issue. One issue, however, is that you've used language version 10.0, which requires .NET 6.0, which requires Visual Studio 2022, which not everybody might have. I'm not sure whether you are using the very latest language features and whether this is really required.


ThomasBarnekow:
@twsouthwick and @mikeebowen, please have a look at the examples.

@mikeebowen, if we keep both your original SVGExample (see `samples/SVGExample/Program.cs` in your PR) and the example provided by the `StronglyTypedTools` class, would it make sense for you to consider some features that I included in my example? For example:

- The creation ID is not a fixed string (GUID in curly braces) but created as a unique ID for each image. Your example suggests it might be a fixed string.
- The image size and position are not fixed but derived from the slide height. This is optional. However, it shows that developers need to think about sizing and positioning.

twsouthwick:
@mikeebowen I like the features @ThomasBarnekow has. Can you incorporate them? It would be great to keep the samples similar.

twsouthwick:
Yes, I'm using C# 10 features such as latest code generator stuff as well as the .NET 6 SDK (which has other goodness besides just C# 10 stuff). Generally, I keep the repo using the latest tool chains and will adopt that as useful. The resulting builds will still work as-is, but want to be able to use the latest as soon as possible :)

twsouthwick:
I like the idea of both. Then it shows the different styles and we can compare them fairly easily

twsouthwick (commented):
LGTM. My only concern is if the image is ok to store in the repo (not sure about licensing of images)

ThomasBarnekow:
> LGTM. My only concern is if the image is ok to store in the repo (not sure about licensing of images)

I took a free image from https://freesvg.org/. They seemingly have no copyright and come with the CC0 1.0 Universal (CC0 1.0)
Public Domain Dedication. However, I could take any other image with copyright by MS that you might have.

twsouthwick:
Can you add an attribution to either the csproj/readme/additional file?

ThomasBarnekow:
> Can you add an attribution to either the csproj/readme/additional file?

Done. See .csproj file.

twsouthwick (approved):

Issue #1093, opened 2021-12-13
Excel workbook generated by Open-XML-SDK are converted to XLSM by Google spread sheet.


Yanal-Yves:
**Description**
First I would like to thank you for this library.

I am posting here a wired behavior when generating an Excel file with Open XML SDK and uploading the file to Google spreadsheet. I believe this is an oddity caused by a bug in Open XML SDK. This is why I am posting here. I don't think this is a question about using Open XML SDK. But in case I'am wrong, I have also posted to [Stack overflow](https://stackoverflow.com/questions/70319867/avoid-google-spreadsheet-to-convert-an-xlsx-file-created-by-open-xml-sdk-to-xlsm) : 

So I am using the [Open XML SDK 2.14.0][1] to create Excel files.
If I upload the XLSX file produced by Open XML SDK to Google Spreadsheet, it opens the file as an XLSX file but if I make any change in the spreadsheet then Google Spreadsheet changes the file type to XLSM. If I generate an empty workbook from MS Excel (Office 360) or Libre Office Calc I don't have this issue : Google Spreadsheet will not change the file type if I change the data.

**Information**

- .NET Target: .net core 3.1
- DocumentFormat.OpenXml Version: (ie 2.14.0)

**Repro**

The minimalistic code bellow ([found in the official documentation][2]) reproduces the issue. It generate a `foo.xlsx` file in `C:\Temp`.

``` c#
var filepath = @"C:\temp\foo.xlsx";

// Create a spreadsheet document by supplying the filepath.
// By default, AutoSave = true, Editable = true, and Type = xlsx.
SpreadsheetDocument spreadsheetDocument = SpreadsheetDocument.Create(filepath, SpreadsheetDocumentType.Workbook);

// Add a WorkbookPart to the document.
WorkbookPart workbookpart = spreadsheetDocument.AddWorkbookPart();
workbookpart.Workbook = new Workbook();

// Add a WorksheetPart to the WorkbookPart.
WorksheetPart worksheetPart = workbookpart.AddNewPart<WorksheetPart>();
worksheetPart.Worksheet = new Worksheet(new SheetData());

// Add Sheets to the Workbook.
Sheets sheets = spreadsheetDocument.WorkbookPart.Workbook.AppendChild<Sheets>(new Sheets());

// Append a new worksheet and associate it with the workbook.
[...]

PR #1107, opened 2021-12-17
add core.xml part to sml create

tomjebo:
this adds a core.xml part to the package when calling SpreadsheetDocument Create(). This works around a problem with Google Sheets per #1093

### DIFF
--- a/src/DocumentFormat.OpenXml/Packaging/SpreadsheetDocument.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/SpreadsheetDocument.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.IO.Packaging;
 using System.Reflection;
+using System.Xml;
 
 namespace DocumentFormat.OpenXml.Packaging
 {
@@ -119,7 +120,15 @@ namespace DocumentFormat.OpenXml.Packaging
         /// <exception cref="ArgumentNullException">Thrown when "path" is null reference.</exception>
         public static SpreadsheetDocument Create(string path, SpreadsheetDocumentType type)
         {
-            return Create(path, type, true);
+            SpreadsheetDocument smlDoc = Create(path, type, true);
+            CoreFilePropertiesPart cfpp = smlDoc.AddCoreFilePropertiesPart();
+            using (XmlTextWriter writer = new XmlTextWriter(cfpp.GetStream(FileMode.Create), System.Text.Encoding.UTF8))
+            {
+                writer.WriteRaw("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\r\n<cp:coreProperties xmlns:cp=\"https://schemas.openxmlformats.org/package/2006/metadata/core-properties\"></cp:coreProperties>");
+                writer.Flush();
+            }
+
+            return smlDoc;
         }
 
         /// <summary>


src/DocumentFormat.OpenXml/Packaging/SpreadsheetDocument.cs
```diff
@@ -119,7 +120,15 @@ private void UpdateDocumentTypeFromContentType()
         /// <exception cref="ArgumentNullException">Thrown when "path" is null reference.</exception>
         public static SpreadsheetDocument Create(string path, SpreadsheetDocumentType type)
         {
-            return Create(path, type, true);
+            SpreadsheetDocument smlDoc = Create(path, type, true);
```
twsouthwick, line 123: I would rather it not be injected here. I would like to see some sort of service that would be run at creation time. Something like

```csharp
public interface IPackageInitializer
{
  void Initialize(OpenXmlPackage package);
}
```

You could inject it via OpenSettings

twsouthwick (changes requested):

src/DocumentFormat.OpenXml/Packaging/SpreadsheetDocument.cs
tomjebo: Ok, we'll take some more time to think about this. Upon reflection, this should not be done in SpreadsheetDocument anyway. I've created a project for this so we don't lose track.  https://github.com/OfficeDev/Open-XML-SDK/projects/9

ThomasBarnekow: Whatever magic we add here should not create breaking changes. When creating documents with the SDC, no parts are added at the moment. However, it would certainly be nice to be able to tell the SDK to create more "complete" documents such as minimal Word, Excel, and PowerPoint documents that fulfill certain criteria. But those should have to be created explicitly (e.g., through settings or otherwise).

ThomasBarnekow:
@tomjebo, having looked at your project description, I wanted to comment on your points here (as I did not see a way to do that in the project).

> Won't add some of the basic parts that some apps like Google Sheets expect. core.xml, app.xml, etc...

This is covered by my comment above regarding the "minimal" documents. It would be a very good idea to add this to the SDK. While other libraries have filled at least some of the gaps, I believe this is so basic that it should be done by the SDK.

> We don't define the root element for CoreFilePropertiesPart and that means the programmer has to add the contents of these parts (at least core.xml) with raw XML like this: [example not shown]

This is true for all parts. In all cases, you have to set that root element explicitly after having added a new part. Using a `WordprocessingDocument` and a `MainDocumentPart` as an example, you'll have to do the following to create a minimal document when using the strongly typed classes:

```csharp
using var wordDocument = WordprocessingDocument.Create(stream, type);
MainDocumentPart part = wordDocument.AddMainDocumentPart();
part.Document =
    new Document(
        new Body(
            new Paragraph()));
```

I am not 100% sure whether the `w:p` element is required by the standard, but Word would add it. I've laid out the code to mimic the layout of the corresponding XML document.

Using the Linq-to-XML way, you'd do it like this.

```csharp
using var wordDocument = WordprocessingDocument.Create(stream, type);
MainDocumentPart part = wordDocument.AddMainDocumentPart();
part.SetXElement(
    new XElement(W.document, new XAttribute(XNamespace.Xmlns + "w", W.w),
        new XElement(W.body,
            new XElement(W.p)))
```

It would be really cool if the SDK could create standards-compliant, minimal parts (e.g., `MainDocumentPart`, `CoreFilePropertiesPart`) when calling the `AddMainDocumentPart`, for example. That way, the `Document` and `RootElement` properties would not have to be marked as nullable.

> We should investigate building these documents more completely like Office, even though Office doesn't throw an error on opening them.

Yes, absolutely!

tomjebo:
When I said we don't define the root element I actually mean we don't wrap it with a generated or hand coded C# class. Other parts we *do* provide this.

tomjebo:
I was going to close this PR but it looks like it might be good to keep it open for discussion. @ThomasBarnekow, thanks for the comments.

twsouthwick:
@tomjebo Let's move relevant content to the issue and close this PR so we don't get stale PRs that we're not planning on merging in at the moment.

tomjebo:
Closing this PR and will continue discussion on Issue #1093.

PR #1108, opened 2021-12-17
Add Examples to solution

mikeebowen:
This adds the samples to the solution.

### DIFF
--- a/samples/DocumentTaskExample/Program.cs
+++ b/samples/DocumentTaskExample/Program.cs
@@ -64,11 +64,8 @@ namespace DocumentTaskSample
 
         private static void AddNewTask(MainDocumentPart mdp, string title, User assignee, User assigner)
         {
-#pragma warning disable TODO // TODO
-                            // TODO: user ids, providers and names will have to be programmatically accessed via directory services.
-                            // These are just examples to show what is expected.
             mdp.AddNewPart<DocumentTasksPart>();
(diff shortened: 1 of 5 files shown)


samples/DocumentTaskExample/Program.cs
```diff
@@ -64,9 +64,11 @@ private static void AddDocumentTask(string filePath)
 
         private static void AddNewTask(MainDocumentPart mdp, string title, User assignee, User assigner)
         {
-            // TODO: userids, providers and names will have to be programmatically accessed via directory services.
-            // These are just examples to show what is expected.
+#pragma warning disable TODO // TODO
```
twsouthwick, line 67: What's this for?

twsouthwick (commented):
LGTM except I'm not sure what the `pragma remove TODO` is for. Can you remove that before merging?

twsouthwick (approved):

samples/DocumentTaskExample/Program.cs
mikeebowen: If the TODO warning isn't disabled, the build fails with stylecop errors. I wanted to leave the TODO in so users would know this was dummy data, so I disabled the error for this line. Thinking about it now, I should have just deleted "TODO" and left the rest of the comment...

twsouthwick (approved):

PR #1125, opened 2022-02-17
Add data files for code generation

twsouthwick:


data/parts/ChartStylePart.json
```diff
@@ -0,0 +1,13 @@
+{
+  "Root": "cs:CT_ChartStyle",
+  "Name": "ChartStylePart",
+  "Base": "OpenXmlPart",
+  "ContentType": "application/vnd.ms-office.chartstyle\u002Bxml",
```
tomjebo, line 5: this is probably something I'll have to look into: why the unicode char in the content type. I don't think those are in what Office emits but I could be wrong. not really an action for you @twsouthwick

data/schemas/schemas-microsoft-com_office_powerpoint.json
```diff
@@ -0,0 +1,27 @@
+{
+  "TargetNamespace": "urn:schemas-microsoft-com:office:powerpoint",
+  "Types": [
+    {
+      "Name": "pvml:CT_Empty/pvml:iscomment",
```
tomjebo, line 5: Just a question: why Name property is type/element like this. Should it be two separate properties?

data/schemas/schemas-microsoft-com_office_word.json
```diff
@@ -0,0 +1,365 @@
...
+      "Name": "w10:CT_Wrap/w10:wrap",
+      "ClassName": "TextWrap",
+      "Summary": "Text Wrapping.",
+      "BaseClass": "OpenXmlLeafElement",
+      "IsLeafElement": true,
+      "Attributes": [
+        {
+          "QName": ":type",
+          "PropertyName": "Type",
+          "Type": "EnumValue\u003CDocumentFormat.OpenXml.Vml.Wordprocessing.WrapValues\u003E",
```
tomjebo, line 87: hm... again with the unicode chars. @twsouthwick do you know why this is?

data/schemas/schemas-microsoft-com_office_powerpoint.json
twsouthwick: space saving :) Makes the file about 10% smaller

data/schemas/schemas-microsoft-com_office_word.json
twsouthwick: JSON encoding

data/parts/ChartStylePart.json
ThomasBarnekow: I can confirm the JSON encoding thing. It looks like you are using `System.Text.Json`, which encodes more characters than `Newtonsoft.Json` when serializing. For example, `Newtonsoft.Json` does not encode `+` characters as `\u002B` (at least not by default) while `System.Text.Json` does. However, both will correctly deserialize the serialized value:

- `"application/vnd.ms-office.chartstyle\u002Bxml"`

into:

- `"application/vnd.ms-office.chartstyle+xml"`.

data/schemas/schemas-microsoft-com_office_word.json
ThomasBarnekow: Yep, as above, both `Newtonsoft.Json` and `System.Text.Json` will deserialize as:

- `"EnumValue<DocumentFormat.OpenXml.Vml.Wordprocessing.WrapValues>"`

And `Newtonsoft.Json` would also not encode the `<` and `>` characters. 

The fact that `System.Text.Json` encodes many characters that are not encoded by `Newtonsoft.Json` leads to surprises for developers who are used to the `Newtonsoft.Json` behavior.

data/schemas/schemas-microsoft-com_office_powerpoint.json
ThomasBarnekow: What is "10%" in terms of absolute space (roughly)? The question is whether the 10% is worth the compromise. If you wanted to save space, you could also use shorter JSON property (key) names (e.g., `"tns"` instead of `"TargetNamespace"`.

twsouthwick: Space savings is in regards to line count

Also, this property is a "key" to which element it is - I believe (and my memory could be wrong here as I just got back from a vacation) that this combination leads to a unique lookup for the types.

data/schemas/schemas-microsoft-com_office_word.json
tomjebo: Thanks Thomas and Taylor for the explanations.

data/schemas/schemas-microsoft-com_office_powerpoint.json
tomjebo: Ok. I just wasn't sure if this was needed by design. I prefer to have separate properties but it sounds like this was needed.

data/parts/ChartStylePart.json
tomjebo: Again, thanks Taylor and Thomas for the explanations.

tomjebo (approved):

PR #1126, opened 2022-02-17
OpenXmlPart: added UnloadRootElement method

### LINKED ISSUE #1122: SlidePart.Slide: how to unload memory
Hi,
I'm reading large pptx files slide-by-slide with limited RAM (azure functions), so after using SlidePart.Slide property the slide is loaded into memory and will be unloaded only when the pptx will be unloaded which means that memory consumption is growing with every slide.

For now I'm using internal `OpenXmlPart.SetPartRootElementToNull()` via reflection to release the memory. 
https://github.com/OfficeDev/Open-XML-SDK/blob/8a29bec8be9221e6b2eefb962f41b95b8391a773/src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs#L608

Is there an official way to unload memory when it's not needed anymore? Or maybe it's possible to mark this method as public?

Thanks.

_Originally posted by @f1nzer in https://github.com/OfficeDev/Open-XML-SDK/discussions/1121_

f1nzer:
Closes #1122

### DIFF
--- a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/OpenXmlPart.cs
@@ -351,6 +351,24 @@ namespace DocumentFormat.OpenXml.Packaging
             sourceStream.CopyTo(targetStream);
         }
 
+        /// <summary>
+        /// Unloads the RootElement.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


twsouthwick (approved):
LGTM! Thanks

PR #1164, opened 2022-06-20
Separate framework from typed classes

twsouthwick:
This is an initial change to move the infrastructure of the SDK to a separate assembly. This is done in a way that will not break the current set up by defining type forwards from `DocumentFormat.OpenXml` to the new `DocumentFormat.OpenXml.Framework` assembly. This is a WIP and updates will be pushed to the PR as progress is made. The change is expected to be made before 2.18.0 and is a precursor to any changes to separate out by document type (word/excel/powerpoint).


twsouthwick:
Blocked by #1185

PR #1171, opened 2022-06-28
M365 May 2022 Update

mikeebowen:
- Added PowerPointCommentPart relationship to SlidePart (https://github.com/OfficeDev/Open-XML-SDK/pull/1137)
  - This deprecates the relationship between PowerPointCommentPart and PresentationPart, which will be removed in a future version
- Removed dogfood Reactions feature


test/DocumentFormat.OpenXml.Packaging.Tests/ParticleTests.cs
```diff
@@ -223,6 +223,7 @@ private void AssertEqual(Dictionary<Type, VersionCollection<ParticleConstraint>>
             var tmp = Path.GetTempFileName();
 
             _output.WriteLine($"Writing output to {tmp}");
+            _output.WriteLine("Expected output in DocumentFormat.OpenXml.Packaging.Tests.data.Particles.json");
```
twsouthwick, line 226:
```suggestion
            _output.WriteLine(@"Expected output in data\Particles.json");
```

that way you can actually find the file

src/DocumentFormat.OpenXml/OpenXmlCompositeElement.cs
```diff
@@ -9,6 +9,7 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
+using System.Reflection;
```
twsouthwick, line 12:
```suggestion
```

unnecessary

twsouthwick (approved):
LGTM

mikeebowen:
> How did you add the shim? I don't see custom code for that anywhere

It's on a branch [on my fork](https://github.com/mikeebowen/Open-XML-SDK/blob/add-extension-list-utils/src/DocumentFormat.OpenXml/Schema/Office2021/Word/CommentsExt/CommentsExt.cs)

It's not included in the PR, because the Schema for Reaction is in dogfood and will change, so we don't want to release something with the current Reaction, because it would allow users to create invalid packages. Without Reaction, there are no children of the CommentsExtensibleExtensionList, so we don't need to shim ExtensionList, yet. When the schema for Reaction is finalized, we'll have the shim ready.

twsouthwick (approved):

PR #1196, opened 2022-08-24
M365 aug22

mikeebowen:


mikeebowen:
Changelog still needs to be updated.

twsouthwick:
I'd like to wait for https://github.com/OfficeDev/Open-XML-SDK/pull/1199 to ensure there are no regressions to the generated stuff

twsouthwick (dismissed):
This LGTM re: the structure. I'm not the one to verify expected schema updates :)

data/schemas/schemas_openxmlformats_org_wordprocessingml_2006_main.json
```diff
@@ -40521,19 +40468,23 @@
           "QName": "w:numId",
           "PropertyName": "NumberID",
           "Type": "Int32Value",
-          "PropertyComments": "Numbering Definition Instance ID",
```
ThomasBarnekow, line 40524: Why do you remove the descriptive `PropertyComments` and replace it with the local name of the Open XML element?

I don't know what the convention for those `PropertyComments` values is, but it looks like there is some inconsistency in those that are introduced in this commit.

data/schemas/schemas_openxmlformats_org_wordprocessingml_2006_main.json
```diff
@@ -40521,19 +40468,23 @@
...
+          "Type": "Int32Value",
+          "PropertyComments": "durableId"
         }
       ],
       "Children": [
         {
           "Name": "w:CT_NonNegativeDecimalNumber/w:abstractNumId",
-          "PropertyName": "AbstractNumId",
-          "PropertyComments": "Abstract Numbering Definition Reference"
+          "PropertyName": "AbstractNumId"
```
ThomasBarnekow, line 40487: Why do you remove the `PropertyComments`?

data/schemas/schemas_openxmlformats_org_wordprocessingml_2006_main.json
```diff
@@ -51361,6 +51312,59 @@
...
+      "ClassName": "LevelOverride",
+      "Summary": "Defines the LevelOverride Class.",
+      "CompositeType": "OneSequence",
+      "BaseClass": "OpenXmlCompositeElement",
+      "Attributes": [
+        {
+          "QName": "w:ilvl",
+          "PropertyName": "LevelIndex",
+          "Type": "Int32Value",
+          "PropertyComments": "Numbering Level ID",
```
ThomasBarnekow, line 51327: See above. This is an example of descriptive `PropertyComments`. In the cases above, either the corresponding XML element's local name was used, or no value was provided (i.e., the key was deleted).

ThomasBarnekow (commented):
While I can't judge the schema changes reflected in the JSON files, it looks like there are some minor inconsistencies in the `PropertyComments`. Assuming those `PropertyComments` are not critical, this should hold up the release.

data/schemas/schemas_openxmlformats_org_wordprocessingml_2006_main.json
tomjebo: @ThomasBarnekow we've noticed this in the past as well and have not gotten around to understanding it. The backend processor does all this automatically. So we'll need to take some time to review the code and understand why it changes these. They should not hold up the code release as they don't affect the logic or API surface.

tomjebo: @ThomasBarnekow  this answer relates to all your questions about the PropertyComments as they all are automatically generated and we didn't explicitly change anything. We will check on all of the issues.

AlfredHellstern (approved):
LGTM

PR #1235, opened 2022-11-19
Persist generated code

twsouthwick:
This is helpful to see what the data files actually generate and potentially catch any changes that are unexpected.


tomjebo (commented):
it there a way to keep from emitting these one liner comment .g.cs files?

tomjebo (approved):

tomjebo (dismissed):

tomjebo (approved):

PR #1236, opened 2022-11-20
FlatOpc: avoid unnecessary MemoryStream.ToArray call

f1nzer:
There is an unnecessary allocation in PR #1232: it allocates a new byte array via `MemoryStream.ToArray()`.

It's possible to avoid this call and read the data into a byte array directly, as it was done before (but was broken due to the [net6 breaking change](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/partial-byte-reads-in-streams)).

### DIFF
--- a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
@@ -132,22 +132,7 @@ namespace DocumentFormat.OpenXml.Packaging
             using var stream = part.GetStream();
             var size = checked((int)stream.Length);
             var buffer = new byte[size];
-
-#if NET7_0_OR_GREATER
             stream.ReadExactly(buffer);
-#else
-            var totalRead = 0;
-            while (totalRead < size)
-            {
-                var bytesRead = stream.Read(buffer, totalRead, size - totalRead);
-                if (bytesRead == 0)
-                {
-                    break;
-                }
-
-                totalRead += bytesRead;
-            }
-#endif
 
             return ToChunkedBase64String(buffer);
         }
--- a/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
@@ -130,11 +130,26 @@ namespace DocumentFormat.OpenXml.Packaging
         private static string ToChunkedBase64String(PackagePart part)
         {
             using var stream = part.GetStream();
-            using var memory = new MemoryStream((int)stream.Length);
+            var size = checked((int)stream.Length);
+            var buffer = new byte[size];
 
-            stream.CopyTo(memory);
+#if NET7_0_OR_GREATER
+            stream.ReadExactly(buffer);
+#else
+            var totalRead = 0;
+            while (totalRead < size)
+            {
+                var bytesRead = stream.Read(buffer, totalRead, size - totalRead);
+                if (bytesRead == 0)
+                {
+                    break;
+                }
 
-            return ToChunkedBase64String(memory.ToArray());
+                totalRead += bytesRead;
+            }
+#endif
+
+            return ToChunkedBase64String(buffer);
         }
 
         private static string ToChunkedBase64String(byte[] byteArray)
--- a/src/common/System/IO/StreamReadExtensions.cs
+++ b/src/common/System/IO/StreamReadExtensions.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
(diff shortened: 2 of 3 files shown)


src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -130,11 +130,26 @@ private static XElement GetBinaryPartContentsAsXml(PackagePart part)
         private static string ToChunkedBase64String(PackagePart part)
         {
             using var stream = part.GetStream();
-            using var memory = new MemoryStream((int)stream.Length);
+            var size = checked((int)stream.Length);
+            var buffer = new byte[size];
```
twsouthwick, line 134: if you're going for perf, would you like to update this to use `ArrayPool<byte>` here as well? Then you'd not need to allocate on platforms that supports that.

NOTE: we're currently not shipping a .NET 6+ build, but I'm fine with having the #if/#defs for perf stuff and we can decide to turn a build for that on fairly easily.

twsouthwick:
Thanks @f1nzer. I'm fine with this change - I was planning on revisiting this once I got the tests working, so thanks for addressing it! If you're interested in further optimziation with ArrayPools, let me know, otherwise I can merge this in as-is.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
f1nzer: > if you're going for perf, would you like to update this to use ArrayPool<byte> here as well? Then you'd not need to allocate on platforms that supports that.

FlatOpc is not a very popular variation of OOXML, so I'm OK with the current implementation. This PR was intended to reduce memory footprint only 😄 

> NOTE: we're currently not shipping a .NET 6+ build, but I'm fine with having the #if/#defs for perf stuff and we can decide to turn a build for that on fairly easily.

I've added if/def just in order not to forget in the future that there is an existing method for that.

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -130,11 +130,26 @@ private static XElement GetBinaryPartContentsAsXml(PackagePart part)
...
 
-            stream.CopyTo(memory);
+#if NET7_0_OR_GREATER
+            stream.ReadExactly(buffer);
+#else
+            var totalRead = 0;
```
twsouthwick, line 139: Could you move this to an internal extension method so that call site here just calls `ReadExactly`? The `common/` folder has a bunch of these polyfills where that could live

twsouthwick:
@f1nzer do you think you'll be able to make the change to an extension method? If not, I'll merge it in as is

src/DocumentFormat.OpenXml/Packaging/OpenXmlPackage.FlatOpc.cs
f1nzer: @twsouthwick
I can add a `StreamReadExtensions` file with the `ReadExactly` shim, but this file should be included only on NET6.0 or below, and I have no idea how it could be nicely done in [this shim targets file](https://github.com/OfficeDev/Open-XML-SDK/blob/main/src/common/FrameworkShims.targets) 😅

twsouthwick: I'm fine with doing either:

1) Create the file with an #if/#def around it from `#if !NET7_0_OR_GREATER` and include it by default in a new itemgroup with no conditions
2) Create the file and add it into all the sections for applicable TFMs, *and* create a new `<ItemGroup Condition=" '$(TargetFramework)' == 'net6.0'"/>`

I've been rethinking how to do the polyfills and am not convinced the current way is ideal (it proved harder than I wanted to update things to support builds for .net 6 and .net 7).

src/common/System/IO/StreamReadExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+                var bytesRead = source.Read(destination, totalRead, destination.Length - totalRead);
+                if (bytesRead == 0)
+                {
+                    break;
+                }
+
+                totalRead += bytesRead;
+            }
+#endif
+        }
```
twsouthwick, line 25:
```suggestion
#if !NET7_0_OR_GREATER
    internal static class StreamReadExtensions
    {
        public static void ReadExactly(this Stream source, byte[] destination)
        {
            var totalRead = 0;
            while (totalRead < destination.Length)
            {
                var bytesRead = source.Read(destination, totalRead, destination.Length - totalRead);
                if (bytesRead == 0)
                {
                    break;
                }

                totalRead += bytesRead;
            }
        }
#endif
```

No need for the extra redirection

twsouthwick (approved):

twsouthwick:
Thanks @f1nzer for the PR!

Issue #1209, opened 2022-09-24
ChangeDocumentType (.xlsm) to (.xlsx) not removing /xl/vbaProject.bin


Asbjoedt:
**Describe the bug**
Using ChangeDocumentType from MacroEnabledWorkbook (.xlsm) to Workbook (.xlsx) results in the error in Excel and spreadsheet cannot be opened (loosely translated from another language):
"The files does not have any macros, but it contains macro-based content"

When looking at the XML of the new .xlsx spreadsheet, the spreadsheet contains the file /xl/vbaProject.bin, which I think it should not. When removing this file, Excel will no longer give error preventing the user from opening the spreadsheet.

I use this code to change document type:

```
public void Convert_to_OOXML_Transitional(string input_filepath, string output_filepath)
{
    byte[] byteArray = File.ReadAllBytes(input_filepath);
    using (MemoryStream stream = new MemoryStream())
    {
        stream.Write(byteArray, 0, (int)byteArray.Length);
        using (SpreadsheetDocument spreadsheet = SpreadsheetDocument.Open(stream, true))
        {
            spreadsheet.ChangeDocumentType(SpreadsheetDocumentType.Workbook);
        }
        File.WriteAllBytes(output_filepath, stream.ToArray());
    }
}
```

I can fix the problem with the following code:

```
public void Remove_VBA(string filepath)
{
    using (SpreadsheetDocument spreadsheet = SpreadsheetDocument.Open(filepath, true))
    {
        VbaProjectPart vba = spreadsheet.WorkbookPart.VbaProjectPart;
        if (vba != null)
        {
            spreadsheet.WorkbookPart.DeletePart(vba);
        }
    }
}
```

The problem can also relate to MacroEnabledTemplate (.xltm) spreadsheets but I have not tested this.

Even though I can fix the problem in my code, I think this should be fixed in Open XML SDK, so I file this issue.

**Screenshots**

**To Reproduce**
1. Run the ChangeDocumentType code on my linked data sample
3. Run the fix to see if it works

**Observed behavior**
VBA project file /xl/vbaProject.bin is not removed from spreadsheet when using ChangeDocumentType from .xlsm to .xlsx. The problem could also relate to MacroEnabled Temaplte (.xltm), but I have not checked this.

**Expected behavior**
It should be removed, because Excel gives error when trying to open spreadsheet that has been changed from .xlsm to .xlsx. When removing /xl/vbaProject.bin Excel does no longer give error.

[...]

PR #1263, opened 2022-12-12
Make IDocumentTypeFeature public

twsouthwick:
Enables #1209

### DIFF
--- a/src/DocumentFormat.OpenXml/Features/IDocumentTypeFeature.cs
+++ b/src/DocumentFormat.OpenXml/Features/IDocumentTypeFeature.cs
@@ -3,10 +3,21 @@
 
 namespace DocumentFormat.OpenXml.Features;
 
-internal interface IDocumentTypeFeature<T>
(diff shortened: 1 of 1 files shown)


Asbjoedt:
Hi @twsouthwick 

A question for my understanding.

Why are you replacing ```System.Enum``` with ```Struct```?

Here's the enums for ```SpreadsheetDocumentType```:
https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheetdocumenttype?view=openxml-2.8.1

Are they obsolete now?

twsouthwick:
It's future proofing. I'm thinking about a generalized document type to help with packages in general - the constraint just means that we aren't limited to using an enum, but any struct.

twsouthwick:
@Asbjoedt this is now merged into the v3.0.0 version. There is a CI feed available to try this out: https://openxmlsdk.blob.core.windows.net/feed/index.json

Issue #1281, opened 2022-12-28
How to open SpreadsheetDocument without calcchain.xml


tomjebo:
### Discussed in https://github.com/OfficeDev/Open-XML-SDK/discussions/1279

<div type='discussions-op-text'>

<sup>Originally posted by **Distancehe** December 22, 2022</sup>
The Excel file itself do not contain calcchain.xml, but calcchain.xml is referenced from workbook.xml.rels as below snapshot. When trying to load the excel with SpreadsheetDocument.Open(), it will throw exception since the calcchain.xml is missing. The question is, how to open the excel file with calcchain.xml missing? Is there some way to ignore this and continue open the file ? Thanks !

![image](https://user-images.githubusercontent.com/4394577/209112903-02252d47-95a6-4be0-a404-b3ff1ca45c63.png)

</div>

PR #1288, opened 2023-01-03
fix for issue1281, ignore missing calcChain part

tomjebo:
This is the simple approach of ignoring the missing part but leaving the relationship entry alone. Excel will handle this but if we feel it's better to remove the relationship, I can rework the fix.

### DIFF
--- a/src/DocumentFormat.OpenXml.Framework/Packaging/OpenSettings.cs
+++ b/src/DocumentFormat.OpenXml.Framework/Packaging/OpenSettings.cs
@@ -10,9 +10,7 @@ namespace DocumentFormat.OpenXml.Packaging
     /// </summary>
     public class OpenSettings
     {
-        private bool? _autoSave;
         private MarkupCompatibilityProcessSettings? _mcSettings;
-        private bool? _ignoreExceptionsOnCalcChainPartMissing;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="OpenSettings"/> class.
@@ -40,11 +38,7 @@ namespace DocumentFormat.OpenXml.Packaging
         /// Gets or sets a value indicating whether to auto save document modifications.
         /// The default value is true.
         /// </summary>
-        public bool AutoSave
-        {
-            get => _autoSave ?? true;
-            set => _autoSave = value;
-        }
+        public bool AutoSave { get; set; } = true;
 
         /// <summary>
         /// Gets or sets the value of the markup compatibility processing mode.
@@ -53,10 +47,7 @@ namespace DocumentFormat.OpenXml.Packaging
         {
             get
             {
-                if (_mcSettings is null)
-                {
-                    _mcSettings = new MarkupCompatibilityProcessSettings(MarkupCompatibilityProcessMode.NoProcess, FileFormatVersions.Office2007);
-                }
+                _mcSettings ??= new MarkupCompatibilityProcessSettings(MarkupCompatibilityProcessMode.NoProcess, FileFormatVersions.Office2007);
 
                 return _mcSettings;
             }
@@ -84,10 +75,6 @@ namespace DocumentFormat.OpenXml.Packaging
         /// Gets or sets a value indicating whether to ignore an exception if the calcChain part is missing.
         /// The default value is false which means missing calcChain part will throw an exception upon package open.
         /// </summary>
-        public bool IgnoreExceptionOnCalcChainPartMissing
-        {
-            get => _ignoreExceptionsOnCalcChainPartMissing ?? false;
-            set => _ignoreExceptionsOnCalcChainPartMissing = value;
-        }
+        public bool IgnoreExceptionOnCalcChainPartMissing { get; set; }
     }
 }
--- a/src/DocumentFormat.OpenXml.Framework/Packaging/OpenSettings.cs
+++ b/src/DocumentFormat.OpenXml.Framework/Packaging/OpenSettings.cs
@@ -12,6 +12,7 @@ namespace DocumentFormat.OpenXml.Packaging
     {
         private bool? _autoSave;
         private MarkupCompatibilityProcessSettings? _mcSettings;
+        private bool? _ignoreExceptionsOnCalcChainPartMissing;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="OpenSettings"/> class.
@@ -32,6 +33,7 @@ namespace DocumentFormat.OpenXml.Packaging
             MarkupCompatibilityProcessSettings.TargetFileFormatVersions = other.MarkupCompatibilityProcessSettings.TargetFileFormatVersions;
             MaxCharactersInPart = other.MaxCharactersInPart;
             RelationshipErrorHandlerFactory = other.RelationshipErrorHandlerFactory;
+            IgnoreExceptionOnCalcChainPartMissing = other.IgnoreExceptionOnCalcChainPartMissing;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


tomjebo:
fixes #1281

src/DocumentFormat.OpenXml.Framework/Packaging/OpenXmlPartContainer.cs
```diff
@@ -1794,10 +1794,17 @@ internal OpenXmlPart CreateOpenXmlPart(string relationshipType)
         /// <param name="loadedParts">Temp collection to detect loaded (shared) parts.</param>
         internal void LoadReferencedPartsAndRelationships(OpenXmlPackage openXmlPackage, OpenXmlPart? sourcePart, RelationshipCollection relationshipCollection, Dictionary<Uri, OpenXmlPart> loadedParts)
         {
-            foreach (var relationship in relationshipCollection)
+            List<string> partsToIgnore = new()
             {
                 // Fix bug https://github.com/OfficeDev/Open-XML-SDK/issues/1205
-                if (relationship.RelationshipType == @"http://schemas.microsoft.com/office/2006/relationships/recovered")
+                @"http://schemas.openxmlformats.org/officeDocument/2006/relationships/calcChain",
```
twsouthwick, line 1800: Is this something that is always missing? are there cases in which we wouldn't want to skip it?

mikeebowen (dismissed):
Looks good to me

twsouthwick:
This should be stored in either OpenSettings or exposed as a feature (these concepts will probably be reconciled at some point, but OpenSettings allows you to set things that occur at opening). Should users be able to override this? Is there a use case for that?

src/DocumentFormat.OpenXml.Framework/Packaging/OpenXmlPartContainer.cs
tomjebo: @twsouthwick  read the comments on #1279, the Excel team and I discussed this. For opening a package with the SDK, it's safe to skip this regardless of whether the relationship is valid (i.e. has a valid part) or not. Excel will always create the relationship and part as needed and doesn't rely on reading it.

tomjebo:
I don't see a need to make this an OpenSettings options. Users don't need to override the behavior.

src/DocumentFormat.OpenXml.Framework/Packaging/OpenSettings.cs
```diff
@@ -12,6 +12,7 @@ public class OpenSettings
     {
         private bool? _autoSave;
         private MarkupCompatibilityProcessSettings? _mcSettings;
+        private bool? _ignoreExceptionsOnCalcChainPartMissing;
```
mikeebowen, line 15: This line is unnecessary

src/DocumentFormat.OpenXml.Framework/Packaging/OpenSettings.cs
```diff
@@ -77,5 +79,15 @@ public MarkupCompatibilityProcessSettings MarkupCompatibilityProcessSettings
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to ignore an exception if the calcChain part is missing.
+        /// The default value is false which means missing calcChain part will throw an exception upon package open.
+        /// </summary>
+        public bool IgnoreExceptionOnCalcChainPartMissing
+        {
+            get => _ignoreExceptionsOnCalcChainPartMissing ?? false;
+            set => _ignoreExceptionsOnCalcChainPartMissing = value;
+        }
```
mikeebowen, line 91: this could be written more succinctly:

```csharp
public bool IgnoreExceptionOnCalcChainPartMissing { get; set; } = false;
```

mikeebowen (changes requested):

src/DocumentFormat.OpenXml.Framework/Packaging/OpenSettings.cs
tomjebo: I was just following the existing pattern (i.e. AutoSave). It's probably more readable but we can use either.

mikeebowen: I prefer using the most modern syntax available. Maybe refactor the autosave to use the c# 6 syntax too

src/DocumentFormat.OpenXml.Framework/Packaging/OpenSettings.cs
tomjebo: again, following the same pattern as _autoSave. Might be better to leave them for readability but will wait for more comments.

mikeebowen: Unless there are reasons not to, I would remove both private fields and update both to the new syntax

tomjebo: VS gives other suggestions for conciseness (i.e. using compound statements) as well for existing code, like the if statement in MarkupCompatibilityProcessSettings.

mikeebowen: I don't know all of what VS suggests, but I would be on board with modernizing syntax in general

Distancehe (dismissed):

tomjebo:
@twsouthwick any thoughts on this openSettings approach?

src/DocumentFormat.OpenXml.Framework/Packaging/OpenSettings.cs
tomjebo: I updated the syntax

tomjebo:
@Distancehe just to recap the functionality here as it may have gotten a little lost in the thread. Excel effectively ignores 3rd party authoring of the calcChain part (whether intact or broken) and will rewrite on save a new or updated part. With this change, the SDK's default behavior will still be to throw an exception on load if the calcChain relationship  is broken, including the part being absent. I felt that this would be desirable for 3rd party producers using the SDK if they actually had some purpose for writing calcChain parts for their own use. This default behavior will do the same processing for calcChain that it does for any other part being loaded. Otherwise, the behavior is overridable with the OpenSettings IgnoreExceptionOnCalcChainPartMissing property. Overriding causes the SDK load path (i.e. SpreadsheetDocument.Open) to completely ignore the calcChain relationship, leaving the package as is wrt that part.

Everyone, please post here if you have any concerns with this approach.

src/DocumentFormat.OpenXml.Framework/Packaging/OpenSettings.cs
```diff
@@ -84,10 +75,6 @@ public MarkupCompatibilityProcessSettings MarkupCompatibilityProcessSettings
         /// Gets or sets a value indicating whether to ignore an exception if the calcChain part is missing.
         /// The default value is false which means missing calcChain part will throw an exception upon package open.
         /// </summary>
-        public bool IgnoreExceptionOnCalcChainPartMissing
-        {
-            get => _ignoreExceptionsOnCalcChainPartMissing ?? false;
-            set => _ignoreExceptionsOnCalcChainPartMissing = value;
-        }
+        public bool IgnoreExceptionOnCalcChainPartMissing { get; set; }
```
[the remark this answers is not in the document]
mikeebowen: You are correct, it doesn't need `= false;` The default value for a bool is false: [bool (C# Reference)](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/bool)

mikeebowen (approved):
LGTM

mikeebowen (approved):
LGTM

PR #1293, opened 2023-01-06
test for missing calcChain part behavior

tomjebo:
This test verifies the fix for #1281. 

Question: should this test be broken into two tests, one for the desired exception and one for the non-exception case?


mikeebowen (dismissed):
Looks good to me

mikeebowen:
Do you mean add a test with `IgnoreExceptionOnCalcChainPartMissing = false`?

tomjebo:
No, both tests (true and false) are in the one function. I was asking if we should put them in separate test functions.

twsouthwick:
They should probably be in separate functions. Makes it easier to diagnose failures

twsouthwick:
btw - I'm not sold on the API design with IgnoreExceptionOnCalcChainPartMissing. I have some ideas I'll noodle on, but may main concern is that (1) it doesn't allow for any additional behavior such as  if we want to remove it, modify it, etc and (2) it is spreadsheet focused. We don't have a better way of handling this at the moment, but would like to see a better design by the time we ship.

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
```diff
@@ -301,5 +304,31 @@ public void TestOpenModel3DWrittenByPowerPoint_DashMime()
...
             Assert.Equal("model/gltf-binary", model3DReferenceRelationshipPart.ContentType);
         }
+
+        // Test issue #1281 for regressions.
+        // When opening a workbook (SpreadsheetDocument.Open) with a missing calcChain part and default OpenSettings
+        // specified, we should throw an exception.
+        [Fact]
+        public void OpenWithMissingCalcChainPart()
+        {
+            Stream stmSpd = GetStream(TestFiles.MissingCalcChainPart, false);
```
twsouthwick, line 314: need a using

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
```diff
@@ -301,5 +304,31 @@ public void TestOpenModel3DWrittenByPowerPoint_DashMime()
...
+            Action act = () => SpreadsheetDocument.Open(stmSpd, false);
```
twsouthwick, line 315: inline this

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
```diff
@@ -301,5 +304,31 @@ public void TestOpenModel3DWrittenByPowerPoint_DashMime()
...
+            InvalidOperationException exception = Assert.Throws<InvalidOperationException>(act);
+
+            Assert.Equal("Specified part does not exist in the package.", exception.Message);
```
twsouthwick, line 318: Should never assert of exception messages - if you need to, use the resource strings

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
```diff
@@ -301,5 +304,31 @@ public void TestOpenModel3DWrittenByPowerPoint_DashMime()
...
+
+            try
+            {
+                SpreadsheetDocument spd = SpreadsheetDocument.Open(stmSpd, false, new OpenSettings() { IgnoreExceptionOnCalcChainPartMissing = true });
```
twsouthwick, line 322: use a using rather than close

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
```diff
@@ -301,5 +304,31 @@ public void TestOpenModel3DWrittenByPowerPoint_DashMime()
...
+
+                Assert.NotNull(spd);
+
+                spd.Close();
+            }
+            catch (InvalidOperationException ex)
```
twsouthwick, line 328: Why not do this in an Assert.Throws?

twsouthwick (changes requested):

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
```diff
@@ -301,5 +304,31 @@ public void TestOpenModel3DWrittenByPowerPoint_DashMime()
 
             Assert.Equal("model/gltf-binary", model3DReferenceRelationshipPart.ContentType);
         }
+
+        // Test issue #1281 for regressions.
```
twsouthwick, line 308: Using issue numbers in code is not helpful - the name of the test should be sufficient. If you feel it's really needed, use a full URI to the issue so it can be found easier

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
tomjebo: or maybe just assert not null on the exception var? I really just want to verify the default behavior is to throw some exception for that package. I was hoping to be more specific but would it be acceptable to just assert that it throws?

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
tomjebo: I actually want to assert that it doesn't throw. but maybe I don't need to catch this, just let it throw?

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
tomjebo: added

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
tomjebo: added

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
tomjebo: proposed code added

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
tomjebo: added

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
tomjebo: prosed new code, no try/catch

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
tomjebo: removed

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
```diff
@@ -301,5 +304,29 @@ public void TestOpenModel3DWrittenByPowerPoint_DashMime()
 
             Assert.Equal("model/gltf-binary", model3DReferenceRelationshipPart.ContentType);
         }
+
+        // When opening a workbook (SpreadsheetDocument.Open) with a missing calcChain part and default OpenSettings
```
twsouthwick, line 308: These comments do not feel necessary

twsouthwick (dismissed):
LGTM modulo the extra comments seem unnecessary - the test case is pretty straightforward

test/DocumentFormat.OpenXml.Packaging.Tests/OpenXmlPackageTests.cs
tomjebo: I'll leave them for now since I'll probably be back in this code again soon.

PR #1295, opened 2023-01-10
Add a package abstraction for System.IO.Packaging

twsouthwick:
The Package and PackagePart types from System.IO.Packaging are not built in a way to make composition easy; they do not expose public overridable methods. These abstractions will help with smoothing over some issues found in those types that are difficult to manage.


lindexi (commented):
Awesome, thank you

src/DocumentFormat.OpenXml.Framework/Packaging/OpenXmlPackage.cs
```diff
@@ -62,11 +60,20 @@ private protected OpenXmlPackage(Package package, OpenSettings? settings = null)
...
+            var feature = new PackageImpl(package);
+
+            Features.Set<IPackageFeature>(feature);
+
+            RegisterOnClose(package.Close);
+
+            Load(feature);
         }
 
+        private List<Action>? _onClose;
```
lindexi, line 73: The `Action` is the delegate type. And we can use `+=` to add the new action. It is fast and thread safe. And I think it better than the `List<Action>` type.

```suggestion
        private Action? _onClose;
        internal void RegisterOnClose(Action action) => _onClose += action;
```

src/DocumentFormat.OpenXml.Framework/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -76,7 +79,7 @@ protected XDocument ToFlatOpcDocument(XProcessingInstruction? instruction)
                 new XElement(
                     Pkg + "package",
                     new XAttribute(XNamespace.Xmlns + "pkg", Pkg.ToString()),
```
mikeebowen, line 81: Would it be better to use `String.Concat()` here instead?

src/DocumentFormat.OpenXml.Framework/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -127,7 +130,7 @@ private static XElement GetBinaryPartContentsAsXml(PackagePart part)
                 new XElement(Pkg + "binaryData", ToChunkedBase64String(part)));
```
mikeebowen, line 130: Would it be better to use `String.Concat()` here instead?

mikeebowen (approved):
Looks good, just a small question about `String.Concat()` vs `+`

lindexi (commented):
Looks good, thank you

src/DocumentFormat.OpenXml.Framework/Packaging/OpenXmlPackage.FlatOpc.cs
twsouthwick: Probably - there are many issues with the FlatOpc implementations regarding performance... this PR was focused on replacing things with the ipackage rather than fixing anything else.

michaelcsikos:
How do I do this now that `Package` is `Obsolete`?
```cs
var newId = (uint)(spreadsheetDocument.Package.GetRelationships().Count() + 1);
var rId   = "relId" + newId;

...

spreadsheetDocument.Package.CreateRelationship(uri,
                                               TargetMode.Internal,
                                               "http://schemas.openxmlformats.org/officeDocument/2006/relationships/worksheet",
                                               rId);
```

It would be nice if the `Obsolete` message pointed us in the right direction. I'm rolling back to `2.19.0` for now.

twsouthwick:
The obsolete has been obsoleted :) I heard enough feedback of people using this that I've reverted that. However, in v3.0, it's a new type with similar shape that should be a drop in replacement (the System.IO.Packaging types are just not as composable as needed)

twsouthwick:
For now, we're not planning on releasing a v2.20.1 - this is mostly because we had to move to GH actions during this timeframe and we haven't back ported those changes to that branch. If having a v2.20.1 with the obsolete removed would be helpful for migrating to v3.0, please thumbs up this and we can prioritize it.

XavierNV:
Thanks a lot for your feedback. We'll wait for v3.0 final version...

PR #1330, opened 2023-01-29
Move relationship handling out of package

twsouthwick:
The SDK handled Strict->Transitional namespace conversion automatically,
but in a way that forces it to be continuously tracked. This has been
moved into a relationship filter that will handle this a single time and
then it doesn't need to be calculated again.


twsouthwick:
@Asbjoedt I believe this may help towards some of your desire to control more the strict/transitional stuff. I'm trying to add hooks into the sdk to centralize how this stuff is managed and this is a step in that direction that may apply to you.

Note: I'm not ready to make the APIs for these things public, but feel free to try a build with an internalsvisibleto attribute for your project and see what you can do.

Asbjoedt:
@twsouthwick  Cool!

Issue #1313, opened 2023-01-18
Split assembly in separate packages for DOCX / XLSX / PPTX


m-gallesio:
Per https://github.com/dotnet/Open-XML-SDK/issues/387
> Consider breaking up library into smaller ones for Word/Powerpoint/Excel.

As well as my comment: https://github.com/dotnet/Open-XML-SDK/issues/387#issuecomment-1140978211
> Yes, please. Having these separated (along with a base / core module, I assume) would improve clarity, make for a smaller base package (currently ~5.8 MB) and allow more code splitting between dependent assemblies. For reference our product does not currently use Powerpoint at all.

I know this topic is in the backlog, but considering its structural impact I think it it worth it to have a dedicated issue, at least to improve its visibility and prevent it from getting buried.

(I have searched for similar issues but if I could not find any; if this is indeed explicitly tracked elsewhere please close / delete this issue).

twsouthwick:
@m-gallesio this has been a long time in the making. I've currently separated out the core SDK infrastructure and enabled the typed classes to be in a separate assembly from all of that. I'm currently working on stabilizing it, but as that occurs, this becomes possible.

@tomjebo This is something we should start looking at from the schema perspective. Can we update the generator to keep these isolated? I know there are a few types that are explicitly shared between the types for some reason (I think similar names) that we would need to separate out.

twsouthwick:
@m-gallesio @czemacleod If there's interest here, the code generator has been moved to this repo. It may be able to isolate the types required for each doc type using some sort of tree-walking. Happy to take any pull requests or have a discussion of how to start tackling this if anyone is interested in contributing!

CZEMacLeod:
@twsouthwick I'm not sure I really have time to delve that far into this project unfortunately. 
My reasoning is that we use this to build Excel files for reports and 'datagrids' in our web applications (as an export to Excel feature). We also use it in a couple of applications to do data import from Excel. 
The point is that I have no need to any other types (other than common/base types) outside that scope and if we could reduce the memory footprint of the application all the better.
Not for ourselves especially, but this would also make a lot of sense in a microservices scenario for e.g. generating word documents from templates, or excel spreadsheets or something similar.

twsouthwick:
I've added this to the v3.0 milestone so it can be tracked there - otherwise, it will probably be until v4.0 (whenever that ends up happening)

m-gallesio:
[...]

PR #1333, opened 2023-01-31
API Rename to match XML ns

tomjebo:
this is a WIP. some samples not building still.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/Open-XML-SDK?pullRequest=1333) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/Open-XML-SDK?pullRequest=1333) before we can accept your contribution.<br/><br/>:x: tomjebo [sign now](https://cla.dotnetfoundation.org/dotnet/Open-XML-SDK?pullRequest=1333)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/Open-XML-SDK?pullRequest=1333) it.</sub>

tomjebo:
I need to address the case on some names, specifically things like SpreadsheetML vs SpreadSheetML (the second "s" vs "S"). Which should we use? 

Also the ML (for Markup Language) where DocumentFormat.OpenXml uses lower case for the ml after X. Should we use Ml instead of ML? Acronyms in general, should we capitalize them or lower case like the SDK's name does?

Otherwise, the renaming is done.

tomjebo:
this renaming will cause a lot of using and scoping changes in just about any project using the Open-XML-SDK. 
An alternative approach could be to limit the namespace changes to Office 2013 and later, leaving the Office 2007 and 2010 namespaces alone. Beyond 2013, we mainly see extensions and new features which still means a lot of code changes for consuming projects but not nearly as many as when we change (effectively) all the SDK namespaces.

twsouthwick:
@tomjebo let's see if we can refactor things so that the namespace generation isn't done on the backend, but is configurable in the source generator here. It'll probably require some changes to how we structure the data files, but then we can look at enabling this change in some new Word/Excel/PPT assemblies (see https://github.com/dotnet/Open-XML-SDK/issues/1313#issuecomment-1462991520)

tomjebo:
> @tomjebo let's see if we can refactor things so that the namespace generation isn't done on the backend, but is configurable in the source generator here. It'll probably require some changes to how we structure the data files, but then we can look at enabling this change in some new Word/Excel/PPT assemblies (see [#1313 (comment)](https://github.com/dotnet/Open-XML-SDK/issues/1313#issuecomment-1462991520))

I like it. I think we would mainly need to add the ns URI and the short ns name to the json stuff going to the front end, then on the front end have the mapping code do what NamespaceMappingData.xml and cs do now on the backend.

tomjebo:
Sorry, the ns URI and prefixes are already on the front end. It's just the mapping of the ns shortnames to these.

twsouthwick:
@tomjebo  I'm going to close this for now. If you identify things on the backend/sdk side that can be done feel free to open new issues/prs for those

PR #1345, opened 2023-02-25
Convert LINQ generator to run as source generator

twsouthwick:
This change integrates the LINQ generator with the source generator infrastructure so things will stay up-to-date as well as use a single source of truth.


twsouthwick:
FYI @ThomasBarnekow - there are a few differences still that I'm trying to track down, but overall looks to be very similar, but running in a source generator so it'll get updated automatically

twsouthwick:
This is needed for https://github.com/dotnet/Open-XML-SDK/issues/1349

twsouthwick:
@ThomasBarnekow this over all looks like it gives the same output as you had generated, but it is missing some of the documentation (especially for children/parents/siblings of `NoNamespace`).

I'd like to get this in, but don't want to break anything that you may need here so would like to have your sign off or feedback

ThomasBarnekow:
@twsouthwick, let me look into this.

Do you know why the documentation is missing?

twsouthwick:
After looking at this, I think it's because at runtime, you'll get attributes/children from the base classes, but the `OpenXmlGeneratorContext` by default does not do this.

twsouthwick:
I've verified that a random sampling indeed shows that the failure for the comments to show up is because it is a derived class. I'm going to go ahead and merge this and open a tracking issue since this affects comments and is blocking some updates to better enable AOT scenarios.

PR #1352, opened 2023-03-02
Fix part extension lookup to set the correct found extension. Fixes #1305

### LINKED ISSUE #1305: New ImagePart is created as .bin instead of real extension
**Describe the bug**
I have asked the question on Stackoverflow:
https://stackoverflow.com/questions/75100804/embedding-image-as-new-imagepart-creates-a-bin-file-using-open-xml-sdk

I don't expect to get any answer there, and I am also increasingly curious whether, this is an implementation error but an issue in the SDK.

Issue is, I create new embedded image in Excel and it is saved to ```/xl/media/image.[WhatEverExtensionItIs]```
If I create a new ImagePart with the SDK and feed it data like this:

```
ImagePart new_ImagePart = worksheetPart.AddImagePart(ImagePartType.Tiff);
stream.Position = 0;
new_ImagePart.FeedData(stream);
```

Then the file will be saved as ```/xl/media/image.bin```.

This behaviour it should not do. It should write the extension as .tiff, since this is the ```ImagePartType```, that I specify. Or secondly it should read the stream and identify it as .tiff through i.e. a magic number.

If you unzip the package and change the image extension to real extension, then the image will open perfectly and I have used a TIFF validator to confirm the image is well-formed. 

What do you think of this? Am I doing anything wrong in my implementation, or can we change the SDK to write real extension like Excel does?

**Screenshots**

**To Reproduce**
1. Use the supplied code to stream any TIFF image (or change ImagePartType to any other image type) to a new ImagePart
2. Unzip the spreadsheet and go to /xl/media/ to find the file as a .bin
3. Change extension to .tiff and the image will open perfectly

**Observed behavior**
ImagePart creates a new image with .bin extension

**Expected behavior**
ImagePart should create a new image with extension based on ImagePartType.

 - OS: Windows 11
 - Office version: 2212
 - .NET Target: .Net latest version
 - DocumentFormat.OpenXml Version: 2.19

mikeebowen:
This fixes #1305


twsouthwick (approved):
Should this be back ported for 2.20?

mikeebowen:
> Should this be back ported for 2.20?

Probably, v2 has the same bug and this fix won't cause any breaking changes.

twsouthwick:
I'd like to see a more targeted unit test for this in the main branch. Happy to join up for an hour to pair program this if you want and I can give you some pointers on what to mock out.

twsouthwick:
This fix is needed in v2.20 before release

.github/workflows/codeql-analysis.yml
```diff
@@ -39,11 +39,14 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
-
+      uses: actions/checkout@v3
```
twsouthwick, line 42: why are these changes being added here?

twsouthwick (changes requested):
Looks like your merge is off....

src/DocumentFormat.OpenXml.Framework/Packaging/OpenXmlPackage.FlatOpc.cs
```diff
@@ -0,0 +1,394 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using DocumentFormat.OpenXml.Features;
```
twsouthwick, line 4: This file is something you're adding - did you mean to do so?

test/DocumentFormat.OpenXml.Framework.Tests/OpenXmlPartTests.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Threading.Tasks;
+using Xunit;
+
+namespace DocumentFormat.OpenXml.Framework.Tests
+{
+    public class OpenXmlPartTests
+    {
+
+        [Fact]
+        public void ExtensionTest()
```
twsouthwick, line 18: Can you devise some permutations of things here to validate expected behavior? i.e. what if there is no extension defined or if IPartExtensionFeature has an extension defined on it?

twsouthwick (changes requested):

test/DocumentFormat.OpenXml.Framework.Tests/OpenXmlPartTests.cs
mikeebowen: I added tests for no extension defined,  extension defined on IPartExtensionFeature, and IsContentTypeFixed == true

mikeebowen:
closing this because of pipeline issues

PR #1354, opened 2023-03-04
Update LINQ files

twsouthwick:
This is mostly so that when this is converted to a generator it doesn't have the following changes:

- Additional namespaces/elements
- Source generators have the UTF8-BOM appended


twsouthwick:
This is to minimize non-generator related changes in #1345

PR #1359, opened 2023-03-20
Office schema file update March 2023 release

mikeebowen:


tomjebo (commented):
@mikeebowen 2.20.3.6 CT_TaskHistoryEvent in [MS-PPTX] has only 11 children and 2 attributes but this shows 14 children. Can we double check on this or is the doc just lagging behind? I notice that there are CT_Empty typed children, exactly three. Perhaps that that's the problem? Maybe we should remove those? We should check.

CHANGELOG.md
```diff
@@ -37,6 +40,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [2.20.0]
 
+### Added
```
twsouthwick, line 43: nit: this change doesn't add it to v2.20

twsouthwick (dismissed):

twsouthwick:
@mikeebowen looks like you need to merge (not sure why GH isn't offering that)

PR #1367, opened 2023-03-29
Use OrdinalIgnoreCase in LINQ generator

twsouthwick:
Without this, the current culture is used. Since .NET Framework and .NET Core use different data sources for culture sorting, this caused differences from the CLI and VS. This should keep the generated source consistent across toolchains.


tomjebo (approved):

mikeebowen (approved):
Looks good

PR #1371, opened 2023-03-29
obsolete OpenXmlPackage.Close and remove use of it in sample and tests.

tomjebo:


samples/PowerPointModernCommentSample/Program.cs
```diff
@@ -38,64 +38,63 @@ public static void Main(string[] args)
...
-
-                // create missing PowerPointAuthorsPart if it is null
-                if (presentationDocument?.PresentationPart?.authorsPart is null)
-                {
-                    presentationDocument?.PresentationPart.AddNewPart<PowerPointAuthorsPart>();
-                }
-
-                // Add missing AuthorList if it is null
-                if (presentationDocument?.PresentationPart?.authorsPart?.AuthorList is null)
+                using (PresentationDocument presentationDocument = PowerPointUtils.CreatePresentation(fileInfo.FullName))
```
twsouthwick, line 41: you can replace this with `using PresentationDocument presentationDocument =...;` which doesn't require shifting all the lines

twsouthwick (approved):
LGTM

samples/PowerPointModernCommentSample/Program.cs
twsouthwick: don't worry about it here, though. The change on main branch should do it though

PR #1380, opened 2023-04-03
User correct file extension with images

### LINKED ISSUE #1305: New ImagePart is created as .bin instead of real extension

mikeebowen:
This fixes #1305


twsouthwick (approved):
LGTM. Just to verify, if you revert line 120 in OpenXmlPart, does at least 1 test start failing? That's to catch any future regressions in the behavior.

mikeebowen:
Yes, ExtensionTest and ExtenstionTestKnownContentType both fail if line 120 in OpenXmlPart is reverted.

PR #1387, opened 2023-04-05
Allows for AddEmbeddedPackagePart to use the correct extensions for contenttype

tomjebo:


mikeebowen (approved):

twsouthwick:
@tomjebo this appears to be something you could do as an extension method instead. It's fine on the v2.20 (it's shipped anyway), but for main, can you rework this as an extension method?

PR #1394, opened 2023-04-07
refactor addembeddedpackage and object methods to use one extension method 

tomjebo:
First draft WIP, some names will change after looking at the ImagePart extension and other that get generated in parts.


mikeebowen (dismissed):
Many of the .cs files have empty catch blocks in `src/DocumentFormat.OpenXml/Packaging/`. Are they supposed to eat any exceptions?

tomjebo:
> Many of the .cs files have empty catch blocks in `src/DocumentFormat.OpenXml/Packaging/`. Are they supposed to eat any exceptions?

Yes, they catch the `NotSupportedException` from the `GetCustomXmlPartTypeFromContentType` and return `false` by design. These are Try... variants.

twsouthwick (changes requested):
This is a lot of duplicated code. The struct direction we discussed seems worthwhile to explore, but maybe in the meantime switch ImagePartTypeInfo to an instance class that you can have an inteface that could be queried?

tomjebo:
@twsouthwick I consolidated the Add*Part methods using your PartTypeInfo struct and one OpenXmlPartExtension.InitPart method. I left the extensions for AddImagePart that were just doing the get part by extention (that's only for .tif vs .tiff). Perhaps that can be worked in some other way. I wasn't quite sure about how to use an interface in that situation.

twsouthwick:
Can you do an initial pr to revert what you added before? Then a second one to see what is being added to solve the scenario?

tomjebo:
Closing this PR in favor of #1405

PR #1399, opened 2023-04-14
Allow null values via annotation to operator overloads able to handle null

lahma:
While investigating enabling nullable annotations in [ClosedXml](https://github.com/ClosedXML/ClosedXML) I came across some problems with code like so (simplified and manually created to demonstrate):

```c#
var properties = new SheetFormatProperties();
DoubleValue? openXmlComparableSimpleValue = properties.DefaultRowHeight;
if (openXmlComparableSimpleValue != null)
{
    return;
}
```

Gives: `Possible null reference argument for parameter 'left' in 'DocumentFormat.OpenXml.OpenXmlComparableSimpleValue<T>.operator !='` 

I went through operator overloads and annotated ones allowing null values with `?` when input as null is handled properly.


twsouthwick (approved):
Thanks for the fix! LGTM

PR #1417, opened 2023-04-24
Tif and Jp2 ImagePartTypes

Asbjoedt:
Hi!

I commited this approved PR: https://github.com/dotnet/Open-XML-SDK/pull/1304
The intention was to support ".tif" extension.

I have since recognized the PR did not create support for ".tif" ImagePartType extension properly. The fix supported ".tif" by mapping it to ".tiff" extension.

This is not the right way of doing it. ".tif" should be independently supported, because some organizations may want to only support one of these extensions to reduce complexity (despite the TIFF standard fully supports both extensions).

This new PR fully creates ".tif" as an independent ImagePartType alongside ".tiff".

This PR also adds JPEG-2000 ImagePartType extension ".jp2". Many archives accept JPEG-2000 and TIFF images for long-term preservation.

I try to PR into main, because it is a small fix, that does not break anything or change anything significantly, and I don't really know which other branch to PR into.

PS. I had to close another malformed PR and fork the repo anew, to do this properly, to cancel out the 1304 unintended recommit, which I had put into my own main instead of a branch, when I did that awhile ago.


twsouthwick:
@Asbjoedt this generally looks good, but need the CLA signed (we've moved to the dotnet org) and the tabs seem to be messed up.

Asbjoedt:
@dotnet-policy-service agree

@twsouthwick done.

twsouthwick (approved):
LGTM, although what's the difference between Tif and Tiff?

Asbjoedt:
The scenario
- I use .getStream() of ImagePart
- I use ImageMagick to convert the MemoryStream to **MagickFormat.Tif**
- I create new ImagePart and read the converted MemoryStream into and choose **ImagePartType.Tiff**
- I change relationships of the associated Blip to new ImagePart and delete the old ImagePart
- The new embedded object is, however, saved to the package with **.tiff** extension, but I want it to be saved with **.tif** extension
- Changing ImagePartType to **ImagePartType.Tif** should solve this. Is this true?

twsouthwick:
@tomjebo what are your thoughts here on the extra 'f' here?

tomjebo:
> @tomjebo what are your thoughts here on the extra 'f' here?

I noticed this in our code (tiff and tif extensions) and I think @mikeebowen added or worked on that. So I'll let him comment on whether that's the right way to go. I assumed the code is in place because .tif and .tiff are equivalent and therefore the part type can be shared. Perhaps that's wrong?

tomjebo (commented):
jp2 is an addition. But .tif -> image/tif would be a change from the albeit short lived double mapping. Does this break folks?

tomjebo:
i guess if it's breaking we can just include it into the v3.0.0 beta. So I'll approve also.

twsouthwick:
@tomjebo looks like this is purely additive, so probalby doesn't break anyone

tomjebo:
@twsouthwick  well, it changes the behavior not the API signature. previously if you add a data file of extension .tif, the actual part added in the package would be .tiff. Now it's .tif as it probably should be. I don't know if that's going to break folks who relied on the previous behavior.

twsouthwick:
Feels like a fix for a bug (wrong content type)

PR #1425, opened 2023-04-28
Remove superflous AddImagePart method overload.

tomjebo:


twsouthwick (approved):
Let's remove it so we have that taken care of. I think we can come up with a trick similar to what we did for adding parts that will allow adding by extension

PR #1434, opened 2023-05-16
update GitVersion.yml

mikeebowen:
With this configuration if a prerelease release is created from a branch named `release/<release num>`, then the release will be named `<some release num>-beta<a number>` where "some release num" is the value of `next-version` in GitVersion.yml and "a number" is the number of commits since the version source.

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,5 +1,10 @@
 mode: ContinuousDeployment
 next-version: 3.0.0
-branches: {}
+branches:
+  release:
+    regex: ^release?[/-]
+    tag: beta
+    increment: None
+    is-release-branch: true
 ignore:
   sha: []


GitVersion.yml
```diff
@@ -1,5 +1,10 @@
 mode: ContinuousDeployment
 next-version: 3.0.0
-branches: {}
+branches:
+  release:
+    regex: ^release?[/-]
```
twsouthwick, line 5: how would this work for branches we do work on such as the `release/v2.20.0` that was a WIP until we were ready to ship it?

mikeebowen: It would work the same. Any branch starting with "release/" that is used to create a pre-release will follow the "<some release num>-beta<a number>" pattern.

twsouthwick: what about non-prerelease? that is, we create a branch and do some work on it and then we want to release from it. Will this support creating a v2.20.0 with no beta moniker on a release/v2.20.0 branch?

mikeebowen: You would add something like this to the branches:
```yaml
preview:
  regex: ^preview?[/-]
  tag: feature # or whatever you want the tag to be
  increment: None
  is-release-branch: true
```

twsouthwick (approved):
LGTM I'm interested in how flexible it is (I haven't played with the configuration for a while), but like this direction

PR #1435, opened 2023-05-17
Implement package features on package instead of shared base

twsouthwick:
This will allow to reduce dependencies for each package type for AOT scenarios


twsouthwick:
A subsequent change will limit to only the parts necessary for the package type rather than all the parts

PR #1439, opened 2023-05-25
Move PartTypeInfo into framework library

twsouthwick:


twsouthwick:
@tomjebo  FYI - let me know if you have reason to keep it in the current package

PR #1476, opened 2023-06-29
Introduce equality comparers for OpenXmlElement

PhDuck:
Problem:
The current way to determine equality for two OpenXmlElements seems to
be x.OuterXml.Equals(y.OuterXml) (based on StackOverflow top answer).

This performs horribly due to allocation of strings and XmlWriter.

Solution:
This PR introduces OpenXmlElementEqualityComparer which supports equality determination.
Furthermore, the definition of equality is defined by OpenXmlElementEqualityOptions which allows for controlling how equality should be defined.

Benchmark results comparing to x.OuterXml.Equals(y.OuterXml):
```
BenchmarkDotNet=v0.13.2, OS=Windows 10 (10.0.19045.3086)
AMD Ryzen 7 5800X, 1 CPU, 16 logical and 8 physical cores
.NET SDK=7.0.304
  [Host] : .NET 7.0.7 (7.0.723.27404), X64 RyuJIT AVX2

Job=InProcess  Toolchain=InProcessEmitToolchain

|                       Method |              Mean |           Error |          StdDev |     Gen0 |     Gen1 |     Gen2 | Allocated |
|----------------------------- |------------------:|----------------:|----------------:|---------:|---------:|---------:|----------:|
|         OuterXmlLargeElement | 22,469,563.462 ns | 156,783.6460 ns | 130,921.4452 ns | 437.5000 | 281.2500 | 218.7500 | 7370176 B |
|           EqualsLargeElement |  9,235,127.232 ns | 118,532.7404 ns | 105,076.1870 ns |        - |        - |        - |      10 B |
| EqualsLargeElementSkipPrefix |  4,165,810.498 ns |  78,261.1757 ns |  76,862.9390 ns |        - |        - |        - |       5 B |
|             UnparsedOuterXml |          3.296 ns |       0.0140 ns |       0.0131 ns |        - |        - |        - |         - |
|               UnparsedEquals |          4.721 ns |       0.0149 ns |       0.0125 ns |        - |        - |        - |         - |
|                SmallOuterXml |      3,616.694 ns |      52.6133 ns |      43.9345 ns |   1.7662 |   0.0877 |        - |   29584 B |
|                  SmallEquals |        705.613 ns |       3.4249 ns |       3.2037 ns |   0.0010 |        - |        - |      24 B |
|        SmallEqualsSkipPrefix |        376.926 ns |       3.3287 ns |       2.9508 ns |   0.0014 |        - |        - |      24 B |
```


PhDuck:
@twsouthwick Your eyes on this PR would be much appreciated.

There is a few things that I would like some pointer with:
1. Ordering:
1.2. When unparsed, the order of children and attributes matter, but not when parsed.
2. Elements of type OpenXmlLeafTextElement seems to need special behaviour for comparing their text value. Other places these values seem to be placed as an attribute. I don't quite know why this is.
3. Attributes seems to always be of type OpenXmlComparableSimpleValue, so they support Equals. Can someone please confirm this.
4. Prefixes are quite expensive to lookup, so can we skip it and just compare Namespace.Uri, if so we can save ~40%.

src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
```diff
@@ -280,7 +280,7 @@ public bool HasAttributes
         /// <summary>
         /// Gets all extended attributes (attributes not defined in the schema) of the current element.
         /// </summary>
-        public IEnumerable<OpenXmlAttribute> ExtendedAttributes
+        public IReadOnlyList<OpenXmlAttribute> ExtendedAttributes
```
twsouthwick, line 283: Can't do this for two reasons:

1. Not supported in some targets (.net 3.5/4.0) that we support
2. (techincally for v3 we could but) it's a public API

twsouthwick: feel free to expose some of these internally as lists

twsouthwick:
@PhDuck Great questions. I'll answer them first, then a few thoughts at the end:

1. Ordering:
    1.1. For children it seems order is not defined, so order is non-stable.
	> This is actually not true, generally. Order of children are defined by their schema. However, for comparing, you should just iterate through the children and use the order that is there

    1.2. For attributes it seems their order is defined, so order is stable.
	> @tomjebo is this the case? We may want to have this as an option

    1.3 For extended attributes order is not defined, so ordering is non-stable.
	> We may want to include/exclude these

    1.4 For unparsed none of the above counts, since it depends fully on the inputs textual order.
 2. Elements of type OpenXmlLeafTextElement seems to need special behaviour for comparing their text value. Other places these values seem to be placed as an attribute. I don't quite know why this is.
 3. Attributes seems to always be of type OpenXmlComparableSimpleValue, so they support Equals. Can someone please confirm this.
	> Attributes are always OpenXmlSimpleType which has IEquatable I believe
 4. Prefixes are quite expensive to lookup, so can we skip it and just compare Namespace.Uri, if so we can save ~40% further,
	> Maybe have it as an option

With the points you're making, I think we may not want to implement equality directly on the element at this point, but rather provide helpers to create equality comparers that compare things people care about (you've pointed out questions that could be things people opt in/out of)

I'm thinking something like this:

```csharp
public static class OpenXmlElementComparers
{
	public static IEqualityComparer<OpenXmlElement> Default { get; }
	public static IEqualityComparer<OpenXmlElement> Create(OpenXmlElementEqualityOptions options);
}

public sealed class OpenXmlElementEqualityOptions
{
	// Include options
}
```

tomjebo:
> 1. 1.2. For attributes it seems their order is defined, so order is stable.
>    > @tomjebo is this the case? We may want to have this as an option

According to the XML specification order of attributes is not significant and therefore it shouldn't be for us as well. If we see a problem with this, definitely we want to investigate.

twsouthwick:
@PhDuck we've added GH action tests this week so if you merge in you'll be able to run tests

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementEqualityComparer.cs
```diff
@@ -0,0 +1,91 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+
+namespace DocumentFormat.OpenXml.Equality
```
twsouthwick, line 9: No need for a sub namespace here

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementEqualityOptions.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System;
+
+namespace DocumentFormat.OpenXml.Equality
+{
+    /// <summary>
+    /// Options defining the behaviour of equality for <see cref="OpenXmlElement"/>.
+    /// </summary>
+    [Flags]
+    public enum OpenXmlElementEqualityOptions
```
twsouthwick, line 12: I think I'd prefer this to be a class in case we find we want more fine grained control

src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
```diff
@@ -291,7 +292,7 @@ public IEnumerable<OpenXmlAttribute> ExtendedAttributes
                 }
                 else
                 {
-                    return Enumerable.Empty<OpenXmlAttribute>();
```
twsouthwick, line 294: Can you revert this?

src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
```diff
@@ -473,6 +474,227 @@ internal set
 
         #region public methods
 
+        private static bool MoveNextAndTrackCount(ref OpenXmlElementList.Enumerator e1, ref OpenXmlElementList.Enumerator e2, ref int e1ctr, ref int e2ctr)
```
twsouthwick, line 477: Does this need to be here or can it be in the equality comparer class?

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementEqualityComparer.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    /// <summary>
+    /// Equality comparer for determining value equality for <see cref="OpenXmlElement"/>.
+    /// </summary>
+    public sealed class OpenXmlElementEqualityComparer : IEqualityComparer<OpenXmlElement>
```
twsouthwick, line 14: Can this be a static class for now with internal class that implements the equality comparer?

twsouthwick (commented):
Looking good. Some of the changes won't work well with the multitargeting (I commented on a few things) - I'll kick the build off so we can see the results

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementEqualityOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// Options defining the behaviour of equality for <see cref="OpenXmlElement"/>.
+    /// </summary>
+    public sealed class OpenXmlElementEqualityOptions
+    {
+        /// <summary>
+        /// Gets a value indicating whether extended attributes should be considered when determining equality.
+        /// </summary>
+        public bool IncludeExtendedAttributes { get; init; } = true;
```
twsouthwick, line 14: inits will fail the CI build due to requiring PEVerify on .NET Framework builds for common scenarios. Means we can't use records and other types that use that.

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementEqualityOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether MC attributes should be considered when determining equality.
+        /// </summary>
+        public bool IncludeMCAttributes { get; init; } = true;
+
+        /// <summary>
+        /// Gets a value indicating whether namespace should alone be used when comparing idenity of elements, skipping prefix lookup.
+        /// </summary>
+        public bool CompareNamespaceInsteadOfPrefix { get; init; }
```
twsouthwick, line 24: I'm not sure I like the name of this... other ideas?

src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
```diff
@@ -473,6 +473,208 @@ internal set
 
         #region public methods
 
+        private bool PrefixAndQNameEqual(OpenXmlElement other, OpenXmlElementEqualityOptions options)
```
twsouthwick, line 476: I'd prefer these comparisons to be on the equality comparer itself - if you need any private properties, feel free to expose them as internal

twsouthwick (changes requested):

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementEqualityOptions.cs
PhDuck: Changed it to something else. If you also don't like that, please come up with one you like better :)

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementEqualityOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether mC attributes should be considered when determining equality.
+        /// </summary>
+        public bool IncludeMCAttributes { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether namespace should alone be used when comparing idenity of elements, skipping prefix lookup to improve performance.
+        /// </summary>
+        public bool SkipPrefixComparison { get; set; }
```
twsouthwick, line 24: I like this name better 👍

src/DocumentFormat.OpenXml.Framework/Equality/HashCode.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace DocumentFormat.OpenXml
+{
+    internal ref struct HashCode
+    {
```
twsouthwick, line 7: There's a shim for this already in the common directory and brought in via https://github.com/dotnet/Open-XML-SDK/blob/main/src/common/FrameworkShims.targets

PhDuck: It seems to me that the DocumentFormat.OpenXml project does not include these:

See [here](https://github.com/dotnet/Open-XML-SDK/blob/e7a0331218de7f53582698bdfd1ddcc2124c1a40/src/DocumentFormat.OpenXml/DocumentFormat.OpenXml.csproj#L8C48-L8C48)

twsouthwick: Hmm then for now include it manually and I'll clean it up

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementComparers.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// Gets the default equality comparer.
+        /// </summary>
+        public static readonly IEqualityComparer<OpenXmlElement> Default = GetEqualityComparer(new OpenXmlElementEqualityOptions());
+
+        /// <summary>
+        /// Gets the <see cref="IEqualityComparer{OpenXmlElement}"/> based on the given options./>
+        /// </summary>
+        /// <param name="openXmlElementEqualityOptions">The options defining equality.</param>
+        /// <returns></returns>
+        public static IEqualityComparer<OpenXmlElement> GetEqualityComparer(OpenXmlElementEqualityOptions openXmlElementEqualityOptions)
```
twsouthwick, line 23: Rename to Create

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementEqualityComparer.cs
```diff
@@ -0,0 +1,255 @@
...
+            {
+                return false;
+            }
+
+            if (x.GetType() != y.GetType())
+            {
+                return false;
+            }
+
+            if (!this.Options.RequireParsed)
```
twsouthwick, line 48: This method shiuld delegate out to these additional checks if needed to keep it more readable

src/DocumentFormat.OpenXml.Framework/MarkupCompatibilityAttributes.cs
```diff
@@ -42,5 +44,50 @@ public class MarkupCompatibilityAttributes
...
+            {
+                return false;
+            }
+
+            if (object.ReferenceEquals(this, other))
+            {
+                return true;
+            }
+
+            return (this.Ignorable == null == (other.Ignorable == null)) &&
```
twsouthwick, line 67: Could this use pattern matching to simplify it?

test/DocumentFormat.OpenXml.Benchmarks/EqualityBenchmark.cs
```diff
@@ -0,0 +1,116 @@
...
+        private OpenXmlElement _largeElement2;
+
+        private OpenXmlElement _smallElementUnParsed;
+        private OpenXmlElement _smallElement2UnParsed;
+
+        private OpenXmlElement _smallElementParsed;
+        private OpenXmlElement _smallElement2Parsed;
+
+        private IEqualityComparer<OpenXmlElement> _defaultEqualityComparer = OpenXmlElementComparers.Default;
+        private IEqualityComparer<OpenXmlElement> _fast = OpenXmlElementComparers.GetEqualityComparer(new OpenXmlElementEqualityOptions() { SkipPrefixComparison = true });
```
twsouthwick, line 28: Fast doesn't feel right - maybe skipPrefixComparer?

test/DocumentFormat.OpenXml.Tests/OpenXmlElementEqualityTest.cs
```diff
@@ -0,0 +1,312 @@
...
+using System.Collections.Generic;
+
+using Xunit;
+
+namespace DocumentFormat.OpenXml.Tests
+{
+    public class OpenXmlElementEqualityTest
+    {
+        private static readonly IEqualityComparer<OpenXmlElement> DefaultComparer = OpenXmlElementComparers.Default;
+        private static readonly IEqualityComparer<OpenXmlElement> DefaultComparerRequireParsed = OpenXmlElementComparers.GetEqualityComparer(new OpenXmlElementEqualityOptions
```
twsouthwick, line 15: Can these be created in line for readability?

src/DocumentFormat.OpenXml.Framework/MarkupCompatibilityAttributes.cs
PhDuck: I'm not sure how you would like to use pattern matching to make this simpler.

I'm pretty used to this, so it seems easy to me, but if you have a better way, please expand on it.

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementEqualityComparer.cs
PhDuck: I understood this comment as move all the options that change equation behaviour into their own method, hope that was what you meant.

twsouthwick:
@PhDuck I pushed a change that allows using the built in hashcode shim - it's in a different namespace

src/DocumentFormat.OpenXml.Framework/MarkupCompatibilityAttributes.cs
```diff
@@ -42,5 +45,50 @@ public class MarkupCompatibilityAttributes
...
+                (this.Ignorable == null || this.Ignorable.Equals(other.Ignorable)) &&
+                (this.ProcessContent == null == (other.ProcessContent == null)) &&
+                (this.ProcessContent == null || this.ProcessContent.Equals(other.ProcessContent)) &&
+                (this.PreserveElements == null == (other.PreserveElements == null)) &&
+                (this.PreserveElements == null || this.PreserveElements.Equals(other.PreserveElements)) &&
+                (this.PreserveAttributes == null == (other.PreserveAttributes == null)) &&
+                (this.PreserveAttributes == null || this.PreserveAttributes.Equals(other.PreserveAttributes)) &&
+                (this.MustUnderstand == null == (other.MustUnderstand == null)) &&
+                (this.MustUnderstand == null || this.MustUnderstand.Equals(other.MustUnderstand));
```
twsouthwick, line 77:
```suggestion
            return Equals(Ignorable, other.Ignorable)
                && Equals(ProcessContent, other.ProcessContent)
                && Equals(PreserveElements, other.PreserveElements)
                && Equals(MustUnderstand, other.MustUnderstand);
```

Doesn't seem like pattern matching would simplify it, but just calling into Equals(,) would. All the `(a == null == (other.a == null))` seems to complicate the checks.

src/DocumentFormat.OpenXml.Framework/Equality/OpenXmlElementComparers.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Equality comparer for determining value equality for <see cref="OpenXmlElement"/>.
+    /// </summary>
+    public static class OpenXmlElementComparers
+    {
+        /// <summary>
+        /// Gets the default equality comparer.
+        /// </summary>
+        public static readonly IEqualityComparer<OpenXmlElement> Default = GetEqualityComparer(new OpenXmlElementEqualityOptions());
```
twsouthwick, line 16:
```suggestion
        public static readonly IEqualityComparer<OpenXmlElement> Default { get; } =  GetEqualityComparer(new OpenXmlElementEqualityOptions());
```

twsouthwick: We prefer properties over fields

twsouthwick (approved):

twsouthwick:
Thanks @PhDuck! This is cool feature that will be available in v3.0 when it's released

PR #1478, opened 2023-06-29
initial build and test pipeline using actions

tomjebo:


.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,75 @@
+name: dotnet/Open-XML-SDK main pipeline
```
twsouthwick, line 1: can probably just be "Build" or something

.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,75 @@
...
+on:
+  push:
+    branches: 
+      - 'main'
+      - 'release/*'
+  pull_request:
+    branches: 
+      - 'main'
+      - 'release/*'
+      - '!/docs/'
```
twsouthwick, line 16: :) hopefully this works here

.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,75 @@
...
+        fetch-depth: 0
+
+    # Install the .NET Core workload
+    - name: Install .NET Core
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: 6.0.x
+
+    # Add  MSBuild to the PATH: https://github.com/microsoft/setup-msbuild
+    - name: Setup MSBuild.exe
```
twsouthwick, line 50: we shouldn't need this

.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,75 @@
...
+      uses: microsoft/setup-msbuild@v1.0.2
+
+    # Execute all unit tests in the solution
+    # - name: Execute unit tests
+    #  run: dotnet test
+
+    # Restore the projects
+    - name: Restore the Projects
+      run: msbuild $env:Solution_Name /t:Restore /p:Configuration=$env:Configuration
```
twsouthwick, line 59: we can use dotnet restore here

.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,75 @@
...
+      env:
+        Configuration: ${{ matrix.configuration }}
+        ProjectLoadStyle: ${{ matrix.projectloadstyle }}
+        
+            # Restore the projects
+    - name: Build the Projects
+      run: msbuild $env:Solution_Name /p:Configuration=$env:Configuration
```
twsouthwick, line 66: dotnet build --no-restore ...

.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,75 @@
...
+      env:
+        Configuration: ${{ matrix.configuration }}
+        ProjectLoadStyle: ${{ matrix.projectloadstyle }}
+        
+    - name: Run All Tests
+      run: dotnet test --no-restore --verbosity normal --configuration ${{ matrix.configuration }}
```
twsouthwick, line 72: dotnet test --no-restore --no-build

.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,75 @@
...
+    strategy:
+      matrix:
+        configuration: [Debug, Release]
+        projectloadstyle: [ net35, net40, net46, net6_0, net7_0 ]
+
+    runs-on: windows-latest  # For a list of available runner types, refer to
+                             # https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on
+
+    env:
+      Solution_Name: Open-XML-SDK.sln                         # Replace with your solution name, i.e. MyWpfApp.sln.
```
twsouthwick, line 35: can you remove these comments that aren't really necessary once we get it set up?

twsouthwick:
Linux/Mac test runs?

tomjebo:
> Linux/Mac test runs?

yes, I'm adding those now.

.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,75 @@
...
+
+    # Execute all unit tests in the solution
+    # - name: Execute unit tests
+    #  run: dotnet test
+
+    # Restore the projects
+    - name: Restore the Projects
+      run: msbuild $env:Solution_Name /t:Restore /p:Configuration=$env:Configuration
+      env:
+        Configuration: ${{ matrix.configuration }}
```
twsouthwick, line 61: don't need configuration here and in the line. I kin dof like using the nv block for all these projectloadstyle/configuration/etc

tomjebo: @twsouthwick this should be better and has the os platforms in it.

.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,70 @@
...
+      - 'release/*'
+      - '!/docs/'
+      - '!/*.md'
+
+env:
+  RunPeVerify: true 
+  
+jobs: 
+
+  build:
```
twsouthwick, line 24: Can we name this with a capital?

.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,70 @@
...
+
+    - name: Install .NET Core
+      uses: actions/setup-dotnet@v3
+      with:
+          dotnet-version: 6.0.x
+
+    - name: Restore the Projects
+      run: dotnet restore --verbosity normal
+      env:
+          Configuration: ${{ matrix.configuration }}
```
twsouthwick, line 57: Can these be set as env block for the whole job?

twsouthwick (approved):
Lgtm! A couple questions, but we can get it in

.github/workflows/actions-pipeline.yml
```diff
@@ -0,0 +1,70 @@
+name: Build
+
+concurrency:
```
twsouthwick, line 3: do we need this?

.github/workflows/actions-pipeline.yml
tomjebo: according to the docs this looks valid but I haven't tested this.

.github/workflows/actions-pipeline.yml
tomjebo: Yes, how's: 

```yaml
jobs: 

  build_and_test:
    name: Build and Test
```

.github/workflows/actions-pipeline.yml
tomjebo: looks like that's at the workflow level so no, I think we can nix this.

.github/workflows/actions-pipeline.yml
tomjebo: yes, done.

twsouthwick (approved):

Issue #1457, opened 2023-06-27
Bug with Open XML SDK - Removed Records: Formula from /xl/calcChain.xml part (Calculation properties)


subashnalla:
**Describe the bug**
We are getting the "Removed Records: Formula from /xl/calcChain.xml part (Calculation properties)" while opening Excel. I have used DocumentFormat.OpenXml Nugget package in C#. The template Excel (.xlsm) is Macro Enabled one. We are collecting the data from Microstation OPM (Bentley software) dgn and writing the data to .MDB and then picking up the data from MDB and writing to Excel using DocumentFormat.OpenXml Nugget package.

**Screenshots**
![image](https://github.com/dotnet/Open-XML-SDK/assets/70330710/4e0a6c7e-f4a9-4410-9478-b6d2ca45ea6a)
![image](https://github.com/dotnet/Open-XML-SDK/assets/70330710/0abeda25-eaf3-47be-a473-ffe51c09e23f)

**To Reproduce**
I can show the issue. The Excel formulas looks fine, we checked them using the Formulas tab in Excel.

**Observed behavior**
Removed Records: Formula from /xl/calcChain.xml part (Calculation properties)

**Expected behavior**
Excel should open with out any error

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro 21H2 64-bit operating system, x64-based processor
 - Office version: Office 365 (16.0.16501.20228)
 - .NET Target: .NET Framework 4.6.2
 - DocumentFormat.OpenXml Version: (e.g. 2.11.0)

**Additional context**
Client is chasing me to fix the issue and any help is appreciated.

twsouthwick:
Can you provide a repro so we can see what is happening here?

twsouthwick:
@tomjebo is this something related to your work to auto-remove the calcchain?

kodenine:
I've been having the same problem.

Here is a quick repro:

var spreadsheet = SpreadsheetDocument.Open(filename, true);

WorksheetPart worksheetpart = spreadsheet.WorkbookPart.WorksheetParts.First();

var sheetData = worksheetpart.Worksheet.GetFirstChild<SheetData>();

var firstRow = sheetData.Elements<Row>().FirstOrDefault(p => p.RowIndex.HasValue && p.RowIndex.Value == 1);

var firstFormulaCell = firstRow.Elements<Cell>().Last();
firstFormulaCell.CellValue = new CellValue(2);
[...]

PR #1491, opened 2023-07-12
Add feature to track calculation chain changes

twsouthwick:
This is an example of how we may be able to support calculation chain tracking. It does not handle all cases, nor set all values necessary for the calculation chain. It does, however, show how we can track additions or removals of formulas and react accordingly.

### DIFF
--- a/src/DocumentFormat.OpenXml.Features/CalcChain/CalculationChainExtensions.cs
+++ b/src/DocumentFormat.OpenXml.Features/CalcChain/CalculationChainExtensions.cs
@@ -1,0 +1,166 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using DocumentFormat.OpenXml.Features;
+using DocumentFormat.OpenXml.Packaging;
+using DocumentFormat.OpenXml.Spreadsheet;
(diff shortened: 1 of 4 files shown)


src/DocumentFormat.OpenXml.Features/CalcChain/CalculationChainExtensions.cs
```diff
@@ -0,0 +1,166 @@
...
+                return;
+            }
+
+            var chain = GetCalculationChain();
+
+            var existing = Find(chain, r);
+
+            if (!isAdding && existing is not null)
+            {
+                existing.Remove();
```
tomjebo, line 113: @twsouthwick so after more testing, it looks like it's possible that removal of formulas is the only action that will trigger calcChain.xml part validation and subsequent errors in Excel. This piece would probably deal with the issue in #1457 but I still think we should follow the Excel team guidance to recommend removal of the calcChain.xml part. Adding a formula doesn't appear to trigger the validation in Excel.

tomjebo: @twsouthwick could we instead use this trigger to remove the whole calcChain.xml part in response to adding or removing or modifying formulas in cells? I think that would be a better action to take.

PR #1492, opened 2023-07-19
Add mime types to GetImagePartType extension method

### LINKED ISSUE #1488: New ImagePart media added with extension .bin
**Describe the bug**

This is a follow up to an original bug https://github.com/dotnet/Open-XML-SDK/issues/1305 _"New ImagePart is created as .bin instead of real extension"_ which is still an issue with the most recent 2.20.0 release.

When adding a new `ImagePart` the media in the bundled package has an extension of `.bin` instead of `.jpg` etc.

**Screenshots**

**To Reproduce**

Create a new document and add an `ImagePart` using the following code:

```csharp
// Load an image and get the appropriate mime type
var image = System.Drawing.Image.LoadFromFile("myphoto.jpg");
var codes = ImageCodecInfo.GetImageDecoders();
var mimeType = codes.First(c => c.FormatID == image.RawFormat.Guid).MimeType;

var imagePart = mainPart.AddImagePart(mimeType); // image/jpeg

image.Save(stream, image.RawFormat);
stream.Position = 0;
imagePart.FeedData(stream);
```

**Observed behavior**
The media is added to the word document package with .bin extension

**Expected behavior**
The media is added to the word document package with .jpg extension

 - OS: Windows 10
 - Office version: 2306 - 16529.20154
 - .NET Target: .NET Core 6
 - DocumentFormat.OpenXml Version: 2.20.0

mikeebowen:
Fixes #1488

### DIFF
--- a/src/DocumentFormat.OpenXml/Packaging/ImagePartTypeInfo.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/ImagePartTypeInfo.cs
@@ -62,6 +62,16 @@ namespace DocumentFormat.OpenXml.Packaging
                 ".svg" => ImagePartType.Svg,
                 ".tiff" => ImagePartType.Tiff,
                 ".wmf" => ImagePartType.Wmf,
+                "image/bmp" => ImagePartType.Bmp,
+                "image/gif" => ImagePartType.Gif,
+                "image/png" => ImagePartType.Png,
+                "image/tiff" => ImagePartType.Tiff,
+                "image/x-icon" => ImagePartType.Icon,
+                "image/x-pcx" => ImagePartType.Pcx,
+                "image/jpeg" => ImagePartType.Jpeg,
+                "image/x-emf" => ImagePartType.Emf,
+                "image/x-wmf" => ImagePartType.Wmf,
+                "image/svg+xml" => ImagePartType.Svg,
                 _ => throw new NotSupportedException($"{extension} is not supported"),
             };
     }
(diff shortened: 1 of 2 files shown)


src/DocumentFormat.OpenXml/Packaging/ImagePartTypeInfo.cs
```diff
@@ -62,6 +62,16 @@ internal static ImagePartType GetImagePartType(string extension)
                 ".svg" => ImagePartType.Svg,
                 ".tiff" => ImagePartType.Tiff,
                 ".wmf" => ImagePartType.Wmf,
+                "image/bmp" => ImagePartType.Bmp,
```
tomjebo, line 65: clever. probably should rename the input argument from "extension" to something more general (?). Not sure what.

tomjebo: Also, not sure if this is a good idea. Might be better to have a separate GetImagePartTypeByMime() or something.

twsouthwick: This seems fine - the API name isn't saying it's from extensions. How are we handling this in v3.0?

twsouthwick:
Can you clean up the title and add `fixes #1488` to the body of the message?

src/DocumentFormat.OpenXml/Packaging/ImagePartTypeInfo.cs
twsouthwick: > should rename the input argument

Probably ok, but that can be a source breaking change if someone did `GetImagePartType(extension: ".something")`

mikeebowen: > Having one method that handles this seems fine - the API name isn't saying it's from extensions. How are we handling this in v3.0?

v3.0 has the same issue, but the fix will have to be a bit different

twsouthwick (approved):

PR #1493, opened 2023-07-20
Sign and deploy steps

tomjebo:


.github/workflows/build.yml
```diff
@@ -127,4 +129,96 @@ jobs:
       uses: actions/upload-artifact@v3
       with:
         name: Test Results (${{ matrix.os }} ${{ matrix.configuration }} ${{ matrix.projectloadstyle }})
-        path: test-results/*
\ No newline at end of file
+        path: test-results/*
+        
+  sign:
```
twsouthwick, line 134: this should be a separate workflow that only builds on main or release/ and is triggered by a successful build there

twsouthwick (changes requested):

.github/workflows/build.yml
tomjebo: Yes, I'll separate them out.

tomjebo:
@twsouthwick  another question I had was about the dependencies. It seems deploy should depend on signing which should depend on packing success. Packing should depend on build/test success. 

When I separate these out though it looks like we used manual approval to make sure that sign and deploy were ok to proceed (as well as the release event). I'm not sure how to do approval in github actions, perhaps there's a way?

twsouthwick:
I think for approvals you'll set up a deployment

tomjebo:
sign.yml testing was done from main so this PR is no longer useful. We can submit new PR's for changes to signing and deployment workflows as needed. closing this one.

PR #1497, opened 2023-07-26
Fix release condition

tomjebo:


.github/workflows/release.yml
```diff
@@ -33,6 +24,18 @@ jobs:
         - name: Setup .NET Core
           uses: actions/setup-dotnet@v3
 
+        - name: Approve
+          uses: trstringer/manual-approval@v1
+          with:
+              secret: ${{ github.TOKEN }}
+              approvers: mikeebowen,tomjebo,twsouthwick,alfredhellstern
```
twsouthwick, line 31: we should set up a github team for this

twsouthwick (dismissed):

twsouthwick:
My comment is not a blocker - but we should create an issue to track it

.github/workflows/release.yml
tomjebo: we can look into that.

tomjebo:
> My comment is not a blocker - but we should create an issue to track it

that's what I assumed. I looked into it initially but looks like we don't have the rights to create a team in dotnet org on Github. So we might have to ask someone to do that for us unless there's another way to create a team in Github.

PR #1542, opened 2023-10-18
Allow containing part access for elements

tomjebo:
This would allow part access to be done like this: 

```charp
// pShape is set to an OpenXmlElement
OpenXmlPartRootElement partRoot = pShape.Ancestors<OpenXmlPartRootElement>().First();

if (partRoot.OpenXmlPart is null)
{
    Console.WriteLine("partRoot is null");
    return;
}

OpenXmlPart part = partRoot.OpenXmlPart;
```


mikeebowen (commented):
Why are you deleting or commenting out most of the pipeline?

PR #1543, opened 2023-10-18
allow access to part from elements

tomjebo:
With this change, it's possible to do this: 

```csharp
OpenXmlPartRootElement partRoot = pShape.Ancestors<OpenXmlPartRootElement>().First();

if (partRoot.OpenXmlPart is null)
{
    Console.WriteLine("partRoot is null");
    return;
}

OpenXmlPart part = partRoot.OpenXmlPart;
```

### DIFF
--- a/src/DocumentFormat.OpenXml.Framework/OpenXmlPartRootElement.cs
+++ b/src/DocumentFormat.OpenXml.Framework/OpenXmlPartRootElement.cs
@@ -235,10 +235,10 @@ namespace DocumentFormat.OpenXml
         }
 
         /// <summary>
-        /// Gets or sets the part that is associated with the DOM tree.
+        /// Gets the part that is associated with the DOM tree.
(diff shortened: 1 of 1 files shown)


mikeebowen (approved):

tomjebo:
@ashahabov Please take a look at this. I think this would meet your needs based on your [StackOverflow question](https://stackoverflow.com/questions/76984818/open-xml-sdk-how-to-retrieve-parent-openxmlpart-from-a-typedopenxmlcompositeele).

twsouthwick (approved):
LGTM - do we want this in v3?

tomjebo:
> LGTM - do we want this in v3?

yeah, I think that'd be nice.

tomjebo:
@twsouthwick regression test I guess would be to make sure the property doesn't go back to full internal or private. not sure if we really need a test though.

twsouthwick:
No that's not needed. We'll catch that in review

PR #1641, opened 2024-01-08
Add public apis analyzer

twsouthwick:


twsouthwick:
See https://github.com/dotnet/Open-XML-SDK/pull/1640 for a verification that this is the right set of APIs for "shipped"

PR #1690, opened 2024-03-12
add Feb24 Schema updates

mikeebowen:


src/DocumentFormat.OpenXml/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,215 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
twsouthwick, line 1: let's rethink this so we don't cause breaking changes

twsouthwick (changes requested):

PR #1717, opened 2024-05-13
Add a request for documentation when creating an issue

### LINKED ISSUE #1710: Open XML SDK contains too few samples, isn't documented properly on microsoft's site and questions aren't answered on stackoverflow.
When trying to find out how to create niche stuff, it turns out that the tools provided are too old, does not cover niche cases, questions are not answered on stackoverflow, and the samples provided in the Open-XML-SDK repository does not cover the case.

In my opinion we have a problem when you refer to stackoverflow, eg. how to create a chart in a presentation (there is a Sunburst example, but that does not convert trivially to a bar or pie chart), as you can verify by searching stackoverflow for openxml, presentation and chart. It means that when you refer us to stackoverflow, most of the time you might as well refer to /dev/null. If the questions are answered, they answer by telling the users to use a library on top of the SDK, ie. the question is ignored.

Another example is how to reuse the same embeded workbook for several different charts in a presentation. Does that work? Or must the workbook stream be re-read for each chart? Etc.

Also the API reference on microsofts site contains no example code and it's "learn" section is sorely lacking.

What I think is needed, is more samples. Otherwise, what is the point with the Open-XML-SDK, when it's easier to read the specification and not use the SDK?

twsouthwick:
Fixes #1710

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation_request.md
+++ b/.github/ISSUE_TEMPLATE/documentation_request.md
@@ -1,0 +1,20 @@
+---
+name: Documentation request
+about: Create a request for documentation updates
+title: Documentation Request
+labels: 'documentation'
+assignees: ''
+
+---
+
+**Describe the API or scenario**
+A clear and concise description of what the scenario you'd like documented.
+
+**Attempts**
+Please include any attempts you have made to figure this out.
+
+**Related scenarios**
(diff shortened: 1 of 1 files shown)


mikeebowen (approved):
lgtm

tomjebo (approved):
LGTM

PR #1774, opened 2024-08-21
Add open xml editor extension for Visual Studio Project

mikeebowen:
This moves the archived repo for the Open Xml Editor extension for Visual Studio from [Open-XML-Package-Editor-Power-Tool-for-Visual-Studio](https://github.com/OfficeDev/Open-XML-Package-Editor-Power-Tool-for-Visual-Studio) to the Open-XML-SDK repo.


twsouthwick:
A couple of quick thoughts:

- Can we put it in a subfolder `tools` or something similar?
- It can only build for .net framework - probably can't build on the linux machines. you may need to use a solution filter to create slnf files for linux that will have a subset of the projects

PR #1863, opened 2025-01-24
Optimize ToFlatOpcString to reduce memory usage and improve performan…

QuocDatHoang:
## Optimized the ToFlatOpcString method to address high memory usage and improve performance. 

The previous implementation relied on LINQ, which created a significant amount of intermediate collections and dictionaries, leading to inefficiency, especially when working with Word documents containing images. 

This update removes unnecessary LINQ operations and reduces memory overhead, replace manual chunking logic with Convert.ToBase64String using Base64FormattingOptions.InsertLineBreaks. 

This approach eliminates the need for LINQ-based chunking and aggregation, resulting in cleaner, more efficient code while adhering to the MIME specification for line breaks in Base64 encoding.

A sample benchmark output when working with a 822KB-word document.

image

### DIFF
--- a/src/DocumentFormat.OpenXml.Framework/Packaging/FlatOpcExtensions.cs
+++ b/src/DocumentFormat.OpenXml.Framework/Packaging/FlatOpcExtensions.cs
@@ -141,25 +141,7 @@ public static class FlatOpcExtensions
     }
 
     private static string ToChunkedBase64String(byte[] byteArray)
-    {
-        // The MIME specification defines a maximum line length of 76 characters
-        // for a Base64-encoded string. Therefore, we need to break down the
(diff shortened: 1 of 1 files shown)


QuocDatHoang:
@dotnet-policy-service agree

twsouthwick (approved):
This is great! I had wanted to optimize this area figuring it was not performant, but didn't know about that enum option. Thanks!

mikeebowen (approved):

QuocDatHoang:
Thanks @twsouthwick and @mikeebowen!

twsouthwick:
This is included in 3.3.0!

QuocDatHoang:
Thank @twsouthwick :)

PR #1902, opened 2025-03-19
Add async xmlwriter methods

### LINKED ISSUE #913: Writing asynchronously with OpenXmlWriter
The `OpenXmlWriter` doesn't appear to have any async overloads.
Is it not possible to write in a asynchronous fashion?

mikeebowen:
- closes #913


src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriter.cs
```diff
@@ -56,6 +83,37 @@ public OpenXmlPartWriter(OpenXmlPart openXmlPart, Encoding encoding)
...
         }
 
+#if FEATURE_ASYNC_SAX_XML
+        /// <summary>
+        /// Initializes a new instance of the OpenXmlPartWriter.
+        /// </summary>
+        /// <param name="openXmlPart">The OpenXmlPart to be written to.</param>
+        /// <param name="encoding">The encoding for the XML stream.</param>
+        /// <param name="useAsync">Whether the writer should be initialized using async</param>
+        public OpenXmlPartWriter(OpenXmlPart openXmlPart, Encoding encoding, bool useAsync)
```
twsouthwick, line 93: I'm tempted to create a new type like `OpenXmlPartWriterSettings` that would expose `Encoding` and `UseAsync` and potentially other properties. i.e. there have been asks for "sanitizing methods" and I think a general settings object here would make that easier to plumb through

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriter.cs
```diff
@@ -110,7 +221,6 @@ public override void WriteStartDocument()
         public override void WriteStartDocument(bool standalone)
         {
             ThrowIfObjectDisposed();
```
twsouthwick, line 223: nit: I kind of like the whitespace. Is there a reason you removed it?

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriter.cs
```diff
@@ -365,5 +471,179 @@ public override void Close()
...
+#if FEATURE_ASYNC_SAX_XML
+        /// <summary>
+        /// Asynchronously writes the XML declaration with the version "1.0".
+        /// </summary>
+        public async override Task WriteStartDocumentAsync()
+        {
+            ThrowIfObjectDisposed();
+            await _xmlWriter.WriteStartDocumentAsync().ConfigureAwait(true);
+            return;
+        }
```
twsouthwick, line 485:
```suggestion
        public override Task WriteStartDocumentAsync()
        {
            ThrowIfObjectDisposed();
            return _xmlWriter.WriteStartDocumentAsync();
        }
```

For these kind of methods, we don't need to actually await, but rather just pass the task back.

The only time that doesn't work is if you need to do something with it after (including if you have any using statements)

src/DocumentFormat.OpenXml.Framework/OpenXmlWriter.cs
```diff
@@ -123,13 +126,86 @@ protected OpenXmlWriter()
...
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Asynchronously writes out a start tag of the element and all the attributes of the element.
+        /// </summary>
+        /// <param name="elementObject">The OpenXmlElement object to be written.</param>
+        public virtual Task WriteStartElementAsync(OpenXmlElement elementObject)
+        {
+            throw new NotImplementedException();
```
twsouthwick, line 158: instead of throwing, it may be helpful to implement against sync method. They're new methods, but any implementations of OpenXmlWriter that doesn't override them will throw, so I think we should default to an implementation that won't throw

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriter.cs
```diff
@@ -365,5 +471,179 @@ public override void Close()
...
+
+        // Async Methods
+#if FEATURE_ASYNC_SAX_XML
+        /// <summary>
+        /// Asynchronously writes the XML declaration with the version "1.0".
+        /// </summary>
+        public async override Task WriteStartDocumentAsync()
+        {
+            ThrowIfObjectDisposed();
+            await _xmlWriter.WriteStartDocumentAsync().ConfigureAwait(true);
```
twsouthwick, line 483: do these async methods take a CancellationToken? If so, we should be taking that in as well.

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriter.cs
```diff
@@ -365,5 +471,179 @@ public override void Close()
...
+            }
+
+            if (elementObject is OpenXmlMiscNode)
+            {
+                throw new ArgumentOutOfRangeException(nameof(elementObject));
+            }
+
+            ThrowIfObjectDisposed();
+
+            _xmlWriter.WriteStartElement(elementObject.Prefix, elementObject.LocalName, elementObject.NamespaceUri);
```
twsouthwick, line 516: Looks like there is an async of this: https://apisof.net/catalog/72a43ccf1465a312c9b8d1244e44dbcf

twsouthwick (changes requested):

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriter.cs
mikeebowen: Fixed

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriter.cs
mikeebowen: I double-checked, they don't take a CancellationToken

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriter.cs
mikeebowen: This was VSCode being too aggressive about formatting. I added the blank lines back

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriter.cs
mikeebowen: Added an OpenXmlPartWriterSettings class

twsouthwick (changes requested):
Looking good - with this new settings object, it can be made available everywhere, except that the `Async` property would be around the if/def

src/DocumentFormat.OpenXml.Framework/OpenXmlWriter.cs
```diff
@@ -123,13 +126,118 @@ protected OpenXmlWriter()
...
+
+        /// <summary>
+        /// Asynchronously writes out a start tag of the element. And write the attributes in attributes. The attributes of the element will be omitted.
+        /// </summary>
+        /// <param name="elementObject">The OpenXmlElement object to be written.</param>
+        /// <param name="attributes">The attributes to be written.</param>
+        /// <param name="namespaceDeclarations">The namespace declarations to be written, can be null if no namespace declarations.</param>
+        public virtual Task WriteStartElementAsync(OpenXmlElement elementObject, IEnumerable<OpenXmlAttribute> attributes, IEnumerable<KeyValuePair<string, string>> namespaceDeclarations)
+        {
+            ThrowIfObjectDisposed();
```
twsouthwick, line 195: Are these necessary? Don't the call to the original Write... methods do this?

twsouthwick (dismissed):
Just a nit, but overall good!

tomjebo (approved):

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriterSettings.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace DocumentFormat.OpenXml;
+
+/// <summary>
+/// Settings for the OpenXmlPartWriter.
+/// </summary>
+public class OpenXmlPartWriterSettings
+{
+#if FEATURE_ASYNC_SAX_XML
+    /// <summary>
+    /// Gets or sets a value indicating whether asynchronous OpenXmlPartWriter methods can be used.
```
twsouthwick, line 15:
```suggestion
    /// Gets or sets a value indicating whether asynchronous <see cref="OpenXmlPartWriter" /> methods can be used.
```

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriterSettings.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    public bool Async { get; set; }
+#endif
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the OpenXmlPartWriter should check to ensure that all characters in the document conform to the "2.2 Characters" section of the W3C XML 1.0 Recommendation.
```
twsouthwick, line 21:
```suggestion
    /// Gets or sets a value indicating whether the <see cref="OpenXmlPartWriter" /> should check to ensure that all characters in the document conform to the "2.2 Characters" section of the W3C XML 1.0 Recommendation.
```

Do you have a link to that part of the docs?

twsouthwick: Is that really what the "CloseOutput" property does?

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriterSettings.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    public bool CloseOutput { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the OpenXmlPartWriter should also close the underlying stream or TextWriter when the Close() method is called.
```
twsouthwick, line 26: I think these summaries are off...

twsouthwick (approved):
Take a look at the docs, then it's good to merge

Issue #1548, opened 2023-11-01
New APIs to facillitate getting started


twsouthwick:
This is a placeholder issue as we revamp some of the docs to identify APIs we should add to better handle common cases that are exposed by nullability analysis.  There appears to be a lot of boilerplate to build up a document that requires null checking and then adding parts, etc that we could probably add some helper methods to simplify things.

For example:

```csharp
Sheets? sheets = workbookPart.Workbook.GetFirstChild<Sheets>();
```

this could be null, so defensively, we would want something like:

```csharp
Sheets sheets = workbookPart.Workbook.GetFirstChild<Sheets>() ?? workbookPart.Workbook.AppendChild(new Sheets());
```

This becomes cumbersome, when we could add something similar to:

```csharp
Sheets sheets = workbookPart.Workbook.GetFirstChildOrAdd<Sheets>();
```

Addtional APIs may be helpful here, and so I'm creating this issue to track some of these quality of life related APIs. I think we should also consider identifying patterns for analyzers to catch to move people to these better patterns as well

mikeebowen:
@twsouthwick Would something like this work? https://github.com/dotnet/Open-XML-SDK/compare/main...mikeebowen:Open-XML-SDK:ergonomic-apis

mikeebowen:
closed by #1904

PR #1904, opened 2025-03-20
Add GetOrAddFirstChild method

mikeebowen:
- #1548


twsouthwick (approved):

src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
```diff
@@ -796,11 +796,30 @@ public void RemoveNamespaceDeclaration(string prefix)
...
         /// </summary>
         /// <typeparam name="T">Type of element.</typeparam>
-        /// <returns></returns>
+        /// <returns>The first child element of type T or null</returns>
         public T? GetFirstChild<T>()
             where T : OpenXmlElement
             => ChildElements.First<T>();
 
+        /// <summary>
+        /// Finds the first child element of type T or adds a new element if it does not exist.
```
twsouthwick, line 805:
```suggestion
        /// Finds the first child element of <typeparam ref="T"/> or adds a new element if it does not exist.
```

or something like that so it references it in intellisense better

twsouthwick (approved):

src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
mikeebowen: updated

mkaszewiak (approved):

Issue #1769, opened 2024-08-01
`pictureOptions` and `bubble3D` were removed from some complex types in O15 but not the SDK


tomjebo:
Need to `[Obsolete]` the following: 

`pictureOptions` from `CT_LineSer`, `CT_RadarSer`, `CT_PieSer`, `CT_BubbleSer` and `CT_SurfaceSer`
`bubble3D` from `CT_SurfaceSer`

these children have been removed in the O15 Office schemas but were never removed from the SDK.

Once obsoleted, they will be marked for deletion in the next major release.

tomjebo:
#1905 and #1906 complete this issue.

PR #1905, opened 2025-03-21
obsolete several child elements removed in Office 2015

tomjebo:
this pr obsoletes the properties mentioned in #1769 
it also adds code to the generator to add code attributes like ObsoleteAttribute to complex type classes, child elements and attributes.

### DIFF
--- a/gen/DocumentFormat.OpenXml.Generator.Models/Generators/Elements/DataModelWriterExtensions.cs
+++ b/gen/DocumentFormat.OpenXml.Generator.Models/Generators/Elements/DataModelWriterExtensions.cs
@@ -54,8 +54,8 @@ public static class DataModelWriterExtensions
             //        },
             //        {
             //              This is an example obsoleting a child attribute (property in C#)
-            //              In the json metadata: use for example "QName": ":formatCode",
-            //              ":formatCode",
+            //              In the json metadata: use for example "QName" converted to a TypedQName string: "/:formatCode",
+            //              "/:formatCode",
             //              ObsoleteAttributeWarnList
             //        },
             //  }
@@ -172,11 +172,11 @@ public static class DataModelWriterExtensions
     {
         writer.WriteDocumentationComment(BuildTypeComments(services, element));
 
-        if (_attributeData.TryGetValue(element.Name.ToString(), out Dictionary<TypedQName, List<string>> ctAttributeData))
+        if (_attributeData.TryGetValue(element.Name, out Dictionary<TypedQName, List<string>> ctAttributeData))
         {
             // if the fully qualified CT/tag name is also one of the children of the dictionary that means the attributes of that
             // child's list need to be applied to the whole class, for example, if we're obsoleting an entire class.
-            if (ctAttributeData.TryGetValue(element.Name.ToString(), out List<string> attributeStrings))
+            if (ctAttributeData.TryGetValue(element.Name, out List<string> attributeStrings))
             {
                 foreach (string attributeString in attributeStrings)
                 {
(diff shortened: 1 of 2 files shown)


mikeebowen (dismissed):

gen/DocumentFormat.OpenXml.Generator.Models/Generators/Elements/DataModelWriterExtensions.cs
```diff
@@ -11,6 +11,132 @@ namespace DocumentFormat.OpenXml.Generator.Generators.Elements;
...
+            //        },
+            //  }
+            // },
+            {
+              "c:CT_BubbleSer/c15:ser",
+              new Dictionary<string, List<string>>()
+              {
+                    {
+                          "c:CT_PictureOptions/c:pictureOptions",
+                          new List<string>()
```
twsouthwick, line 64: These are duplicated, so it may be nice to have a readonly static list that can be reused

twsouthwick (dismissed):

gen/DocumentFormat.OpenXml.Generator.Models/Generators/Elements/DataModelWriterExtensions.cs
```diff
@@ -100,7 +240,16 @@ private static void WriteType(this IndentedTextWriter writer, OpenXmlGeneratorSe
             foreach (var attribute in element.Attributes)
             {
                 delimiter.AddDelimiter();
-                writer.WriteAttributeProperty(services, attribute);
+
+                if (_attributeData.TryGetValue(element.Name.ToString(), out Dictionary<string, List<string>> attrAttributeData)
```
twsouthwick, line 244: you could make the dictionary have the type of element.Name (I think it's a `TypedQName`) and it'll make it a bit more performant. The keys you're creating will also automatically be converted to it as well

tomjebo: thanks, that was a good idea. for `SchemaAttributes`, they aren't actually `TypedQName` but I decided to try using the `Type` property of the attribute metadata to construct a string that can convert to a `TypedQName` for the purposes of keys in the dictionary. It was either that or `string.Empty` which would have issues in parsing. So even though the `Type` property of the attribute metadata is actually the c# type (class), it works. If you have an alternative suggestion I'll try that.

mikeebowen (approved):

gen/DocumentFormat.OpenXml.Generator.Models/Generators/Elements/DataModelWriterExtensions.cs
```diff
@@ -11,6 +11,112 @@ namespace DocumentFormat.OpenXml.Generator.Generators.Elements;
 
 public static class DataModelWriterExtensions
 {
+    public static class AttributeStrings
+    {
+        public const string ObsoletePropertyWarn = "[Obsolete(\"Unused property, will be removed in a future version.\", false)]";
```
twsouthwick, line 16: Do we need all these? looks like only two are used

twsouthwick (approved):

gen/DocumentFormat.OpenXml.Generator.Models/Generators/Elements/DataModelWriterExtensions.cs
tomjebo: Not right now, but I added them to make clear the distinction in the different types of obsolescence so we follow some pattern in naming.

twsouthwick: Not a blocker, but I prefer to compose them from smallerbits of text to ensure they're built up the same.

PR #1915, opened 2025-04-04
Add OpenXmlElement.IsValidChild method

### LINKED ISSUE #1212: Add API to determine if a child is valid for a given element
When trying to modify a document, one needs to know if a particular element can add a particular child before doing so.  Before, this was one way of doing so which is no longer works:

```cs
public static bool CanAddChild<T>(this OpenXmlElement element)
{
	if (element == null)
		return false;
	var attributes = element.GetType().GetCustomAttributes<ChildElementInfoAttribute>();
	return attributes.Any(r => r.ElementType == typeof(T));
}
```

There does not appear to be a way to do this anymore as all of the metadata information is internal and created in the internal `ConfigureMetadata` override.

**Describe the solution you'd like**
An API such as the above to determine and/or list the valid children of a given element.

**Describe alternatives you've considered**
One way might be to add the child then validate it, but this seems the wrong approach.

mikeebowen:
- closes #1212

### DIFF
--- a/src/DocumentFormat.OpenXml.Framework/OpenXmlCompositeElement.cs
+++ b/src/DocumentFormat.OpenXml.Framework/OpenXmlCompositeElement.cs
@@ -3,7 +3,6 @@
 
 using DocumentFormat.OpenXml.Framework;
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
--- a/src/DocumentFormat.OpenXml.Framework/OpenXmlCompositeElement.cs
+++ b/src/DocumentFormat.OpenXml.Framework/OpenXmlCompositeElement.cs
@@ -3,6 +3,7 @@
 
 using DocumentFormat.OpenXml.Framework;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
--- a/src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
+++ b/src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
@@ -1330,6 +1330,21 @@ namespace DocumentFormat.OpenXml
             return GetOrder(this, element) == ElementOrder.Before;
         }
 
+        /// <summary>
+        /// Determines if the specified element is a valid child of the current element.
+        /// </summary>
+        /// <param name="element">The element to check.</param>
+        /// <returns>True if the specified element is a valid child; otherwise, false.</returns>
+        public bool IsValidChild(OpenXmlElement element)
+        {
+            if (element is null)
+            {
+                throw new ArgumentNullException(nameof(element));
+            }
+
+            return Metadata.Children.Elements.Any(el => el.Type.Name.Equals(element.Metadata.Type.Name));
+        }
+
         private enum ElementOrder
         {
             Same, // same element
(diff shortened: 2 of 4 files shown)


test/DocumentFormat.OpenXml.Tests/ofapiTest/OpenXmlElementTest2.cs
```diff
@@ -240,5 +240,43 @@ public void GetOrAddFirstChildTest()
...
+
+            // Act
+            var result = parentElement.IsValidChild(invalidChild);
+
+            // Assert
+            Assert.False(result);
+        }
+
+        [Fact]
+        public void IsValidChild_NullChild_ThrowsArgumentNullException()
```
twsouthwick, line 273: should this throw or should it just return false?

src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
```diff
@@ -1330,6 +1330,21 @@ public bool IsBefore(OpenXmlElement element)
...
+        /// <param name="element">The element to check.</param>
+        /// <returns>True if the specified element is a valid child; otherwise, false.</returns>
+        public bool IsValidChild(OpenXmlElement element)
+        {
+            if (element is null)
+            {
+                throw new ArgumentNullException(nameof(element));
+            }
+
+            return Metadata.Children.Elements.Any(el => el.Type.Name.Equals(element.Metadata.Type.Name));
```
twsouthwick, line 1345: Wouldn't this only apply on OpenXmlCompositeElement? Should this be virtual and default to false and overriden on that type with this implementation?

mikeebowen: That's true, updated in latest commit

test/DocumentFormat.OpenXml.Tests/ofapiTest/OpenXmlElementTest2.cs
mikeebowen: Updated to return false in latest commit

src/DocumentFormat.OpenXml.Framework/OpenXmlCompositeElement.cs
```diff
@@ -9,6 +9,7 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
+using System.Linq;
```
twsouthwick, line 12:
```suggestion
```

if we remove the linq below

twsouthwick (approved):
LGTM just a thought about not using LINQ

src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
```diff
@@ -1330,6 +1330,16 @@ public bool IsBefore(OpenXmlElement element)
             return GetOrder(this, element) == ElementOrder.Before;
         }
 
+        /// <summary>
+        /// Determines if the specified element is a valid child of the current element.
+        /// </summary>
+        /// <param name="element">The element to check.</param>
+        /// <returns>True if the specified element is a valid child; otherwise, false.</returns>
+        public virtual bool IsValidChild(OpenXmlElement element)
```
twsouthwick, line 1338: do we need this method on here? Would having it on OpenXmlCompositeElement be enough?

twsouthwick (approved):
LGTM modulo the question

src/DocumentFormat.OpenXml.Framework/OpenXmlElement.cs
mikeebowen: I added the virtual method so it would be overridable

mikeebowen: @twsouthwick see the latest commit

PR #1920, opened 2025-04-23
Add IsMinimumDocument method to WordprocessingDocument, SpreadsheetDocument, and PresentationDocument

### LINKED ISSUE #816: Is it possible to check if a file is a valid document and not some other file type renamed to .docx or some other office extension?

Is this a:
- [ ] Issue with the OpenXml library
- [ X ] Question on library usage

---------------- Remove this line and above before posting ----------------

**Description**

Is it possible to check if a file is a valid document and not some other file type renamed to .docx or some other office extension? I know this may be off-topic here but it is becoming an increasing problem for me as people are renaming other files e.g. videos to .docx and uploading them to my servers.

**Information**

- .NET Target: .NET Core 3.1
- DocumentFormat.OpenXml Version: Any version

mikeebowen:
- Adds methods to determine if a document is a minimum document for WordprocessingDocument, SpreadsheetDocument, and PresentationDocument.
- closes #816

### DIFF
--- a/src/DocumentFormat.OpenXml/Packaging/PresentationDocument.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/PresentationDocument.cs
@@ -2,11 +2,14 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using DocumentFormat.OpenXml.Builder;
+using DocumentFormat.OpenXml.Drawing;
 using DocumentFormat.OpenXml.Features;
+using DocumentFormat.OpenXml.Presentation;
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.IO.Packaging;
+using System.Linq;
 using System.Reflection;
 
 namespace DocumentFormat.OpenXml.Packaging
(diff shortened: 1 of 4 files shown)


twsouthwick (changes requested):
This is a great idea. However, it would be great for it to have the following characteristics:

- Be overrideable/replaceable
- Be accessed off of OpenXmlPackage directly rather than implemented separately

Questions that would decide the API shape:

- Should this be a separate API? Or should it be opt in while opening a package? I tend not to be a fan of APIs that check if something is valid then doing it - there's always the potential something gets changed underneath
- Do we really want to validate the extension? What about if it's a stream/package? If it is overrideable, we could have a check if we've opened it with a path to validate that if needed
- Are we OK with it throwing an exception if it is an ill-formed package (i.e. we already do that)?

mikeebowen:
@twsouthwick 

> Be overrideable/replaceable

Created virtual method in `OpenXmlPackage`

> Be accessed off of OpenXmlPackage directly rather than implemented separately

There is a default implementation in `OpenXmlPackage` but the individual Wordprocessing, Spreadsheet, and Presentation Documents have unique implementations for their different requirements.

> Should this be a separate API? Or should it be opt in while opening a package? I tend not to be a fan of APIs that check if something is valid then doing it - there's always the potential something gets changed underneath

Added `VerifyMinimumPackage` option to `OpenSettings` when true, the package contents are examined and throw an exception if minimum package requirements are not met

> Do we really want to validate the extension? What about if it's a stream/package? If it is overrideable, we could have a check if we've opened it with a path to validate that if needed

The minimum package is verified on all Open overloads with `VerifyMinimumPackage` true. The Open method handles exceptions for the file path such as invalid extension or invalid characters

> Are we OK with it throwing an exception if it is an ill-formed package (i.e. we already do that)?

Removed file checks. The Open method already handles those.

src/DocumentFormat.OpenXml/Packaging/WordprocessingDocument.cs
```diff
@@ -310,6 +310,96 @@ public static WordprocessingDocument Open(Package package, OpenSettings openSett
...
+
+                using (WordprocessingDocument wordprocessingDocument = Open(path, false))
+                {
+                    return wordprocessingDocument?.MainDocumentPart?.Document?.Body is not null;
+                }
+            }
+            catch
+            {
+                return false;
+            }
```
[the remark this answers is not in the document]
mikeebowen: No longer applicable

src/DocumentFormat.OpenXml/Packaging/SpreadsheetDocument.cs
```diff
@@ -267,6 +268,111 @@ public static SpreadsheetDocument Open(System.IO.Stream stream, bool isEditable)
...
+                    Sheet? sheet = spreadsheetDocument?.WorkbookPart?.Workbook?.Sheets?.GetFirstChild<Sheet>();
+                    SheetData? sheetData = spreadsheetDocument?.WorkbookPart?.WorksheetParts?.FirstOrDefaultAndMaxOne()?.Worksheet?.GetFirstChild<SheetData>();
+
+                    return sheet is not null && sheetData is not null;
+                }
+            }
+            catch
+            {
+                return false;
+            }
```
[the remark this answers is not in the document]
mikeebowen: No longer applicable

src/DocumentFormat.OpenXml/Packaging/PresentationDocument.cs
```diff
@@ -267,6 +270,120 @@ public static PresentationDocument Open(Package package, OpenSettings openSettin
...
+
+                    return notesSize is not null && notesSize.Cx is not null && notesSize.Cx.HasValue &&
+                        notesSize.Cx >= 0 && notesSize.Cx <= 27273042316900 && notesSize.Cy is not null && notesSize.Cy.HasValue &&
+                        notesSize.Cy >= 0 && notesSize.Cy <= 27273042316900;
+                }
+            }
+            catch
+            {
+                return false;
+            }
```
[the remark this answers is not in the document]
mikeebowen: No longer applicable

twsouthwick:
@mikeebowen this would be a good use case for a feature so it can be overridden at runtime vs compile time. Here's an example of how to do it: https://github.com/dotnet/Open-XML-SDK/commit/fa893dd6686124d2490ba92f7178b64308448698

twsouthwick:
A few questions about the direction/goal here

**Only a few kinds of packages are supported (such as in the presentation doc) - it will fail if it's an unsupported one**

Is this what we want? I'd expect it to validate it but fail if it is invalid. It appears to also fail if we don't have a check which seems a little unexpected.

**Validate() throws and returns true/false**

I'd expect it to do one or the other. Is the manual throwing due to custom messages? If so, let's make it consistent

**Mixture of exception types**

I think we should use an existing OpenXmlPackageException type or create a new one for this

**How to think about this vs validation**

At some point, how is this different from the existing OpenXmlValidator logic? Should we be hooking this up to that and performing a full validation?

src/DocumentFormat.OpenXml.Framework/Validation/DocumentValidator.cs
```diff
@@ -106,6 +107,20 @@ private void ValidatePart(OpenXmlPart part, ValidationContext context)
                     {
                         Validate(context);
                     }
+                    else if (part.Uri.ToString().EndsWith(".xml", System.StringComparison.InvariantCultureIgnoreCase) &&
```
twsouthwick, line 110: need to dispose the stream from getpart. probably best to add an extension method of ".IsEmptyPart()" that can handle correctly disposing the stream

twsouthwick (changes requested):

mkaszewiak (dismissed):

src/DocumentFormat.OpenXml.Framework/Validation/DocumentValidator.cs
```diff
@@ -106,6 +107,19 @@ private void ValidatePart(OpenXmlPart part, ValidationContext context)
                     {
                         Validate(context);
                     }
+                    else if (part.Uri.ToString().EndsWith(".xml", System.StringComparison.InvariantCultureIgnoreCase) && part.IsEmptyPart())
```
twsouthwick, line 110: this check should be moved into the IsEmptyPart method

src/DocumentFormat.OpenXml.Framework/Packaging/OpenXmlPart.cs
```diff
@@ -448,6 +448,18 @@ internal virtual bool IsInVersion(FileFormatVersions version)
...
+        internal override bool IsEmptyPart()
+        {
+            bool isEmptyPart;
+
+            using (Stream stream = GetStream())
+            {
+                isEmptyPart = stream.Length == 0;
+            }
+
+            return isEmptyPart;
```
twsouthwick, line 460:
```suggestion
            using (Stream stream = GetStream())
            {
                return stream.Length == 0;
            }
```

src/DocumentFormat.OpenXml.Framework/Validation/DocumentValidator.cs
mikeebowen: Done

mkaszewiak (approved):

PR #1936, opened 2025-06-11
Fixes #1933 - CreateElement calls now parametrized UnknownOpenXmlElement constructor

### LINKED ISSUE #1933: When the root element is an unknown element, OpenXmlPartReader could not get the correct values of LocalName, NamespaceUri and Prefix properties
**Describe the bug**
In the CreateElement method of OpenXmlPartReader, when the element is unknown, it returns a new instance of OpenXmlUnknownElement but does not pass the qualified name and prefix parameters to its constructor, resulting in the reader being unable to retrieve the correct values of some properties such as LocalName

**Observed behavior**
When the root element is an unknown element, the LocalName, NamepaceUri and Prefix properties of the reader are string.Empty

**Expected behavior**
The reader returns the correct values of the element

**Additional context**
It seems that even for the construction of known elements, the relevant parameters are not passed. If I use a different namespace prefix, the reader's Prefix property still returns the element's default prefix, not the actual value. Is this also a bug?

mkaszewiak:
Fixes #1933

### DIFF
--- a/src/DocumentFormat.OpenXml.Framework/OpenXmlPartReader.cs
+++ b/src/DocumentFormat.OpenXml.Framework/OpenXmlPartReader.cs
@@ -734,6 +734,7 @@ namespace DocumentFormat.OpenXml
 
             // create the root element object
             var rootElement = CreateElement(new(_xmlReader.NamespaceURI, _xmlReader.LocalName));
+
             if (rootElement is null)
             {
                 throw new InvalidDataException(ExceptionMessages.PartUnknown);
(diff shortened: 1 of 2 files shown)


twsouthwick (approved):

twsouthwick:
can you add a test that validates this behavior?

mikeebowen (changes requested):
This looks good to me, except as @twsouthwick already mentioned there should be a test for this behavior

mikeebowen (approved):
LGTM

PR #1963, opened 2025-09-26
Schema update Q3 2025

mkaszewiak:


twsouthwick:
Have we always used FY? I thought we'd been using calendar year since that makes more sense for those external to microsoft.

PR #2039, opened 2026-03-19
Fix ChangeDocumentType not removing VBA parts when converting to non-macro type

### LINKED ISSUE #618: ChangeDocumentType does not fully remove vbaProject reference
**Description**

I'm using the sample code at https://docs.microsoft.com/en-us/office/open-xml/how-to-convert-a-word-processing-document-from-the-docm-to-the-docx-file-format to remove macros from a docm file and convert it to docx because we have some filtering software in place that prevents transferring files with macros.

Using that sample code I delete the VbaProjectPart, change the document type, and change the file extension.  However, our filtering software identifies the resulting file as corrupt (Word 2016 opens the file just fine though, so it is probably within spec).

When I examine the contents of the .docx file, I notice that the [Content_Types].xml file at the root   still contains the following line:
<Default ContentType="application/vnd.ms-office.vbaProject" Extension="bin"/>

The VbaProjectPart PartName reference is gone and no other content in the .docx file structure contains any macro components.  I think that one line in [Content_Types].xml is enough to trip up our scanner.

Is there some other way to get rid of this line that I'm missing, is this a bug, or is this structure just something that our scanning software should accept?

**Information**

- .NET Target: .NET Core 2.1.12
- DocumentFormat.OpenXml Version: 2.10.0-beta0002

**Repro**

            bool fileChanged = false;

            using (WordprocessingDocument document = WordprocessingDocument.Open(sourcePath, true))
            {
                // Access the main document part.
                var docPart = document.MainDocumentPart;

                // Look for the vbaProject part. If it is there, delete it.
                var vbaPart = docPart.VbaProjectPart;
                if (vbaPart != null)
                {
                    // Delete the vbaProject part and then save the document.
                    docPart.DeletePart(vbaPart);
                    docPart.Document.Save();

                    // Track that the document has been changed.
                    fileChanged = true;
                }

                // Change the document type to not macro-enabled
                document.ChangeDocumentType(WordprocessingDocumentType.Document);
            }

            if (fileChanged)
            {
                // If it already exists, it will be deleted!
                if (File.Exists(destinationPath))
                    {
                        File.Delete(destinationPath);
                    }

                // Rename the file and save changes
                Directory.CreateDirectory(destinationDirectory);
                File.Move(sourcePath, destinationPath);
            }

**Observed**

file.docx [Content_Types].xml still contains a macro reference.

**Expected**
file.docx should not contain any references to macros.

iamcymentho:
## Summary

Fixes #618

When changing a macro-enabled document (`.docm`) to a standard document (`.docx`) via `ChangeDocumentType()`, VBA-related parts (`VbaProjectPart`, `VbaDataPart`, etc.) were carried over to the new document type unchanged. This left orphaned `vbaProject` content type entries in `[Content_Types].xml`, producing a file that:

- Could not be opened correctly by some tools
- Silently retained macro capabilities in a supposedly non-macro document

### Changes

- **`TypedPackageFeatureCollection.cs`** — After `ChangeDocumentTypeInternal` completes, check if the new content type is non-macro-enabled. If so, remove parts with VBA relationship types (`vbaProject`, `keyMapCustomizations`) from the new main part.
- **`DocxTests01.cs`** — Added test that creates a `.docm` with a `VbaProjectPart`, converts it to `.docx`, and verifies:
  1. `VbaProjectPart` is `null` after conversion
  2. The `vbaProject` content type is absent from the package

### Test results

| Test Project | Total | Passed | Failed |
|---|---|---|---|
| DocumentFormat.OpenXml.Tests | 2316 | 2313 | 0 |
| DocumentFormat.OpenXml.Framework.Tests | 192 | 192 | 0 |
| DocumentFormat.OpenXml.Packaging.Tests | 190 | 190 | 0 |

## Test plan

- [x] New test `ChangeDocumentType_MacroEnabledToDocument_RemovesVbaParts` passes
- [x] All 3 existing `ChangeDocumentType` tests still pass
- [x] Full test suite (2698 tests) passes with 0 failures

### DIFF
--- a/src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
@@ -96,6 +96,7 @@ internal abstract partial class TypedPackageFeatureCollection<TDocumentType, TMa
         try
         {
             Package.ChangeDocumentTypeInternal(CreateMainPart());
+            OnDocumentTypeChanged(newType);
         }
         catch (OpenXmlPackageException e)
         {
@@ -115,6 +116,14 @@ internal abstract partial class TypedPackageFeatureCollection<TDocumentType, TMa
     public abstract TDocumentType? GetDocumentType(string contentPart);
 
     protected abstract TMainPart CreateMainPart();
+
+    /// <summary>
+    /// Called after the document type has been changed. Subclasses can override to perform
+    /// cleanup such as removing parts that are not valid for the new document type.
+    /// </summary>
+    protected virtual void OnDocumentTypeChanged(TDocumentType newType)
+    {
+    }
 
     bool IKnownDataPartFeature.IsKnown(string relationshipId) => false;
 }
--- a/src/DocumentFormat.OpenXml/Packaging/WordprocessingDocument.cs
+++ b/src/DocumentFormat.OpenXml/Packaging/WordprocessingDocument.cs
@@ -582,6 +582,21 @@ namespace DocumentFormat.OpenXml.Packaging
                 "application/vnd.ms-word.template.macroEnabledTemplate.main+xml" => WordprocessingDocumentType.MacroEnabledTemplate,
                 _ => default,
             };
+
+            protected override void OnDocumentTypeChanged(WordprocessingDocumentType newType)
+            {
+                if (newType is WordprocessingDocumentType.MacroEnabledDocument
+                    or WordprocessingDocumentType.MacroEnabledTemplate)
+                {
+                    return;
+                }
+
+                if (MainPart is { } mainPart)
+                {
+                    mainPart.DeletePartsRecursivelyOfTypeBase<VbaProjectPart>();
+                    mainPart.DeletePartsRecursivelyOfTypeBase<CustomizationPart>();
+                }
+            }
         }
     }
 }
--- a/test/DocumentFormat.OpenXml.Tests/DocxTests01.cs
+++ b/test/DocumentFormat.OpenXml.Tests/DocxTests01.cs
@@ -362,6 +362,50 @@ namespace DocumentFormat.OpenXml.Tests
         }
 
         [Fact]
+        public void ChangeDocumentType_MacroEnabledToDocument_RemovesVbaParts()
+        {
+            using var stream = new MemoryStream();
+
+            // Create a macro-enabled document with a VbaProjectPart
+            using (var doc = WordprocessingDocument.Create(stream, WordprocessingDocumentType.MacroEnabledDocument))
+            {
+                var mainPart = doc.AddMainDocumentPart();
+                mainPart.Document = new W.Document(new W.Body(new W.Paragraph(new W.Run(new W.Text("Test")))));
+
+                // Add a VbaProjectPart (the part that causes the issue)
+                mainPart.AddNewPart<VbaProjectPart>();
+
+                Assert.NotNull(mainPart.VbaProjectPart);
+                Assert.Equal(WordprocessingDocumentType.MacroEnabledDocument, doc.DocumentType);
+            }
+
+            // Re-open and change type to a standard document
+            stream.Position = 0;
+
+            using (var doc = WordprocessingDocument.Open(stream, true))
+            {
+                Assert.Equal(WordprocessingDocumentType.MacroEnabledDocument, doc.DocumentType);
+                Assert.NotNull(doc.MainDocumentPart.VbaProjectPart);
+
+                doc.ChangeDocumentType(WordprocessingDocumentType.Document);
+
+                Assert.Equal(WordprocessingDocumentType.Document, doc.DocumentType);
+
+                // After changing to a non-macro type, VbaProjectPart should be removed
+                Assert.Null(doc.MainDocumentPart.VbaProjectPart);
+            }
+
+            // Re-open and verify the VBA part is no longer present
+            stream.Position = 0;
+
+            using (var doc = WordprocessingDocument.Open(stream, false))
+            {
+                Assert.Null(doc.MainDocumentPart.VbaProjectPart);
+                Assert.Empty(doc.MainDocumentPart.GetPartsOfType<VbaProjectPart>());
+            }
+        }
+
+        [Fact]
         public void W038_DocxCreation_Package()
         {
             using (var stream = GetStream(TestFiles.Document, true))


iamcymentho:
@dotnet-policy-service agree

src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
```diff
@@ -116,5 +131,30 @@ void IDocumentTypeFeature<TDocumentType>.ChangeDocumentType(TDocumentType newTyp
...
 
+    /// <summary>
+    /// Removes VBA-related parts (vbaProject, vbaData, keyMapCustomizations) from the main part.
+    /// These parts are specific to macro-enabled document types and must be removed when
+    /// converting to a non-macro type.
+    /// </summary>
+    private static void RemoveVbaPartsFromMainPart(OpenXmlPart mainPart)
+    {
+        // Relationship types for VBA-related parts that should not exist in non-macro documents
+        var vbaRelationshipTypes = new HashSet<string>(StringComparer.Ordinal)
```
twsouthwick: this should be a static set. also, is it expected to grow? if not, just do it inline

src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
```diff
@@ -96,6 +98,19 @@ void IDocumentTypeFeature<TDocumentType>.ChangeDocumentType(TDocumentType newTyp
         try
         {
             Package.ChangeDocumentTypeInternal(CreateMainPart());
+
+            // When changing from a macro-enabled type to a non-macro type,
+            // remove VBA-related parts that are no longer valid for the new type.
+            // This prevents orphaned vbaProject content type entries in the package.
+            // See: https://github.com/dotnet/Open-XML-SDK/issues/618
+            var newContentType = GetContentType(newType);
```
twsouthwick: Is this something that affects all project types?

twsouthwick (changes requested):

src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
twsouthwick: if this is just for docx, implement it via a custom IDocumentTypeFeature and hook it up following the features pattern in the project

src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
```diff
@@ -96,6 +98,19 @@ void IDocumentTypeFeature<TDocumentType>.ChangeDocumentType(TDocumentType newTyp
...
+
+            if (newContentType is not null
+                && !newContentType.Contains("macroEnabled", StringComparison.OrdinalIgnoreCase)
```
twsouthwick: should we hardocde the contenttypes that support it? or should we use the newType enum value?

src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
```diff
@@ -116,5 +131,30 @@ void IDocumentTypeFeature<TDocumentType>.ChangeDocumentType(TDocumentType newTyp
...
+        var vbaRelationshipTypes = new HashSet<string>(StringComparer.Ordinal)
+        {
+            "http://schemas.microsoft.com/office/2006/relationships/vbaProject",
+            "http://schemas.microsoft.com/office/2006/relationships/keyMapCustomizations",
+        };
+
+        var partsToRemove = mainPart.ChildrenRelationshipParts
+            .Where(p => vbaRelationshipTypes.Contains(p.Value.RelationshipType))
+            .Select(p => p.Value)
+            .ToList();
```
mikeebowen: Using a foreach loop here is more efficient than a Linq query    
```csharp
List<OpenXmlPart> partsToRemove = new();

foreach (var kvp in mainPart.ChildrenRelationshipParts)
{
    if (vbaRelationshipTypes.Contains(kvp.Value.RelationshipType))
    {
        partsToRemove.Add(kvp.Value);
    }
}
```

src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
```diff
@@ -3,9 +3,11 @@
 
 using DocumentFormat.OpenXml.Features;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.IO.Packaging;
+using System.Linq;
```
mikeebowen: the Linq using can be removed - see line 148 below

mikeebowen (changes requested):

iamcymentho:
Thanks for the feedback @twsouthwick @mikeebowen! I've reworked the implementation:

- Moved VBA cleanup out of the shared `TypedPackageFeatureCollection` into `WordprocessingDocumentFeatures`
- Added a `virtual OnDocumentTypeChanged` hook in the base class so each document type can handle its own cleanup
- Replaced the `HashSet` + LINQ approach with direct `DeletePartsOfType<VbaProjectPart>()` and `DeletePartsOfType<CustomizationPart>()` calls ,  no LINQ, no extra `using`
- Uses enum pattern matching (`is MacroEnabledDocument or MacroEnabledTemplate`) instead of string checks

If Excel/PowerPoint need the same treatment, their feature classes can override `OnDocumentTypeChanged` independently.

src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
twsouthwick: for efficiency, the `partsToRemove` only needs to be created when needed

src/DocumentFormat.OpenXml/Packaging/WordprocessingDocument.cs
```diff
@@ -582,6 +582,21 @@ public WordprocessingDocumentFeatures(OpenXmlPackage package)
...
+                    or WordprocessingDocumentType.MacroEnabledTemplate)
+                {
+                    return;
+                }
+
+                if (MainPart is { } mainPart)
+                {
+                    mainPart.DeletePartsOfType<VbaProjectPart>();
+                    mainPart.DeletePartsOfType<CustomizationPart>();
+                }
```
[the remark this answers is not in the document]
twsouthwick: there's a DeletePartsOfTypeRecursively function (or something to that effect)

src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
```diff
@@ -96,6 +96,7 @@ void IDocumentTypeFeature<TDocumentType>.ChangeDocumentType(TDocumentType newTyp
         try
         {
             Package.ChangeDocumentTypeInternal(CreateMainPart());
+            OnDocumentTypeChanged(newType);
         }
```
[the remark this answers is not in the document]
twsouthwick: @mikeebowen @tomjebo are these concerns? how should we handle this?

test/DocumentFormat.OpenXml.Tests/DocxTests01.cs
```diff
@@ -361,6 +361,53 @@ public void W039_ChangeDocumentType()
...
+            stream.Position = 0;
+
+            using (var package = Package.Open(stream, FileMode.Open, FileAccess.Read))
+            {
+                var contentTypes = package.GetParts()
+                    .Select(p => p.ContentType)
+                    .ToList();
+
+                Assert.DoesNotContain("application/vnd.ms-office.vbaProject", contentTypes);
+            }
```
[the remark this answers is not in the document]
twsouthwick: @iamcymentho since that's one of the reported issues, we probably should handle that

src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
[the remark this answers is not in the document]
iamcymentho: @twsouthwick Thanks for the thorough review! I've addressed the recursive deletion and test improvements locally.

On the Excel/PowerPoint question, happy to go either way. I can add overrides for all three document types in this PR, or keep it scoped to Word and follow up separately. Let me know what you'd prefer and I'll push the update.

test/DocumentFormat.OpenXml.Tests/DocxTests01.cs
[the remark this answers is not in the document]
iamcymentho: @twsouthwick This has been handled.

src/DocumentFormat.OpenXml/Packaging/TypedPackageFeatureCollection.cs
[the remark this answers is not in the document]
iamcymentho: @twsouthwick @mikeebowen @tomjebo Just checking in, all review feedback has been addressed in the latest push:

- Moved VBA cleanup to `WordprocessingDocumentFeatures` via `OnDocumentTypeChanged` hook
- Using `DeletePartsRecursivelyOfTypeBase` for nested containers
- No LINQ, no HashSet, no unnecessary allocations

Still waiting on guidance for the Excel/PowerPoint scope question. Happy to add those overrides in this PR or a follow-up,  just let me know.

iamcymentho:
@twsouthwick @mikeebowen Friendly follow-up, this is a reminder about  latest push which addresses all review feedback. The only open question is whether to extend the VBA cleanup to Excel/PowerPoint in this PR or a separate one. Happy to proceed either way once you've had a chance to align.

PR #2042, opened 2026-03-22
avoid array allocation in qname parsing

SimonCropp:


twsouthwick (changes requested):
Are there qnames that don't have the `:`? I thought they'd all have them. If some don't than that's fine

twsouthwick:
@SimonCropp this is fine, but the readability of the code suffers. can you add this as a `TryGet(...)` kind of pattern or a tuple deconstruction to centralize the code that does this?

SimonCropp:
Good catch on both. Centralized into `PrefixName.TryParsePrefixed`, next to the existing `PrefixName.Parse` that already owned this split.

On the `:` question — these values come from `mc:ProcessContent`/`mc:PreserveElements`/`mc:PreserveAttributes` content in the file, so they're arbitrary text rather than guaranteed QNames; the old `Split(':')` + `Length != 2` already treated an unprefixed token as invalid, so that branch was reachable before too. Reviewing it though, my version had tightened things: it also rejected `:name` and `prefix:`, which `Length == 2` accepted. That's fixed — the check is now purely "exactly one colon", matching the old behavior exactly, so this is back to a pure allocation change. Added tests covering the edge cases.

PR #2054, opened 2026-03-31
Remove net35 and net40

### LINKED ISSUE #2046: Consider dropping legacy frameworks
eg net35 has been out for 19 years. 

my preference would be to have the min supported moved to 472. out for 8 years

SimonCropp:
fixes #2046


twsouthwick:
Let's start with identifying any issues that require this before we pull the trigger on it

PR #2058, opened 2026-04-07
Add forward only api with better performance 

SimonCropp:
This pull request introduces a new experimental "forward-only" API for writing OpenXML packages (such as .docx, .xlsx, .pptx) directly to any writable stream, including non-seekable streams. The new API avoids the need for buffering documents in memory, making it more efficient for large documents and for streaming scenarios (e.g., HTTP responses, cloud uploads). The implementation includes new types, documentation, and experimental API registration.

**Forward-Only OpenXML Writing API:**

* Added the new `OpenXmlPackageWriter` class, which allows forward-only creation of OpenXML documents to any writable stream without requiring a seekable stream or memory buffering. This class manages part creation, relationships, and package finalization.
* Introduced the `OpenXmlPartEntry` class, representing a single part being written in forward-only mode, with support for streaming XML content and adding part-level relationships.
* Added the `PartRelationship` struct, used to declare relationships when writing parts with the forward-only writer.
* Registered the new API as experimental (`OOXML0005`) in the `ExperimentalApis` class.

**Documentation:**

* Added comprehensive documentation (`ForwardOnlyWriting.md`) describing the new API, usage scenarios, code samples, API reference, behavioral notes, comparison with the standard API, and performance benchmarks.

### DIFF
--- a/docs/ForwardOnlyWriting.md
+++ b/docs/ForwardOnlyWriting.md
@@ -1,0 +1,196 @@
+# Forward-Only Document Writing
+
+> **Experimental**: This API is marked with `OOXML0005` and is subject to change.
+
+The `OpenXmlPackageWriter` enables forward-only creation of OpenXML documents (.docx, .xlsx, .pptx) directly to any writable stream, including non-seekable streams. This eliminates the need for a temporary `MemoryStream` that the standard `Create` methods require.
+
+## When to Use
+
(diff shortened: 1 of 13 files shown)


SimonCropp:
i went with a different approach https://github.com/SimonCropp/OpenXmlStreaming

PR #2064, opened 2026-04-15
add opt-in XML character sanitization for package save

### LINKED ISSUE #1532: Feature request for Open XML SDK - Exclude unsupported XML characters during file generation
I have encountered an scenario where a user has entered data into the database which included control characters (most likely from a 3rd party system).

Subequently when attempting to export the data to an OpenXML spreadsheet directly from the database e.g. creating a spreadsheet using a DataTable with no pre-processing, the Open XML plugin fails because of the invalid hex characters detected.

**Suggested solution**
Upon saving a datatable and while parsing each column value, detect Invalid Xml Characters and exclude them from the output file or transform them to an XML equivalent character where possible.

**The workaround**
The workaround for us is to find every field at source and exclude invalid characters being added to the database, but this isn't feasable for a large system.

The other solution is to map the data table manually parse every column while mapping the columns and exclude special characters there, but that adds additional overhead for large datasets because the table will be iterated again when OpenXML opens the dataset.

**Example error**

image

- Exception message

'', hexadecimal value 0x02, is an invalid character.

- Exception stacktrace

   at System.Xml.XmlUtf8RawTextWriter.InvalidXmlChar(Int32 ch, Byte* pDst, Boolean entitize)
   at System.Xml.XmlUtf8RawTextWriter.WriteElementTextBlock(Char* pSrc, Char* pSrcEnd)
   at System.Xml.XmlUtf8RawTextWriter.WriteString(String text)
   at System.Xml.XmlWellFormedWriter.WriteString(String text)
   at DocumentFormat.OpenXml.OpenXmlLeafTextElement.WriteContentTo(XmlWriter w)
   at DocumentFormat.OpenXml.OpenXmlElement.WriteTo(XmlWriter xmlWriter)
   at DocumentFormat.OpenXml.OpenXmlCompositeElement.WriteContentTo(XmlWriter w)
   at DocumentFormat.OpenXml.OpenXmlElement.WriteTo(XmlWriter xmlWriter)
   at DocumentFormat.OpenXml.OpenXmlCompositeElement.WriteContentTo(XmlWriter w)
   at DocumentFormat.OpenXml.OpenXmlElement.WriteTo(XmlWriter xmlWriter)
   at DocumentFormat.OpenXml.OpenXmlCompositeElement.WriteContentTo(XmlWriter w)
   at DocumentFormat.OpenXml.OpenXmlElement.WriteTo(XmlWriter xmlWriter)
   at DocumentFormat.OpenXml.OpenXmlCompositeElement.WriteContentTo(XmlWriter w)
   at DocumentFormat.OpenXml.OpenXmlPartRootElement.WriteTo(XmlWriter xmlWriter)
   at DocumentFormat.OpenXml.OpenXmlPartRootElement.Save(Stream stream)
   at DocumentFormat.OpenXml.OpenXmlPartRootElement.SaveToPart(OpenXmlPart openXmlPart)
   at DocumentFormat.OpenXml.OpenXmlPartRootElement.Save()
   at DocumentFormat.OpenXml.Packaging.OpenXmlPackage.TrySavePartContent(OpenXmlPart part)
   at DocumentFormat.OpenXml.Packaging.OpenXmlPackage.SavePartContents(Boolean save)
   at DocumentFormat.OpenXml.Packaging.OpenXmlPackage.Dispose(Boolean disposing)
   at DocumentFormat.OpenXml.Packaging.OpenXmlPackage.Dispose()
   at TestOpenXmlSpreadsheet.Helper.ExportToExcelDownload(DataSet dataSet) in C:\Users\user\Source\repos\TestOpenXmlSpreadsheet\TestOpenXmlSpreadsheet\Form1.cs:line 118
   at TestOpenXmlSpreadsheet.Form1.btnTest_Click(Object sender, EventArgs e) in C:\Users\user\Source\repos\TestOpenXmlSpreadsheet\TestOpenXmlSpreadsheet\Form1.cs:line 39
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ButtonBase.WndProc(Message& m)
   at System.Windows.Forms.Button.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, WM msg, IntPtr wparam, IntPtr lparam)

SimonCropp:
Introduces `IXmlWriterFactoryFeature` and the `UseXmlTextSanitization()` extension on `OpenXmlPackage`. When enabled, invalid XML 1.0 characters (control chars, orphaned surrogates, non-characters) are silently removed from text content, attribute values, CDATA, comments, and PI bodies before the underlying `XmlWriter` sees them. Element/attribute names and `WriteRaw` payloads are untouched.

Previously, saving a document with text like `"Hello\u0002World"` — common when exporting data straight from legacy databases — threw deep inside `System.Xml.XmlUtf8RawTextWriter`. The feature is pay-for-play: packages that don't opt in pay nothing.

Plumbing centralizes `XmlWriter` creation behind the feature so that `OpenXmlPartWriter`, `OpenXmlPartRootElement.Save`, and every `XmlDOMTextWriter` call site (including `InnerXml`/`OuterXml` paths) route through it, inherited transparently from package → part → element.

Covers Word, Excel, and PowerPoint round-trip tests plus unit tests for the sanitizer, the decorator, and the feature registration.

fixes #1532


SimonCropp:
IMO this should be the default behaviour.

src/DocumentFormat.OpenXml.Framework/Features/SanitizingXmlWriterFactoryFeature.cs
```diff
@@ -0,0 +1,22 @@
...
+internal sealed class SanitizingXmlWriterFactoryFeature : IXmlWriterFactoryFeature
+{
+    public static readonly SanitizingXmlWriterFactoryFeature Instance = new();
+
+    private SanitizingXmlWriterFactoryFeature()
+    {
+    }
+
+    public XmlWriter Create(Stream stream, XmlWriterSettings? settings)
+        => new SanitizingXmlWriter(DefaultXmlWriterFactoryFeature.Instance.Create(stream, settings));
```
twsouthwick, line 18: this shouldn't just use the default - it should pass in the existing and wrap it

src/DocumentFormat.OpenXml.Framework/Features/XmlCharacterSanitizer.cs
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Text;
+
+namespace DocumentFormat.OpenXml.Features;
+
+internal static class XmlCharacterSanitizer
+{
+    public static bool IsValidXmlChar(char c)
```
twsouthwick, line 10: is there a reason to not use `XmlConvert.IsXmlChar` here?

src/DocumentFormat.OpenXml.Framework/Features/XmlTextSanitizationExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+    public static OpenXmlPackage UseXmlTextSanitization(this OpenXmlPackage package)
+    {
+        if (package is null)
+        {
+            throw new ArgumentNullException(nameof(package));
+        }
+
+        if (package.Features.Get<IXmlWriterFactoryFeature>() is not SanitizingXmlWriterFactoryFeature)
+        {
+            package.Features.Set<IXmlWriterFactoryFeature>(SanitizingXmlWriterFactoryFeature.Instance);
```
twsouthwick, line 46: wrap the existing one here

src/DocumentFormat.OpenXml.Framework/OpenXmlPartWriter.cs
```diff
@@ -87,9 +90,12 @@ public OpenXmlPartWriter(OpenXmlPart openXmlPart, OpenXmlPartWriterSettings sett
 #endif
             };
 
-            _xmlWriter = XmlWriter.Create(partStream, xmlWriterSettings);
+            _xmlWriter = ResolveFactory(openXmlPart).Create(partStream, xmlWriterSettings);
         }
 
+        private static IXmlWriterFactoryFeature ResolveFactory(OpenXmlPart openXmlPart)
```
twsouthwick, line 96: I'd prefer to just use `GetRequiredFeature` - there's a default one already so we don't need it here as well

src/DocumentFormat.OpenXml.Framework/XmlDOMTextWriter.cs
```diff
@@ -12,16 +13,31 @@ internal class XmlDOMTextWriter : XmlWriter
         private readonly XmlWriter _writer;
 
         public XmlDOMTextWriter(Stream stream)
+            : this(stream, features: null)
         {
-            _writer = Create(stream);
+        }
+
+        public XmlDOMTextWriter(Stream stream, IFeatureCollection? features)
```
twsouthwick, line 20: pass in a IXmlWriterFactoryFeature  here instead of a full collection

twsouthwick (changes requested):
Thanks for the implementation!

Issue #2007, opened 2026-02-10
BUG：BiLevel.Threshold


waydalee:
DocumentFormat.OpenXml.Drawing.BiLevel.Threshold属性不应该是int类型，应该是string类型；（50%）
这是官方文档关于该属性的说明：
https://learn.microsoft.com/zh-cn/dotnet/api/documentformat.openxml.drawing.bilevel?view=openxml-3.0.1
<img width="874" height="149" alt="Image" src="https://github.com/user-attachments/assets/06810533-490b-4ca4-bf84-719136a8cccb" />

https://learn.microsoft.com/en-us/openspecs/office_standards/ms-oi29500/1c25f0ef-b879-47f0-af0c-b1b0d40f7d05
<img width="957" height="266" alt="Image" src="https://github.com/user-attachments/assets/32137ad7-2d33-4c12-b714-622ac1a6bafd" />

这是真实案例：
<img width="855" height="299" alt="Image" src="https://github.com/user-attachments/assets/8da23ad8-f9b4-486a-8690-041e01e6cda6" />

这是程序的异常信息：
System.FormatException: The input string '50%' was not in a correct format.

mkaszewiak:
Hi @waydalee thank you for reporting this. 
Could you please share code snippet which is causing System.FormatException? 

We reviewed the issue you have reported and SDK is behaving the same way the MS Office does - Allow to read trailing percentage and allow to write as a 1000th only. 

_**Office will read percentages formatted with a trailing percent sign or formatted as 1000th of a percent without a trailing percent sign, but only write percentages as 1000th's of a percent without a trailing percent sign.**_
https://learn.microsoft.com/en-us/openspecs/office_standards/ms-oi29500/1c25f0ef-b879-47f0-af0c-b1b0d40f7d05

Please share code sample so we can investigate it further. Thanks

mkaszewiak:
Solved by https://github.com/dotnet/Open-XML-SDK/pull/2065

PR #2065, opened 2026-04-15
Accept percentage string form in Int32Value.Parse

SimonCropp:
Office emits ECMA-376 percentage attributes such as a:biLevel/@thresh in the string form ("50%") rather than 1000ths-of-a-percent integers, causing FormatException when loading those files. Parse the "%" suffix as a fallback so affected documents round-trip.

i think this is a fix for #2007


tomjebo (dismissed):
LGTM

mikeebowen (dismissed):

mkaszewiak (dismissed):

PR #2080, opened 2026-05-06
Avoid 1 element array alloc per child in open xml element.copy children

SimonCropp:
Append(element.CloneNode(deep)) was resolving to the
  Append(params OpenXmlElement[]) overload, which allocates a fresh
  1-element array per cloned child. Switching to AppendChild eliminates
  that array. Adds a CloneNode benchmark covering several element shapes;
  deep clones now allocate ~20% fewer bytes:

    Run_Deep                       952 →   792 B  (-17%)
    Paragraph_Simple_Deep          832 →   640 B  (-23%)
    Paragraph_Formatted_Deep     2,176 → 1,728 B  (-21%)
    Body_TenParagraphs_Deep     11,129 → 8,889 B  (-20%)
    Body_HundredParagraphs_Deep 110,495 → 88,094 B (-20%, -8% time)


twsouthwick:
Awesome! Thanks for the fix

PR #2094, opened 2026-06-10
Improve loading performance.

jscarle:
Improves DOM population performance by reducing per-element overhead in the OpenXML SDK hot path: child element factory lookup no longer allocates a temporary lookup object for every XML node, tiny child factory lists use a cheaper linear scan while larger lists keep binary search, Populate reuses cached namespace resolver and markup-compatibility state, skips no-op MC processing work in the default NoProcess mode, and attribute loading avoids repeated/cold lookups by caching stable values. These changes are internal-only and preserve existing load semantics while making large worksheet/document materialization cheaper.


twsouthwick:
Do you have evidence this is faster? there are some benchmarks (that unfortunately we keep forgetting about) that may be useful as a starting point

jscarle:
Sure, this is what I was able to measure:
```
  Environment: BenchmarkDotNet 0.15.8, .NET SDK 10.0.204, .NET runtime 10.0.8, Windows 11, Intel Core Ultra 7 155H.

   Benchmark                                           Base          PR     Time change    Base alloc    PR alloc    Alloc change
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━  ━━━━━━━━━━  ━━━━━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━  ━━━━━━━━━━━━━━
   Worksheet DOM from outer XML                    39.46 ms    33.83 ms    14.3% faster      15.14 MB    12.47 MB      17.6% less
  ──────────────────────────────────────────────  ──────────  ──────────  ──────────────  ────────────  ──────────  ──────────────
   Wordprocessing DOM from outer XML               37.68 ms    27.41 ms    27.3% faster      12.54 MB     9.12 MB      27.3% less
  ──────────────────────────────────────────────  ──────────  ──────────  ──────────────  ────────────  ──────────  ──────────────
   .xlsx package open + worksheet DOM traversal    53.06 ms    43.90 ms    17.3% faster      15.30 MB    12.63 MB      17.5% less
```

The strongest evidence is the allocation reduction: all three scenarios allocate materially less memory per operation, and the wall-clock timings move in the same direction.

jscarle:
@twsouthwick Would you like additional performance testing?

mikeebowen:
- closes #2095

jscarle:
@twsouthwick Any chance we can review this?

src/DocumentFormat.OpenXml.Framework/OpenXmlCompositeElement.cs
```diff
@@ -633,6 +633,14 @@ private protected override void Populate(XmlReader xmlReader, OpenXmlLoadMode lo
 
             if (!xmlReader.IsEmptyElement)
             {
+                var context = OpenXmlElementContext;
```
twsouthwick, line 636: is this related to the performance gains as above? this is difficult to read now - what 's the point of it? if it doesn't contribute to the performance gains, let's revert it

twsouthwick (changes requested):
there are two changes going on here - can you explain why each are needed? I'd like to see each as their own change to understand if they are needed.

src/DocumentFormat.OpenXml.Framework/OpenXmlCompositeElement.cs
jscarle: Yes, this line is performance-related. OpenXmlElementContext is not just read from a field on most elements. Usually it is obtained through RootElementContext, which follows the parent chain toward the root.

In the prior implementation, Populate performed that same lookup several times for every child it loaded. The context does not change while this element’s children are being populated, so storing it once before the loop avoids repeatedly walking the same parent chain.

The cached value also lets us determine once whether markup compatibility needs to be processed. Being that NoProcess is the common mode, the loop can then skip all of the repeated evaluations and lookups.

jscarle:
There are two separate optimizations that work together to improve the combined benchmark performance during loading.

The first changes how child elements are found. Every lookup used to create a temporary ElementFactory object just so Array.BinarySearch could search for it. The new implementation compares the element name directly, avoiding that allocation. It uses a simple scan when there are four or fewer possible children because checking a few entries directly is cheaper. Larger collections still use binary search so they remain efficient.

The second change reduces repeated work while loading children and attributes. Values that remain the same during loading, such as:
- the element context
- markup-compatibility settings
- target version
- the namespace resolver
- strict-relationship flag
are calculated once and reused.

It also skips markup-compatibility tracking entirely when processing is disabled.

jscarle:
@twsouthwick Still looking to merge this if possible as it improves loading performance in very large spreadsheets.

